PR #227, opened 2017-01-03
Go 1.8 plugin architecture.

benbjohnson:
## Overview

Over the holidays I tested out the plugin support with Go 1.8 and tried integrating into Pilosa. The plugin support is not complete but I wanted to get feedback before continuing down this path much further.

This code has a few separate parts:

1. `PluginRegistry` is an object which associates plugin names with constructors for those plugins.

2. `PluginRegistry.Load()` loads a `.so` file and executes its `Init()` function which allows the plugin to register with the registry.

3. The `PQL` AST & parser have been updated to support generic `ExternalCall` functions. Any function in a query which isn't recognized gets instantiated as an `ExternalCall` and it can be looked up via the registry at execution time.

4. The `Executor` has special handling for executing the `Map()` and `Reduce()` methods on a plugin. It will also evaluate functions within the external call so you could do: `MyPlugin(Bitmap(1))`

Also, do we have more information on the use cases for plugins?

### Pros

- Creating plugins in Go means that they'll be fast and they can take advantage of all our existing types and interfaces.

- There's not a lot to tweak beyond the initial integration. Plugins that fit with the map/reduce framework should just work.

- We all know Go already so there's no additional training.

### Cons

- A plugin has to be compiled against an _exact_ copy of the codebase. If you change one thing in `pilosa` then you have to recompile your plugins. This is mostly just annoying during development. We'll have to compile version-specific plugins with each release too. End users who want to develop their own plugins will have to compile against the same source as the release.

- Plugins cannot be unloaded once they're loaded. This is primarily a development issue since production servers probably don't need to hot swap plugins. Simply restarting the server works well enough.

## TODO

- [x] Load plugins on startup. (Hot loading can be added later with `fsnotify`).
- [x] Finish `Executor` integration.
- [x] Test coverage


travisturner:
I'm going to close this PR.
Because the go plugins are so coupled with the build version anyway, @tgruben is working on a version where plugins are directly compiled into the binary; until/unless we come up with a better solution, there's no reason to dynamically load them at this point.

PR #468, opened 2017-04-21
Metrics

raskle:
Node state and StatsD metrics


raskle:
Need to recreate PR after repo went public

datadog/datadog.go
```diff
@@ -69,47 +72,61 @@ func (c *StatsClient) Tags() []string {
...
+	if err := c.client.Count("pilosa."+name, value, tags, rate); err != nil {
 		c.logger().Printf("datadog.StatsClient.Count error: %s", err)
 	}
 }
 
 // Gauge sets the value of a metric.
-func (c *StatsClient) Gauge(name string, value float64) {
-	if err := c.client.Gauge(name, value, c.tags, Rate); err != nil {
+func (c *StatsClient) Gauge(name string, value float64, rate float64) {
+	if err := c.client.Gauge("pilosa."+name, value, c.tags, rate); err != nil {
```
raskle: change prefix to constant

datadog/datadog.go
```diff
@@ -24,35 +24,38 @@ import (
 	"github.com/pilosa/pilosa"
 )
 
+// StatsD protocal wrapper using the DataDog library that added Tags to the StatsD protocal
+// statsD defailt host is "127.0.0.1:8125"
+
 const (
 	// Rate represents a metric rate of 1/sec.
-	Rate = 1
+	// Rate = 1
```
raskle: remove commented code

server.go
```diff
@@ -434,6 +447,34 @@ func checkMaxSlices(hostport string) (map[string]uint64, error) {
 	return pb.MaxSlices, nil
 }
 
+// monitorRuntime periodically polls the Go runtime metrics.
+func (s *Server) monitorRuntime() {
+	if s.MetricInterval > 0 {
```
raskle: return early

cache.go
```diff
@@ -89,10 +94,14 @@ func (c *LRUCache) Get(id uint64) uint64 {
...
 
 // Invalidate is a no-op.
-func (c *LRUCache) Invalidate() {}
+func (c *LRUCache) Invalidate() {
+	c.stats.Gauge("LRUCache", float64(c.cache.Len()), 1.0)
+}
 
 // Recalculate is a no-op.
-func (c *LRUCache) Recalculate() {}
+func (c *LRUCache) Recalculate() {
```
raskle: Add a count of when recalculate happens

cache.go
```diff
@@ -258,6 +278,11 @@ func (c *RankCache) recalculate() {
 	}
 }
 
+// SetStats passes the stats client used the the frame that owns the cache
```
raskle: Track how ofter we cross threshold, and recalculate

cmd/import.go
```diff
@@ -25,6 +25,7 @@ import (
 
 var Importer *ctl.ImportCommand
 
+// NewImportCommand runs the Pilosa import subcommand for ingesting bulk data
```
travisturner: comments should end with a period.

cmd/server.go
```diff
@@ -24,13 +24,15 @@ import (
...
 
+	"github.com/pilosa/pilosa"
 	"github.com/pilosa/pilosa/ctl"
 	"github.com/pilosa/pilosa/server"
 )
 
 // Server is global so that tests can control and verify it.
 var Server *server.Command
 
+// NewServeCmd create a pilosa server and run it with command line flags
```
travisturner: should this be "creates" and "runs"? also, should end with a period.

server.go
```diff
@@ -60,6 +62,10 @@ type Server struct {
 	// Background monitoring intervals.
 	AntiEntropyInterval time.Duration
 	PollingInterval     time.Duration
+	MetricInterval      time.Duration
+
+	// Threshold for logging long queries
+	LongQueryTime time.Duration
```
travisturner: is this used anymore? it seems like this was moved to the `Cluster` struct.

handler.go
```diff
@@ -126,7 +136,30 @@ func (h *Handler) methodNotAllowedHandler(w http.ResponseWriter, r *http.Request
...
 // ServeHTTP handles an HTTP request.
 func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	t := time.Now()
 	h.Router.ServeHTTP(w, r)
+	dif := time.Since(t)
+
+	// Handle some stats tagging
+	statsTags := make([]string, 0, 3)
+
+	if h.Cluster.LongQueryTime > 0 && dif > h.Cluster.LongQueryTime {
```
travisturner: let's revisit how the changes in `ServerHTTP` are implemented. these changes panic if `h.Cluster` or `h.Holder` are `nil`, which might be the case (at least in tests). See test `TestHandler_Query_Params_Err` as an example. Perhaps there's a cleaner way to indicate whether stats in enabled.

PR #503, opened 2017-04-28
Prevent row labels that match the column label.

### LINKED ISSUE #352: Prevent identical row and column labels.
### Description

Users should not be allowed to create a frame with the same row and column labels.

- [x] An error should be returned when passing `{"options":{"rowLabel":"X","columnLabel":"X"}}` during frame creation.

benbjohnson:
Fixes #352


travisturner:
no longer blocked by #506

PR #510, opened 2017-05-01
Webui updated

linhvo:


alanbernstein:
Does this include all the changes from https://github.com/alanbernstein/pilosa/pull/1/files?

linhvo:
@alanbernstein: I fixed :use <index> instead of :use index <index>, and this PR included your changes branched out from master

PR #517, opened 2017-05-01
Support compiled plugins

tgruben:
The idea behind this PR is to provide a platform to discuss the direction of plugins but providing executable examples to stimulate conversation. I've mainly removed the dynamic feature of @benbjohnson's original design and just built some crude examples to demonstrate the current capabilities. 

 Source Plugins
===========

The purpose of a plugin is to provide a process of extending Pilosa to handle
new or custom computations.  At this time only source level plugins are
supported. The Plugin will have full access to the entire capabilities of the
Pilosa cluster and can be referenced from its registration string. 

Plugin Creation Process
===========
- create a file located pilosa/plugins package
- init()
   In order for pilosa to be aware of the plugin, it must be resisted in its package init function via the pilosa.RegisterPlugin function

- Implement the pilosa.Plugin interface ( be aware the output type of the map is the input
  type of the reduce hidden behind the generic interface)
  - Map(ctx context.Context, index string, children []interface{}, args map[string]interface{}, slice uint64) (interface{}, error) 
  - Reduce(ctx context.Context, prev, v interface{}) interface{}
   
Examples
=======

 debug.go (argument handling)
----------------------

Prints out the arguments in the log and returns nil

```
Debug(k=key,v=1)
```
mask.go (bitmap construction)
----------------------

Creates a temporary bitmap, can't be larger than current maxslice
```
Mask(start=0, stop=10000, step=2)
```

diff.go ( frame and top-n cache navigation)
----------------------

intersect the Top-n bitmaps 

```
DiffTop(frame="frame" n=10)
```
 merge.go
----------------------

union the Top-n bitmaps
```
MergeTop(frame="frame" n=10)
```

maxmin.go ( conditional execution and custom responses)
----------------------

pick the slice with the max/min count and return its index and count

```
MaxIndex(Bitmap(id=1, frame="frame"),Bitmap(id=2,frame="frame"))
```


tgruben:
CI tests fixed

travisturner:
@tgruben did you have any other thoughts about determining and enforcing argument types for plugins?

yuce:
Wrote my first Pilosa plugin which hooks the `GetBitmap` call (with code stolen from the `Bitmap` call)

A few observations:

- Being constrained to a single file for the plugin code (due to Go's import rules) is not fun but I guess we can keep the main plugin in a git repo and just keep the bootstrap file in the `plugins` directory.
- I guess there is a plan to have hooks to other parts of Pilosa (e.g., http handler) not just PQL. 
 So I think instead of `pilosa.RegisterPlugin` something like `pilosa.RegisterPQLCallPlugin` would be more appropriate for the current plugins.
- Would it make sense to replace `args` and `children` arguments  of plugin `Map` args with a `pql.Call` to decrease the number of arguments? (bonus: the added benefit of getting the `name` of the call)

yuce:
It would be great to have a plugin function which can execute code outside of map-reduce scenarios. I've  written a `GetFrameOptions` call at: https://github.com/yuce/okapi which works around that by abusing `Map` to get frame options and `Reduce` to return it.

yuce:
@tgruben `Reduce` callback doesn't receive the same plugin object with `Map` callback. Is that intended ?

tgruben:
its known @yuce, I can’t remember why though, something got messy with the executor when I made it the same. if you can see a clean way to do it, I'd like to see it

yuce:
@tgruben Would it make sense to pass the executor and raw children to the plugin in `executeExternalCallSlice` function in `executor.go` (about line 302) so the plugins have more control over the children ?

yuce:
@tgruben I wonder whether it would be a better idea to keep plugin instances around once created instead of creating/destroying them on each call. Currently making the data for plugins survive a call is not easy (I currently work around by using the module namespace, but that would break if more than one instance of a plugin should be active with different data).

yuce:
@tgruben I've pushed a PR which makes the same plugin instance to be passed to Map and Reduce: https://github.com/tgruben/pilosa/pull/2 Does that look OK ?

plugins/merge.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/pilosa/pilosa"
+	"github.com/pilosa/pilosa/pql"
+)
+
+func init() {
+	pilosa.RegisterPlugin("MergeTop", NewMergeTopPlugin)
+}
+
+// MergeTopPlugin represents a plugin that give a set of all the bits in the top-n
```
travisturner: this comment is confusing. also, it should end with a period.

executor.go
```diff
@@ -55,6 +55,7 @@ type Executor struct {
 func NewExecutor() *Executor {
 	return &Executor{
 		HTTPClient: http.DefaultClient,
+		//	PluginRegistry: NewPluginRegistry(),
```
tgruben: just left that code from the dynamic loading plugin route, will remove

yuce:
@tgruben @travisturner I've added a PR to this PR which adds return type validation for custom calls: https://github.com/tgruben/pilosa/pull/6

jaffee:
closing for now - referenced in #998

PR #519, opened 2017-05-02
Add LICENSE to internal/ folder.

travisturner:
Because the *pb.go files get auto-generated, we can't easily
keep the license block on the top of the *.pb.go files.
Instead, we keep a copy of the LICENSE file in the same directory
as the generated *.pb.go files.


internal/public.pb.go
```diff
@@ -2590,7 +2576,7 @@ func init() { proto.RegisterFile("public.proto", fileDescriptorPublic) }
 
 var fileDescriptorPublic = []byte{
 	// 576 bytes of a gzipped FileDescriptorProto
-	0x1f, 0x8b, 0x08, 0x00, 0x00, 0x00, 0x00, 0x00, 0x02, 0xff, 0x8c, 0x54, 0x4b, 0x8e, 0xd3, 0x40,
+	0x1f, 0x8b, 0x08, 0x00, 0x00, 0x09, 0x6e, 0x88, 0x02, 0xff, 0x8c, 0x54, 0x4b, 0x8e, 0xd3, 0x40,
```
jaffee: ...? just curious

travisturner: i'm not sure what causes that. 
if you regenerate the proto files with my changes, does it change that gzipped data to something else?

jaffee: no, I get no diff running `make generate`

PR #520, opened 2017-05-02
Default partition n

travisturner:


holder_test.go
```diff
@@ -232,7 +232,7 @@ func (h *Holder) MustCreateFrameIfNotExists(index, frame string) *Frame {
 // MustCreateFragmentIfNotExists returns a given fragment. Panic on error.
 func (h *Holder) MustCreateFragmentIfNotExists(index, frame, view string, slice uint64) *Fragment {
 	idx := h.MustCreateIndexIfNotExists(index, pilosa.IndexOptions{})
-	f, err := idx.CreateFrameIfNotExists(frame, pilosa.FrameOptions{})
+	f, err := idx.CreateFrameIfNotExists(frame, pilosa.FrameOptions{CacheType: pilosa.CacheTypeRanked})
```
jaffee: I did this when I was trying to debug - I assume most of these tests want ranked frames, but I guess there probably should be a way to set it.

server/server_test.go
```diff
@@ -291,14 +291,14 @@ func TestMain_FrameRestore(t *testing.T) {
 
 	// Create frames.
 	client := m0.Client()
-	if err := client.CreateIndex(context.Background(), "x", pilosa.IndexOptions{}); err != nil && err != pilosa.ErrIndexExists {
+	if err := client.CreateIndex(context.Background(), "i", pilosa.IndexOptions{}); err != nil && err != pilosa.ErrIndexExists {
```
jaffee: What was going on with this test? I was getting an error like "index does not exist" - did you see that?

travisturner: yes, it was doing a restore on an empty node for an index that it didn't think it was responsible for. by changing the index name to `i`, the partition lined up with the node as the test expected.

holder_test.go
travisturner: oh, i see. i saw that and thought it was part of making the cache type default to "ranked". let me take a look.

PR #521, opened 2017-05-02
make ranked frames default and fix tests

jaffee:


travisturner:
@jaffee i don't think this is the way to achieve this. One problem is that it makes all the test fragments of type `lru`. While this causes the tests to pass, the tests are actually wrong. All of the `TopN` fragment tests should be of type `ranked`. Making them `ranked` causes them to fail because they need `f.RecalculateCache()` after the `SetBit()` functions. Are tests have effectively been doing `TopN` tests on `lru` caches.

I also don't like adding a `cacheType` argument to the `NewFragment()` function. It seems odd to do that when there's a default.

I took a stab at modifying things, but I now realize that we'll need more control around the `MustOpenFragment()` function in the tests. Perhaps we can change its signature to `MustOpenFragment(index, frame, view string, slice uint64, cacheType string) *Fragment` instead of changing the signature of `pilosa.NewFragment()`.

What do you think?

jaffee:
I'll take another pass @travisturner  - you're saying that many of our TopN tests were already wrong previous to this PR?

travisturner:
Yes, they were wrong in that their underlying cache was actually an LRU cache. They just happened to pass because they were running against a small data set and `TopN` was basically returning the entire cache.

jaffee:
closing in favor of https://github.com/pilosa/pilosa/pull/524

PR #524, opened 2017-05-02
change default cache type to ranked

jaffee:
All tests which use MustOpenFragment now explicitly pass a cacheType parameter.
If this parameter is an empty string, it means that whether the test works
should not depend on the cache type of the fragment. Otherwise the test should
explicitly set the cache type it needs rather than relying on the default.

### DIFF
--- a/fragment_test.go
+++ b/fragment_test.go
@@ -17,7 +17,6 @@ package pilosa_test
 import (
 	"bytes"
 	"flag"
-	"fmt"
 	"io/ioutil"
 	"math"
 	"os"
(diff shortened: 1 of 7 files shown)


fragment_test.go
```diff
@@ -642,13 +645,18 @@ func NewFragment(index, frame, view string, slice uint64) *Fragment {
...
 
 // MustOpenFragment creates and opens an fragment at a temporary path. Panic on error.
-func MustOpenFragment(index, frame, view string, slice uint64) *Fragment {
-	f := NewFragment(index, frame, view, slice)
+func MustOpenFragment(index, frame, view string, slice uint64, cacheType string) *Fragment {
+	if cacheType == "" {
+		cacheType = pilosa.DefaultCacheType
+	}
+	f := NewFragment(index, frame, view, slice, cacheType)
+	fmt.Println("CacheType", f.CacheType)
```
travisturner: you can probably remove this print statement.

jaffee: woops

jaffee:
@travisturner I changed some error messages to have differentiation if/when the server test fails again - if that looks good, I'll merge

travisturner:
@jaffee looks ok to me

PR #527, opened 2017-05-03
docker update

yuce:
I've removed the micro docker stuff from this PR and only include updates that fix pilosa binary version, etc. since docker recently introduced a new feature called multistage builds which enables creating the micro image without a build container.

Also, our submission to the official images may require us to keep the docker stuff in its own repo.


yuce:
Great, thanks!

PR #530, opened 2017-05-05
Use "make" to generate bare-bones docker images

codysoyland:
This commit modifies the Makefile so it creates creates a Dockerfile
dynamically using "scratch" as the base image, builds Pilosa outside of
Docker, and creates a Docker image from the resulting Dockerfile


codysoyland:
At looks like Docker 17.05.0 is out, so I think the multi-stage build process is going to be better for us anyway. Good call on the CGO dynamic linking point. New PR addresses everything: https://github.com/pilosa/pilosa/pull/535

PR #535, opened 2017-05-08
Use Docker multi-stage build in Dockerfile

codysoyland:
This utilizes Docker multi-stage builds to produce a minimal Docker image. Requires Docker 17.05.0.


yuce:
That looks cool. I'll update my docker and test it.

PR #537, opened 2017-05-08
Add syntax hints tooltip to WebUI

alanbernstein:
screen shot 2017-05-08 at 6 00 17 pm

### DIFF
--- a/webui/assets/style.css
+++ b/webui/assets/style.css
@@ -304,12 +304,15 @@ td{
 .query-tooltip {
   position: relative;
   display: inline;
-  color: #1db598;
+  color: #000;
   margin-left: 5px;
 }
 
+.query-tooltip:hover {
+  color: #000;
+}
+
 .query-tooltip-content {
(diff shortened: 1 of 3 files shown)


webui/assets/style.css
```diff
@@ -295,4 +295,42 @@ td{
...
+  background-color: rgb(250, 250, 250);
+  border: solid 1.5px #e4eff4;
+  color: #102445;
+  border-radius: 2px;
+  padding: 15px;
+  margin-bottom: 15px;
+
+  /* Position the tooltip */
+  position: absolute;
+  /*
```
codysoyland: Remove commented css

alanbernstein: This was me trying to figure out how to move the tooltip without breaking everything else

codysoyland:
The latest changes look good! I did just notice that the examples use "row" and "column", but I would prefer we use "rowID" and "columnID" as those are the defaults.

PR #543, opened 2017-05-12
Tanimoto plugin

linhvo:
This branched out from todd/compile-plugin, I created a PR to todd branch, but I couldn't assign reviewer on that PR so I create this PR for better review I think. Tanimoto plugin is in `tanimoto.go` and `tanimoto_test.go`, I also removed tanimoto part from Top() function in `fragment.go`.


linhvo:
@travisturner: I'm closing this one since plugin-in is now separate package and we can install with `make install-plugin` after Todd's plugin is merged. The tanimoto is now in: https://github.com/linhvo/tanimoto. Let me know if I should create this package inside pilosa.

Issue #413, opened 2017-03-29
export sub command needs view argument 


raskle:
The Pilosa Export sub command needs an additional parameter for View.   We should also consider if we want to make this an optional parameter and by default export the standard view.

### Expected behavior
Now that we have multiple views we need to specify which view to export in the pilosa export sub command. 

### Actual behavior
Currently the pilosa export command will return nothing as the view argument is blank and cannot be set.

### Steps to reproduce the behavior
./pilosa export --host 127.0.0.1:10101 -d <my_db> -f <my_frame> 
returns nothing

The API does yield the expected data with a view specified
curl --header "Accept: text/csv" "http://localhost:10101/export?db=my_db&frame=my_frame&slice=0&view=standard"

PR #548, opened 2017-05-15
#413 set standard view as default for export command

linhvo:


cmd/export.go
```diff
@@ -53,6 +53,7 @@ The file does not contain any headers.
 	flags.StringVarP(&Exporter.Host, "host", "", "localhost:10101", "host:port of Pilosa.")
 	flags.StringVarP(&Exporter.Index, "index", "i", "", "Pilosa index to export into.")
 	flags.StringVarP(&Exporter.Frame, "frame", "f", "", "Frame to export into.")
+	flags.StringVarP(&Exporter.View, "view", "v", "", "View to export into - default standard")
```
travisturner: I think we can remove the word `into` from this line and the lines above.

client.go
```diff
@@ -394,8 +394,8 @@ func (c *Client) ExportCSV(ctx context.Context, index, frame, view string, slice
 		return ErrIndexRequired
 	} else if frame == "" {
 		return ErrFrameRequired
-	} else if view == "" {
-		view = ViewStandard
+	} else if view != ViewInverse || view != ViewStandard {
```
travisturner: won't this condition always return an error?

linhvo: oh right, my bad. I'm fixing it

linhvo:
^ I agree, it's totally better. I'll update as your suggestion then

PR #550, opened 2017-05-15
added goveralls

yuce:


.travis.yml
```diff
@@ -1,14 +1,18 @@
...
   apt:
     sources:
       - sourceline: 'ppa:masterminds/glide'
     packages:
       - glide
-script: make test
+before_install:
+  - go get github.com/mattn/goveralls
+script:
+  - make vendor && $HOME/gopath/bin/goveralls -service=travis-ci -ignore "internal/internal.go,internal/public.pb.go,internal/private.pb.go"
```
codysoyland: Just to confirm, this doesn't run tests on the whole vendor dir, does it? We had to explicitly exclude it in `make test`.

yuce: As far as I see at: https://coveralls.io/builds/11529694 it doesn't include anything from vendor.

codysoyland: Oh yeah, and the build time on Travis is about the same (roughly 5 minutes). Looks good to me!

yuce: Great!

PR #559, opened 2017-05-18
fix 3 separate bugs in bitmapCountRange

jaffee:
I think these are all real, but let me know if I've just completely misunderstood how this is supposed to work. Hoping to get at least 2 reviews since we're pretty deep into the bowels here.

in order of the diff:

1. When the start and end of the range fall in the same word, special handling
is needed to "mask" off the beginning and end of the word simultaneously to
avoid counting bits at the beginning or end of the word that aren't in the
range.

2. `i++` is needed at the end of the first partial word to avoid counting this
word in the next block.

3. the shift amount for right shifts is 64 - (end % 64) rather than just end
% 64. If end is (e.g.) 68, then 68 - 64 is 4 and we are only interested in the
first 4 bits of the word, so we must right shift by 60 bits, not 4 bits.

### DIFF
--- a/roaring/roaring.go
+++ b/roaring/roaring.go
@@ -907,13 +907,13 @@ func (c *container) bitmapCountRange(start, end uint32) int {
 	// Special case when start and end fall in the same word.
 	if i == j {
 		offi, offj := start%64, 64-end%64
-		n += popcount((c.bitmap[i] << offi) >> (offj + offi))
+		n += popcount((c.bitmap[i] >> offi) << (offj + offi))
 		return int(n)
 	}
 
 	// Count partial starting word.
 	if off := start % 64; off != 0 {
-		n += popcount(c.bitmap[i] << off)
+		n += popcount(c.bitmap[i] >> off)
 		i++
 	}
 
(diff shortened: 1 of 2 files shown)


jaffee:
I just pushed another commit which corrects this based on my understanding that the least significant bit in one of a bitmap's 64 bit integers is the "zeroth" bit in that integer. So setting bit 0 sets the first integer in the bitmap to 1 rather than 2**63 as I had been thinking.

roaring/roaring.go
```diff
@@ -916,8 +924,8 @@ func (c *container) bitmapCountRange(start, end uint32) int {
 
 	// Count partial ending word.
 	if int(j) < len(c.bitmap) {
-		if off := end % 64; off != 0 {
-			n += popcount(c.bitmap[j] >> off)
+		if off := 64 - (end % 64); off != 0 {
```
travisturner: in this case, `off` will never be `0`.
this check should probably be `off != 64` (it's currently not affecting results because it's shifting out the entire 64 bits anyway and getting a count of `0`, but it's an unnecessary operation)

jaffee:
good call on that fix @travisturner - do you think we should just remove the branch entirely? simplify the code and avoid the branching instruction in what is probably the most common case - doing a bit of extra work when `off == 64`

travisturner:
@jaffee i suppose we could since that case isn't going to happen very often, and the performance hit we take is just a 64-bit shift. i'm not sure what the performance hit is for checking equality.

travisturner:
@jaffee before this gets merged we need to add the license header to the new .go file.

Issue #316, opened 2017-02-14
Support extended attributes for frames and databases


alanbernstein:
**Pilosa should maintain some additional information about databases and frames**, and allow users to set, modify, and view it, as appropriate. For the sake of consistency I refer to this information as "attributes", although Pilosa currently uses the terms "options" and "schema" for similar features.

Like BitmapAttrs and ProfileAttrs, FrameAttrs and IndexAttrs should support storing arbitrary key-value pairs. Additionally, **certain reserved keys should provide specific functionality**. Conceptually, rowLabel and columnLabel are examples of reserved keys. **Independently of (the rest of) this proposal, from a user's perspective, it is preferable to consolidate the terminology and interfaces for interacting with metadata**. The proposed Attrs system might incorporate rowLabel and columnLabel, or the current "options" system might incorporate new options.

## Existing similar functionality that could be consolidated

- SetProfileAttrs and SetBitmapAttrs PQL commands
  The current interface for these is sufficient in itself. Adding new functionality for FrameAttrs and IndexAttrs, via a new interface, would mean we have two different interfaces for working with different types of attributes.
- /schema endpoint
  Certain index and frame attributes are arguably part of the schema, so a user might expect to see this information available at the schema endpoint.
- "options" for databases and frames, as set at the /index and /frame endpoints
  From a user perspective, this interface already has significant overlap with new proposed attributes.

## Motivation 
A major motivation for this proposal is the desire for a system to support a generic web app that can automatically interface with any Pilosa database, as long as certain FrameAttr keys are defined. This app would bear some resemblance to the transportation demo, including summary and visualization components. These same features could also be used by the web console, or homepage demo. These components would be supported by at least three new FrameAttrs: rowMapping, maxRowID, description (in descending order of importance).

### rowMapping
In the case that row IDs represent external entities in some nontrivial way, a mapping must be maintained. If the mapping is just a big hashmap between rowIDs and strings, or rowIDs and external IDs, then this can't be captured compactly*. However, a straightforward mapping, like that used for indexing numeric data, can be represented compactly, and stored as an attribute. A machine-readable representation of the mapping can be used to reconstruct the function, or its inverse. The inverse function can then be used by a generic client to reconstruct the original data, or an approximation to it. 

[...]

PR #561, opened 2017-05-19
Implement handleGetFrames

alanbernstein:
Is this the intended use of this endpoint? This would be useful for me.

```
$ curl :10101/index/benchindex/frame

{
    "index": {
        "frames": [
              {
                "name": "slice-height",
                "views": [
                    {
                        "name": "standard"
                    }
                ]
            },
            {
                "name": "zipf",
                "views": [
                    {
                        "name": "standard"
                    }
                ]
            }
        ],
        "name": "benchindex"
    }
}
```

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -87,7 +87,7 @@ func NewRouter(handler *Handler) *mux.Router {
 	router.HandleFunc("/index/{index}", handler.handlePostIndex).Methods("POST")
 	router.HandleFunc("/index/{index}", handler.handleDeleteIndex).Methods("DELETE")
 	router.HandleFunc("/index/{index}/attr/diff", handler.handlePostIndexAttrDiff).Methods("POST")
-	//router.HandleFunc("/index/{index}/frame", handler.handleGetFrames).Methods("GET") // Not implemented.
+	router.HandleFunc("/index/{index}/frame", handler.handleGetFrames).Methods("GET") // Not implemented.
 	router.HandleFunc("/index/{index}/frame/{frame}", handler.handlePostFrame).Methods("POST")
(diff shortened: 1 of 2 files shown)


travisturner:
I would expect the output you suggested to be here instead:
```
router.HandleFunc("/index/{index}", handler.handleGetIndex).Methods("GET")
```
I think that `handleGetFrames` should return a list of frames for that index. whether that list includes the full schema or not i don't know, but it should include a list of endpoints to each frame.

@codysoyland do you have an idea on expectations for these unimplemented endpoints?

codysoyland:
I agree with @travisturner that the output in @alanbernstein's comment should live at `/index/{index}` instead of `/index/{index}/frame`. I'm not sure that we need to implement `/index/{index}/frame` because it becomes redundant if both endpoints returned the list of frames. I would probably leave it alone. If we do choose to implement this, let's move the "frames" key to the top level and remove the "Not implemented" comment.

alanbernstein:
`/schema` and `/status` currently return big JSON objects containing all index and frame names, plus some frame info that doesn't seem to fit either endpoint well.

`/status` returns the result of `ClusterStatus()` which makes sense from an internal perspective, but it's weird from a user perspective.

Are there any design docs or opinions about how this stuff should be organized? I have a few suggestions I can write up (as a counterpart to #316)

travisturner:
@alanbernstein I'm happy for you to write up your suggestions. Those two endpoints were never really thought through very well, and so far have been primarily used to provide information to the WebUI.

alanbernstein:
I'm working on a proposal that will likely affect this, closing for now.

PR #571, opened 2017-05-22
added xor support to roaring

tgruben:
Added xor support for roaring

### DIFF
--- a/roaring/roaring.go
+++ b/roaring/roaring.go
@@ -1563,6 +1563,127 @@ func differenceBitmapBitmap(a, b *container) *container {
 	}
 	return output
 }
+func (b *Bitmap) Xor(other *Bitmap) *Bitmap {
+	output := &Bitmap{}
+
+	ki, ci := b.keys, b.containers
+	kj, cj := other.keys, other.containers
+
+	for {
+		var key uint64
+		var container *container
+
+		ni, nj := len(ki), len(kj)
+		if ni == 0 && nj == 0 { // eof(i,j)
+			break
+		} else if ni == 0 || (nj != 0 && ki[0] > kj[0]) { // eof(i) or i > j
+			key, container = kj[0], cj[0].clone()
+			kj, cj = kj[1:], cj[1:]
+		} else if nj == 0 || (ki[0] < kj[0]) { // eof(j) or i < j
+			key, container = ki[0], ci[0].clone()
+			ki, ci = ki[1:], ci[1:]
+		} else { // i == j
+			key, container = ki[0], xor(ci[0], cj[0])
+			ki, ci = ki[1:], ci[1:]
+			kj, cj = kj[1:], cj[1:]
+		}
+
+		output.keys = append(output.keys, key)
+		output.containers = append(output.containers, container)
+	}
+
+	return output
+}
+
+func xor(a, b *container) *container {
+	if a.isArray() {
+		if b.isArray() {
+			return xorArrayArray(a, b)
+		} else {
+			return xorArrayBitmap(a, b)
+		}
+	} else {
+		if b.isArray() {
+			return xorArrayBitmap(b, a)
+		} else {
+			return xorBitmapBitmap(a, b)
+		}
+	}
+}
+
+func xorArrayArray(a, b *container) *container {
+	output := &container{}
+	na, nb := len(a.array), len(b.array)
+	for i, j := 0, 0; ; {
+		if i >= na && j >= nb {
+			break
+		} else if i < na && j >= nb {
+			output.add(a.array[i])
+			i++
+			continue
+		} else if i >= na && j < nb {
+			output.add(b.array[j])
+			j++
+			continue
+		}
+
+		va, vb := a.array[i], b.array[j]
+		if va < vb {
+			output.add(va)
+			i++
+		} else if va > vb {
+			output.add(vb)
+			j++
+		} else { //==
+			i++
+			j++
+		}
+	}
+	return output
+}
+
+func xorArrayBitmap(a, b *container) *container {
+	output := &container{}
+	itr := newBufIterator(newBitmapIterator(b.bitmap))
+	for i := 0; ; {
+		vb, eof := itr.next()
+		if i >= len(a.array) && eof {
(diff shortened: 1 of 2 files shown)


roaring/roaring.go
```diff
@@ -1563,6 +1563,127 @@ func differenceBitmapBitmap(a, b *container) *container {
...
+
+func xorArrayBitmap(a, b *container) *container {
+	output := &container{}
+	itr := newBufIterator(newBitmapIterator(b.bitmap))
+	for i := 0; ; {
+		vb, eof := itr.next()
+		if i >= len(a.array) && eof {
+			break
+		} else if i >= len(a.array) {
+			output.add(vb)
```
jaffee: if you started by creating a copy of the bitmap container, we could just iterate over the array stop as soon as the array is done. rather than iterating over the whole array and bitmap.

roaring/roaring.go
```diff
@@ -1563,6 +1563,127 @@ func differenceBitmapBitmap(a, b *container) *container {
...
+			i++
+		} else if va > vb {
+			output.add(vb)
+			j++
+		} else { //==
+			i++
+			j++
+		}
+	}
+	return output
```
jaffee: shouldn't we convert to a bitmap if cardinality is higher than 4096?

tgruben: @jaffee container.add already handles that check/conversion

roaring/roaring.go
tgruben: @jaffee I'm not sure i buy that's much improvement unless the cardinality of the operands is significantly different.  Worse case scenario in the copy first situation, they are identical bitmaps.  The optimal path in that situation would be just to traverse and not allocate anything.

Worse case in this situation is an empty bitmap and fulll bitmap resulting in an allocation and a traversal.  Do you think that would warrant a benchmark?

jaffee: The cardinality will most likely be significantly different right? Since array type containers are guaranteed to be < 4096 bits, and bitmap containers will be 4096-65535 bits. 

My thinking is that because the bitmap will usually have many more bits set and they will probably all appear in  the output (with the exception of the few which happen to collide with the array), it makes sense to copy the bitmap and set or unset the comparatively few bits which are in the array.

We should certainly benchmark - especially since we already have this implementation, but I think my way is better 😛.

Also I'm just trying to slow you down so that I can get RLE in first 😈

roaring/roaring.go
```diff
@@ -1563,6 +1563,127 @@ func differenceBitmapBitmap(a, b *container) *container {
 	}
 	return output
 }
+func (b *Bitmap) Xor(other *Bitmap) *Bitmap {
```
travisturner: this may seem picky, but i would move the `Xor` function up to be with the other functions on the `(b *Bitmap)` struct, and i would move the remaining functions directly after `differenceBitmapBitmap` in the file just to be consistent with the existing organization of the file.

roaring/roaring.go
```diff
@@ -1563,6 +1563,127 @@ func differenceBitmapBitmap(a, b *container) *container {
...
+	output := &container{
+		bitmap: make([]uint64, bitmapN),
+	}
+
+	for i := 0; i < bitmapN; i++ {
+		v := a.bitmap[i] ^ b.bitmap[i]
+		output.bitmap[i] = v
+		output.n += int(popcnt(v))
+	}
+	return output
```
travisturner: do we need to worry about the case where the xor results in a bitmap with count < 4096. i.e. do we need to call `output.convertToArray()` prior to returning?

roaring/roaring.go
```diff
@@ -1564,6 +1597,88 @@ func differenceBitmapBitmap(a, b *container) *container {
...
+			return xorBitmapBitmap(a, b)
+		}
+	}
+}
+
+func xorArrayArray(a, b *container) *container {
+	output := &container{}
+	na, nb := len(a.array), len(b.array)
+	for i, j := 0, 0; ; {
+		if i >= na && j >= nb {
```
jaffee: this can be done more cleanly like:
```
	for i, j := 0, 0; i < na && j < nb; {
```

roaring/roaring.go
```diff
@@ -1564,6 +1597,88 @@ func differenceBitmapBitmap(a, b *container) *container {
...
+	return output
+}
+
+func xorArrayBitmap(a, b *container) *container {
+	output := b.clone()
+	for _, v := range a.array {
+		if b.bitmapContains(v) {
+			output.remove(v)
+		} else {
+
```
jaffee: extra line

roaring/roaring_test.go
```diff
@@ -130,6 +130,46 @@ func TestBitmap_Union(t *testing.T) {
...
+		bm1.Add(i)
+	}
+
+	result := bm0.Xor(bm1)
+	if n := result.Count(); n != 4999 {
+		t.Fatalf("unexpected n: %d", n)
+	}
+}
+
+func TestBitmap_Xor_BitmapBitmap(t *testing.T) {
```
jaffee: I'd like to see the case tested where the two bitmaps have the same bits as well.

roaring/roaring_test.go
```diff
@@ -130,6 +130,46 @@ func TestBitmap_Union(t *testing.T) {
 	}
 }
 
+func TestBitmap_Xor_ArrayArray(t *testing.T) {
```
jaffee: I think we should test the case where one or both of the arrays are empty.

roaring/roaring_test.go
```diff
@@ -130,6 +130,46 @@ func TestBitmap_Union(t *testing.T) {
...
+	bm0 := roaring.NewBitmap(0, 1000001, 1000002, 1000003)
+	bm1 := roaring.NewBitmap(0, 50000, 1000001, 1000002)
+	result := bm0.Xor(bm1)
+	if n := result.Count(); n != 2 {
+		t.Fatalf("unexpected n: %d", n)
+	}
+}
+
+func TestBitmap_Xor_ArrayBitmap(t *testing.T) {
```
jaffee: test empty array/container

PR #572, opened 2017-05-23
discriminate error messages in server.Open

jaffee:
was getting an error from not specifying internal hosts properly (i think) but couldn't tell exactly where it was from by the error message.


jaffee:
I agree @raskle - can we do that in a separate issue/PR though? I'd like to get this merged to make it easier to track down config issues when starting clusters.

PR #573, opened 2017-05-23
33 validate config

linhvo:
I spent a lot of time on using reflection to get valid config tag and just realized that I don't need it for validate field so I removed that part.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -22,6 +22,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
+	"reflect"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


cmd/root.go
```diff
@@ -118,10 +123,16 @@ func setAllConfig(v *viper.Viper, flags *pflag.FlagSet, envPrefix string) error
 		if err != nil {
 			return fmt.Errorf("error reading configuration file '%s': %v", c, err)
 		}
+
+		for _, key := range v.AllKeys() {
+			if _, ok := validTags[key]; !ok {
+				return fmt.Errorf("invalid tag: %v", key)
```
jaffee: instead of "invalid tag", I think "invalid option in configuration file" would be a more helpful message.

cmd/root_test.go
```diff
@@ -171,3 +171,27 @@ func TestRootCommand(t *testing.T) {
...
+	}
+	config := `data-dir = "/tmp/pil5_0"
+bind = "127.0.0.1:15000"
+
+[cluster]
+  poll-interval = "2m0s"
+  replicas = 2
+  partitions = 128
+  hosts = [
+    "127.0.0.1:15000",
```
jaffee: let's try to use 10101 everywhere - just to keep it consistent

linhvo:
Thanks @jaffee, I fixed your comments. I'm gonna merge and close this.

PR #574, opened 2017-05-24
Handler plugins

yuce:


jaffee:
In order to clean up our PR list, I've pushed this branch to https://github.com/pilosa/pilosa/tree/handler-plugins and it is ticketed at #997

PR #577, opened 2017-05-24
rewrite intersectCountArrayBitmap for perf test

jaffee:
This is a 15x performance boost to TopN queries with one Bitmap on the NYC taxi dataset. 

e.g. ```TopN(Bitmap(frame=passenger_count, rowID=1), frame=total_amount_dollars)```

Results appear to be the same with both implementations. 

Basically the idea is to iterate over the array container rather than the bitmap container since the bitmap container basically means iterating over each bit of each uint64 (65536 bits), vs guaranteed <4096 uint32s in the array container.

I believe this is strictly faster than the previous implementation, but let me know if there are cases where performance would be lost (or if it's just incorrect).

### DIFF
--- a/roaring/roaring.go
+++ b/roaring/roaring.go
@@ -1228,7 +1228,7 @@ func intersectionCountArrayArray(a, b *container) (n uint64) {
 	return n
 }
 
-func intersectionCountArrayBitmap(a, b *container) (n uint64) {
+func intersectionCountArrayBitmapOld(a, b *container) (n uint64) {
 	// Copy array header so we can shrink it.
 	array := a.array
 	if len(array) == 0 {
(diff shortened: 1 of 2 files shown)


travisturner:
@jaffee this change looks good to me. i can't speak to the performance gain, but it makes sense that iterating over the array (instead of the bitmap) would result in fewer operations on average.

i wonder if there would be additional performance gain in recognizing which array elements fell into the same 64-bit space (`i` in `intersectionCountArrayBitmap()`). then you would only need to do one `&` per 64-bit space, as opposed to every array element. with that said, the work needed to determine the array values that fall in the same 64-bit space may negate the gains, so it might not be worth it.

Issue #312, opened 2017-02-14
Pilosa server ignores unknown query arguments


yuce:
### Expected behavior

If an unknown query argument passed, Pilosa server should return an error.

### Actual behavior

Pilosa server ignores the query argument

### Steps to reproduce the behavior

```
curl -XPOST 'localhost:15000/query?db=sample&not_exists=true' -d 'Bitmap(id=1, frame="foo")'
```

### Information about your environment (OS/architecture, CPU, RAM, cluster/solo, configuration, etc.)

For feature requests, please provide the following:

yuce:
Closing this, since we don't use query arguments anymore.

yuce:
Reopening this issue, since I was wrong: passing unknown query arguments raises no errors.

PR #578, opened 2017-05-24
#312 validate unkown query params

linhvo:


handler.go
```diff
@@ -843,6 +844,12 @@ func (h *Handler) readProtobufQueryRequest(r *http.Request) (*QueryRequest, erro
 // readURLQueryRequest parses query parameters from URL parameters from r.
 func (h *Handler) readURLQueryRequest(r *http.Request) (*QueryRequest, error) {
 	q := r.URL.Query()
+	validQuery := h.getValidURLQuery(r)
```
yuce: Since the value of `validQuery` is always the same, would it make sense to save it in a global variable ?

handler.go
```diff
@@ -875,6 +882,20 @@ func (h *Handler) readURLQueryRequest(r *http.Request) (*QueryRequest, error) {
 	}, nil
 }
 
+func (h *Handler) getValidURLQuery(r *http.Request) map[string]bool {
```
yuce: Since this function doesn't depend on a Handler, would it make sense to make it a standalone function? Also name of the function is a bit misleading. Would something like `validURLQueryArgs` work better ?

handler.go
```diff
@@ -875,6 +882,20 @@ func (h *Handler) readURLQueryRequest(r *http.Request) (*QueryRequest, error) {
...
+	validQuery := make(map[string]bool)
+	args := reflect.ValueOf(QueryRequest{})
```
yuce: Small nitpick: since `args.Type()` is used in a loop below, I think it would be better to save it in a var like `argsType` instead of `args`.

linhvo:
@yuce: I fixed your comments and have change it a bit. Ss you mentioned to make it a standalone function, I think probably writing this as general for validate any struct that we might use it to validate other params in the future by accept `interface`, instead of specific `QueryRequest`. So I changed it to `validOptions` for general name. How do you think?

yuce:
This looks good!

PR #580, opened 2017-05-25
Add statsd gauges for row count.

### LINKED ISSUE #421: Add StatsD gauges for row and column count
### Description
With statsD gauges track 
 Row and Column counts
 Max Slice

This gauge will also be tagged with the:
DB
Frame
View
Slice

The gauges report the latest counts per axis
This value is reset when reloading the pilosa cluster

benbjohnson:
## Overview

This pull request adds a `"rows"` gauge that tracks the maximum row for each fragment. I originally tried to add a `"columns"` gauge too but I couldn't get it to quickly compute the max on start-up without scanning all the data files.

I also changed the `"maxSlice"` stat from a `count` to a `gauge`. Previously it was tracking the number of fragments created instead of the maximum slice.

Fixes #421.

### DIFF
--- a/fragment.go
+++ b/fragment.go
@@ -86,6 +86,9 @@ type Fragment struct {
 	cache     Cache
 	CacheSize uint32
 
+	// Stats reporting.
+	maxRowID uint64
+
 	// Cache containing full rows (not just counts).
 	rowCache BitmapCache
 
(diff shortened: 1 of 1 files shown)


benbjohnson:
@raskle I like your approach better. I removed the `maxSlice` changes in my PR in f04b8d4.

fragment.go
```diff
@@ -409,6 +417,12 @@ func (f *Fragment) setBit(rowID, columnID uint64) (changed bool, err error) {
 
 	f.stats.Count("setN", 1)
 
+	// Update row & column count if they have increased.
```
travisturner: I think we should remove `column` from this comment.

benbjohnson: Fixed in a961473.

PR #586, opened 2017-05-29
Improve test coverage for ctl package

linhvo:
Package ctl is now 75% percentage cover. There are 2 tests for check.go and inspect.go I need to spend more time to understand roaring, that they both need a correct roaring bitmap file to test happy path to improve its coverage to increase more percentage cover, I did add ToDo in the file to come back later. 

The other thing is I standup a server on tests, which is similar what pilosa_test does to standup pilosa server on random port. I’m not sure if there’s a better way to reuse code in pilosa_test (like import package?) rather than duplicate many code. 
Any recommendation @jaffee

### DIFF
--- a/ctl/backup_test.go
+++ b/ctl/backup_test.go
@@ -1,0 +1,15 @@
+package ctl
+
+import (
+	"testing"
+	"golang.org/x/net/context"
+)
+
+func TestBackupCommand_FileRequired(t *testing.T){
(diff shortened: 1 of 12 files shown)


linhvo:
thanks @codysoyland, I'm gonna have this merged first and will create another issue for cluster creation code refactor.

PR #587, opened 2017-05-30
Document inverse=true for TopN

alanbernstein:


docs/query-language.md
```diff
@@ -346,11 +347,21 @@ TopN(frame="stargazer")
...
 
 ```
+TopN(frame="stargazer", inverse=true)
+```
+
+Returns `[{"key": 1, "count": 2}, {"key": 2, "count": 2}, {"key": 3, "count": 1}]`
+
+* key is a repository ID
+* count is amount of users
+* Results are the number of user that starred each repository in descending order for all respositories in the stargazer frame.
```
jaffee: number of userS

travisturner:
@jaffee is this good now?

PR #592, opened 2017-05-31
add support for bit flip (negate) in roaring

tgruben:


roaring/roaring_test.go
```diff
@@ -199,6 +199,49 @@ func TestBitmap_Xor_BitmapBitmap(t *testing.T) {
 	}
 }
 
+// Ensure bitmap turn on and turn off
+func TestBitmap_Fipp_Empty(t *testing.T) {
```
jaffee: Fipp

roaring/roaring.go
```diff
@@ -706,6 +706,27 @@ func (b *Bitmap) Check() error {
 	return a
 }
 
+func (b *Bitmap) Flip(start, end uint64) *Bitmap {
```
travisturner: This function should have a comment describing what it does.

roaring/roaring_test.go
```diff
@@ -199,6 +199,49 @@ func TestBitmap_Xor_BitmapBitmap(t *testing.T) {
 	}
 }
 
+// Ensure bitmap turn on and turn off
```
travisturner: This comment is awkwardly worded. Also, comments should end with a period.

roaring/roaring.go
tgruben: I went down that route, to begin with, then I discovered the array flip function, given the range constraint, would be exactly the same as the higher level function and then I would get got the bitmap for "free" if just implemented at a higher level.  I suspect bitmap flip can be a bit more performant but I am skeptical of the magnitude of improvement.

roaring/roaring.go
```diff
@@ -706,6 +706,28 @@ func (b *Bitmap) Check() error {
 	return a
 }
 
+//Perform a logical negate of the bits in the range [start,end].
+func (b *Bitmap) Flip(start, end uint64) *Bitmap {
+	result := NewBitmap()
+	itr := b.Iterator()
+	itr.Seek(start)
+	v, eof := itr.Next()
+	for i := start; i < end; i++ {
```
travisturner: This logic doesn't handle the case where `start` is greater than existing bits in the bitmap.
I think you need to handle the case `if v < start` before the for loop.
you can see an example of this in test `TestBitmap_Flip_Array`: change `results := bm.Flip(0, 5)` to `results := bm.Flip(2, 5)` and you still get the same count on results.

roaring/roaring.go
```diff
@@ -101,6 +102,14 @@ func (b *Bitmap) Add(a ...uint64) (changed bool, err error) {
 
 	return changed, nil
 }
+func (b *Bitmap) String() string {
```
travisturner: 1. maybe move this function before `Slice()` in the code. it's weird to have it dropped in the middle of the two `Add, add` functions.
2. it seems odd to include `\n` at the end of a `String()` function
3. https://github.com/pilosa/pilosa/blob/master/cache.go#L396 contains a similar `String()` function that doesn't have an extra space on the final iteration.

roaring/roaring_test.go
```diff
@@ -199,6 +199,67 @@ func TestBitmap_Xor_BitmapBitmap(t *testing.T) {
...
+		t.Fatalf("unexpected %s ", results.String())
+	}
+	results = results.Flip(0, 4)
+	if !reflect.DeepEqual(results.Slice(), []uint64{0, 1, 2, 3, 4, 8, 16, 32, 64, 128, 256, 512, 1024}) {
+		t.Fatalf("unexpected %s ", results.String())
+	}
+
+}
+
+func TestBitmap_Flip_Bitmap(t *testing.T) {
```
travisturner: add a comment here describing what's being tested.

roaring/roaring_test.go
```diff
@@ -199,6 +199,67 @@ func TestBitmap_Xor_BitmapBitmap(t *testing.T) {
...
+	if n := results.Count(); n != size/2 {
+		t.Fatalf("unexpected n: %d", n)
+	}
+	results = results.Flip(0, size-1) //flipping back should be the same
+	if n := results.Count(); n != size/2 {
+		t.Fatalf("unexpected n: %d", n)
+	}
+}
+
+func TestBitmap_Flip_After(t *testing.T) {
```
travisturner: add a comment here describing what's being tested.

roaring/roaring.go
tgruben: I am not sure what the go best practice is in regards to 1.  I personally like to have all the public/exported functions grouped together followed by the private functions.  You have anything other than "wierd"?

travisturner: I'm not sure what best practice is, but looking at the existing functions on `(b *Bitmap)`:
```
Clone
Add
add
Contains
Remove
remove
Max
Count
CountRange
Slice
SliceRange
ForEach
ForEachRange
OffsetRange
container
```
I agree with you that it makes sense to put public functions grouped together, but I often see related, unexported functions directly after their related public function (like Add/add, Remove/remove). I woudn't be surprised to see a `count` function directly after `Count`. The reason why I thought it was "weird" to put `String` in between `Add` and `add` is because it seems to ignore this structure.

PR #596, opened 2017-05-31
334 rle

alanbernstein:


roaring/roaring.go
```diff
@@ -1267,12 +1267,33 @@ func (c *container) bitmapToRun() {
...
+		if v - c.array[i-1] > 1 {
+			// if current-previous > 1, one run ends and another begins
+			c.runs = append(c.runs, interval32{start, c.array[i-1]})
+			start = v
+		}
+	}
+	// append final run
+	c.runs = append(c.runs, interval32{start, c.array[i]})
+	c.array = nil
+	c.mapped = false
```
alanbernstein: I'm not exactly sure about the implications of this, I copied it from the existing conversion functions

jaffee: probably a good call

tgruben: i'm surprised that works, seems to me like the first iteration of that loop would throw an index out of bounds exception

roaring/roaring.go
```diff
@@ -1267,12 +1267,33 @@ func (c *container) bitmapToRun() {
...
+	start := c.array[0]
+	for i, v := range c.array[1:] {
+		if v - c.array[i] > 1 {
+			// if current-previous > 1, one run ends and another begins
+			c.runs = append(c.runs, interval32{start, c.array[i]})
+			start = v
+		}
+	}
+	// append final run
+	c.runs = append(c.runs, interval32{start, c.array[c.n-1]})
```
tgruben: I would prefer c.array[len(c.array)-1].  Though I guess we should trust c.n to be accurate.

alanbernstein: yea, if we can't trust c.n, we should address that rather than avoid it here

PR #599, opened 2017-06-01
Minor docs fix, indent in JSON

jnovinger:
Properly indent `Fields.Mappers.Max` field in example JSON.


jnovinger:
Wow, the indent thing just jumped out at me, but I totally missed the whole "should be valid JSON" thing. 

:neutral_face:

PR #606, opened 2017-06-06
Add sorting flag to import command.

### LINKED ISSUE #557: Make `sort` subcommand an option to `import`
### Expected behavior
`pilosa import --sort -i testindex -f testframe input.csv`

`-s` preferred, currently used for `--buffer-size`

### Actual behavior
`pilosa sort input.csv > temp.csv`
`pilosa import -i testindex -f testframe temp.csv`

benbjohnson:
Fixes #557


benbjohnson:
I did some benchmarking and I don't see any substantial difference in timing between `sort + import` vs `import --sort`. The only reason I can think of to have a `sort` command is if you want to generate a file and sort it once and then let people import that data. That way you only perform the sort once.

I'll defer to @travisturner's judgment about whether to keep `sort` or remove it.

travisturner:
i don't have a strong opinion, but my vote would be to simplify and remove the `pilosa sort`

benbjohnson:
Sounds good to me. We can always re-add it if it becomes necessary. I removed it in 345aba3.

alanbernstein:
Code looks good, can you update the docs to reflect the change? `pilosa sort` is mentioned in `administration.md` and `installation.md`, I think that's it.

benbjohnson:
Fixed in the docs in 1c34ea7.

Issue #607, opened 2017-06-06
Input Definition: Epic


travisturner:
## Description

The purpose of the Input Definition is to provide a way for a user to define
a schema based on their data. This allows the user to provide data to Pilosa in 
a more standard format (for example: json or csv) instead of writing custom application
code to do low-level operations like `SetBit()`.

Once a user has provided an Input Definition, then they can send data to Pilosa in
a format that adheres to that definition and Pilosa will internally perform all
of the appropriate mutations.

Issue #639, opened 2017-06-13
Input definition: Broadcast Delete


raskle:
### Description
Delete input definitions messages need to be Broadcasted to all nodes in the cluster.
Each node will contain the same copy of all input definitions.

We will also need a protobuf for the DeleteInputDefinitionMessage.

### Success criteria (What criteria will consider this ticket closeable?)
Existing input definitions will be deleted from all nodes in the cluster

Issue #638, opened 2017-06-13
Input definition: Broadcast handler Create


raskle:
### Description
New input definitions need to Broadcast a create message to all nodes in the cluster.
Each node will contain a copy of all input definitions

### Success criteria (What criteria will consider this ticket closeable?)
Each node will contain a copy of all input definitions

Issue #611, opened 2017-06-06
Input Definition: load input definitions from disk on startup


travisturner:
This is related to and dependent upon the work done in #610, and is part of the proposal described in #538.

On Pilosa startup, as part of loading indexes, check for any `.input-definitions` files and when present load them into memory so they are available for use.

PR #646, opened 2017-06-16
Input definition

linhvo:
## Overview

This PR cover #607

## Pull request checklist

## Code review checklist


linhvo:
@travisturner: Michael is still working on delete broadcast handler (#639) and adding more tests, but I still create a PR for your review of what we've done so far. It has already covered #608, #610, #611, #638

index.go
```diff
@@ -585,3 +611,113 @@ type importData struct {
...
+			return err
+		}
+		inputFiles, err := inputPath.Readdir(0)
+		for _, file := range inputFiles {
+			input, err := i.newInputDefinition(i.InputDefPath(), file.Name())
+			if err != nil {
+				return err
+			}
+			input.Open()
+			i.inputDefinitions[file.Name()] = input
```
travisturner: is there a way we can share some of the logic from `createInputDefinition` around frame creation such that on startup opening the inputDefinition will create any frames that don't yet exist?

index.go
```diff
@@ -585,3 +611,113 @@ type importData struct {
...
+
+	inputDef.LoadDefinition(pb)
+	if err = inputDef.saveMeta(); err != nil {
+		return nil, err
+	}
+	i.inputDefinitions[pb.Name] = inputDef
+	return inputDef, nil
+}
+
+func (i *Index) newInputDefinition(path, name string) (*InputDefinition, error) {
```
travisturner: will there ever be a case where `path` is not `i.InputDefPath()`? if not, then we can remove the `path` argument from this function signature and just replace `i.name` with `i.InputDefPath()` inside the function.

linhvo:
@travisturner: I removed WIP and it's ready for another review if necessary before sending another PR for /input handling

input_definition.go
```diff
@@ -0,0 +1,317 @@
...
+}
+
+func (i *InputDefinition) AddFrame(frame InputFrame) error {
+	i.frames = append(i.frames, frame)
+	if err := i.saveMeta(); err != nil {
+		return err
+	}
+	return nil
+}
+
```
raskle: Add comments to all public functions

handler.go
```diff
@@ -1495,3 +1498,134 @@ func errorString(err error) string {
...
+				continue
+			} else {
+				err = fmt.Errorf("primary field's name not match columnLabel")
+				http.Error(w, err.Error(), http.StatusBadRequest)
+				return
+			}
+		}
+	}
+	if numPrimaryKey > 1 {
+		err = errors.New("duplicate primaryKey with other field")
```
travisturner: let's make these errors part of the system error constants.
that way, when we check for them in tests, we don't have to compare the body of the error.

travisturner:
@linhvo @raskle the test coverage on the `input-definition` branch is pretty good with the exception of error return cases. you might review those and try to test for the more likely cases.

handler.go
```diff
@@ -1495,3 +1498,134 @@ func errorString(err error) string {
...
+	err := h.Broadcaster.SendSync(
+		&internal.DeleteInputDefinitionMessage{
+			Index: indexName,
+			Name:  inputDefName,
+		})
+	if err != nil {
+		h.logger().Printf("problem sending CreateInputDefinition message: %s", err)
+	}
+
+	if err := json.NewEncoder(w).Encode(postInputDefinitionResponse{}); err != nil {
```
linhvo: since they are both empty struct, I will change it to `defaultInputDefinitionResponse` if it sounds ok

travisturner:
@linhvo @raskle 
I refactored some of the `Encode()` logic to simplify it a bit, but we need to handle validation somewhere other than in `Encode()` (see #694). I don't think `Encode()` should return errors; we should check for errors prior to that. I think validation should happen either during or directly after the JSON decoding step.

PR #655, opened 2017-06-19
334 rle

jaffee:
DO NOT REVIEW THIS IN GITHUB
DO NOT REVIEW THIS IN GITHUB
DO NOT REVIEW THIS IN GITHUB

Rebase 334-rle onto master. This didn't go completely smoothly, but tests are passing. 

****YOU CAN'T TRUST THE FILES CHANGED VIEW IN GITHUB**** - each thing is compared to their common ancestor rather than each other - see https://stackoverflow.com/questions/32570825/github-comparison-view-for-2-branches-is-incorrect
We won't want to merge this directly, but rather force push it to upstream/334-rle

Fetch my fork locally, and then do `git diff upstream/334-rle origin/334-rle`.
Make sure that you've also fetched upstream.

You probably want to look mostly at `roaring.go` and `roaring_internal_test.go` - you can just append the file to the above command and it will just show you that file. I also made changes to `ctl/check_test.go` which you should take a look at.

### DIFF
--- a/ctl/check_test.go
+++ b/ctl/check_test.go
@@ -17,7 +17,6 @@ package ctl
 import (
 	"bytes"
 	"encoding/hex"
-	"golang.org/x/net/context"
 	"io"
 	"io/ioutil"
 	"math/rand"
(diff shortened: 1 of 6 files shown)


jaffee:
Pushed to 334-rle-rebased

Issue #591, opened 2017-05-31
Runtime error: makeslice: len out of range


jarutis:
Hi,

I tried to use pilosa with uint64 values, however it seems that the server stops working after I index a large value. I followed getting started tutorial with the following steps:

```
$ curl localhost:10101/index/repository  -X POST  -d '{"options": {"columnLabel": "repo_id"}}'
{}
$ curl localhost:10101/index/repository/frame/stargazer  -X POST -d '{"options": {"rowLabel": "stargazer_id",  "timeQuantum": "YMD", "inverseEnabled": true}}'
{}
$ curl localhost:10101/index/repository/frame/language  -X POST -d '{"options": {"rowLabel": "language_id", "inverseEnabled": true}}'
{}
```

Imported data from the tutorial, played with all works fine. But if I run:

```
$ curl localhost:10101/index/repository/query -X POST -d 'SetBit(frame="stargazer", repo_id=77777, stargazer_id=6018148517796380732)'
{"results":[true]}
$ curl localhost:10101/index/repository/query -X POST -d 'Bitmap(frame="stargazer", stargazer_id=6018148517796380732)'
curl: (52) Empty reply from server
```

And the server prints the following error:

```
2017/05/31 10:18:51 http: panic serving 127.0.0.1:46892: runtime error: makeslice: len out of range
goroutine 16 [running]:
net/http.(*conn).serve.func1(0xc420e51b80)
	/usr/local/go/src/net/http/server.go:1491 +0x12a
panic(0xa41ae0, 0xc428682d00)
	/usr/local/go/src/runtime/panic.go:458 +0x243
github.com/pilosa/pilosa.(*Executor).Execute(0xc420258690, 0xde71c0, 0xc420876840, 0xc4208766fc, 0xa, 0xc4254dd860, 0xc428682cf8, 0x1, 0x1, 0xc428682c20, ...)
	/Users/cody/gopath/src/github.com/pilosa/pilosa/executor.go:97 +0x1ba
github.com/pilosa/pilosa.(*Handler).handlePostQuery(0xc4201dc100, 0xde62c0, 0xc420ad7930, 0xc4208f1950)
	/Users/cody/gopath/src/github.com/pilosa/pilosa/handler.go:204 +0x34b
github.com/pilosa/pilosa.(*Handler).(github.com/pilosa/pilosa.handlePostQuery)-fm(0xde62c0, 0xc420ad7930, 0xc4208f1950)
[...]

Issue #593, opened 2017-05-31
Research ability to support very high slices without allocating for intermediate slices


jaffee:
### Description

In #591 we see that Pilosa can't handle extremely high column values, even when many intermediate columns aren't used.

### Success criteria (What criteria will consider this ticket closeable?)

Either we've changed the way that Pilosa tracks slices to allow for this kind of behavior, or we've decided to explicitly not support it, documented it, and returned reasonable errors to clients who try to do it (which #591 should account for).

benbjohnson:
I tried digging into this and #591 but there's several issues I've run into.

First, we can get around the initial `inverseSlices` panic by changing `slices []uint64` to a generator (e.g. `slices func() uint64`) that can stream all slice numbers instead of building a slice of slice numbers.

However, that still leaves us with 5,739,353,673,740 slices to process. We don't track which slices have data AFAIK so we'd have to check each one.

One option would be reducing the number of slices by increasing the slice width. We're currently at 1,048,576 so we'd have to increase it by quite a bit. But even if we increased slice width to `MaxUint32` then we'd still have `1,401,209,393` slices.

The only option I can think of that would make any sense would be to provide a translation table for `columnID` so that a `6018148517796380732` could be stored as `1` internally. Maybe the translation table could use strings so we could support hexidecimal UUIDs as well. That way we could support the user's value as `"6018148517796380732"` instead and it's just an opaque string.

@travisturner What do you think?

jaffee:
@benbjohnson being able to do that translation (for rows as well as columns actually) would be extremely valuable. This is definitely something we've discussed (at length), and basically haven't been confident enough to do. There are a lot of finicky bits(pun?) in the distributed case, and the translation needs to be fast in both directions to support ingestion and queries.

For this particular issue, I was wondering if we could do something similar to the way that we track which containers are in use in roaring with the slice of "keys" (see `Bitmap.insertAt` to jog your memory). I can tell, though, that we'll probably have to alter the query execution and fanout logic fairly significantly (or have strong consensus about which slices exist across the cluster). It may be that doing it this way isn't much easier than translating opaque strings.

TLDR; if you think doing that translation layer is feasible, I'd love to hear more.

travisturner:
@benbjohnson A translation table like you suggested sounds great, and we've talked about something like that for a while, but we have yet to come up with a solution that isn't too complex. There are several concerns, a couple of which are:
- ensuring that the lookup doesn't (too badly) affect performance
[...]

PR #688, opened 2017-06-27
Catch make() panic in executor

alanbernstein:
## Overview

Prevents an empty response after panic inside make(), when a query is sent with an ID too large for all intermediate slice indices to fit in memory.

Addresses #591

## Pull request checklist

## Code review checklist

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -62,7 +62,16 @@ func NewExecutor() *Executor {
 }
 
 // Execute executes a PQL query.
-func (e *Executor) Execute(ctx context.Context, index string, q *pql.Query, slices []uint64, opt *ExecOptions) ([]interface{}, error) {
+func (e *Executor) Execute(ctx context.Context, index string, q *pql.Query, slices []uint64, opt *ExecOptions) (results []interface{}, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			if r.(error).Error() == "runtime error: makeslice: len out of range" {
+				// Catch panic from make([]uint64, veryLargeNumber).
+				results, err = nil, ErrSliceIndexTooLarge
+			}
+		}
+	}()
+
 	// Verify that an index is set.
 	if index == "" {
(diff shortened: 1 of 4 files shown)


executor.go
```diff
@@ -62,7 +62,16 @@ func NewExecutor() *Executor {
 }
 
 // Execute executes a PQL query.
-func (e *Executor) Execute(ctx context.Context, index string, q *pql.Query, slices []uint64, opt *ExecOptions) ([]interface{}, error) {
+func (e *Executor) Execute(ctx context.Context, index string, q *pql.Query, slices []uint64, opt *ExecOptions) (results []interface{}, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			if r.(error).Error() == "runtime error: makeslice: len out of range" {
```
alanbernstein: Is there a better way to identify runtime errors? I don't know if the message string is stable across versions/platforms.

jaffee: better to be too specific than too general I think. We wouldn't want to report the wrong thing back.

alanbernstein:
I was going to also catch all unknown panics, but I made a separate issue for that: https://github.com/pilosa/pilosa/issues/707. So this PR is ready, but would be redundant when #708 is implemented.

jaffee:
Should we just hold off on merging this and assuming all goes well in #707 use that instead?

alanbernstein:
I think we should hold off but keep it open for now. For this particular error, it might be worthwhile to catch it and show some explanatory message. Otherwise I think the github issue is the only clue anyone would have about why it doesn't work. If https://github.com/pilosa/pilosa/issues/593 turns out well this won't be necessary.

alanbernstein:
@jaffee do you think this is worth keeping alongside the general recover?

jaffee:
I don't think it's worth keeping - recover clutters the code a lot. Do you see any significant benefits?

jaffee:
I don't think it's worth keeping around

PR #696, opened 2017-06-28
Validate incoming input definitions through the InputDefinitionInfo V…

raskle:
## Overview
Validate incoming input definitions through the InputDefinitionInfo Validate rather than through Encode and piecemeal throughout the code.

## Code review checklist


handler.go
```diff
@@ -1520,34 +1520,16 @@ func (h *Handler) handlePostInputDefinition(w http.ResponseWriter, r *http.Reque
 		return
 	}
 
-	// TODO: validation before/after encode?
+	// validation definition before/after encode?
```
linhvo: This line can be removed?

input_definition.go
```diff
@@ -254,6 +265,56 @@ type InputDefinitionInfo struct {
...
 }
 
+// Validate the InputDefinitionInfo data
+func (i *InputDefinitionInfo) Validate(columnLabel string) error {
+	numPrimaryKey := 0
+	accountRowID := make(map[string]uint64)
+
+	if len(i.Frames) == 0 {
+		return fmt.Errorf("At least one frame is required per Input Definition")
+	}
```
linhvo: as discussed with Travis before, "frames and fields are both required (i suppose fields cold be empty, but in that case it would only be used to auto-create frames on startup). maybe for now we make them both required, and if we want to relax that later we can". I did that in createInputDefinition func (line 660-662 in index.go), but since we check it here, I think those lines can be remove and use ErrInputDefinitionAttrsRequired here insead of `fmt.Errorf("At least one frame is required per Input Definition")`.

input_definition.go
```diff
@@ -254,6 +265,56 @@ type InputDefinitionInfo struct {
...
+
+	for _, frame := range i.Frames {
+		if err := frame.Validate(); err != nil {
```
linhvo: if the Validate func (line 246-253) only call ValidateName func, does calling ValidateName(frame.Name) itself here has the same result without adding extra code:
```
if err := ValidateName(frame.Name); err != nil {
			return err
		}
```

input_definition.go
```diff
@@ -254,6 +265,56 @@ type InputDefinitionInfo struct {
...
+			numPrimaryKey++
+			if field.Name != columnLabel {
+				return ErrInputDefinitionColumnLabel
+			}
+		}
+		for _, action := range field.Actions {
+			if err := action.Validate(); err != nil {
+				return err
+			}
+			if action.ValueDestination == InputSingleRowBool && action.Frame != "" {
```
linhvo: I think we can remove `&& action.Frame != ""` since action.Validate() return ErrFrameRequired if frame is empty

input_definition.go
raskle: I'd like to keep the validate in the input frame.  Then we can add a FrameOption validate and do that in one place.

travisturner: note that if you do keep the `InputFrame.Validate()` method, it can be simplified to:
```
// Validate the InputFrame data
func (i *InputFrame) Validate() error {
    return ValidateName(i.Name)
}
```
With that said, you may be considering adding a call to the frame option validation in this method as well, which would cause this method to change anyway.

input_definition.go
```diff
@@ -209,19 +197,34 @@ type Action struct {
...
-	// TODO: this check needs to happen somewhere other than Encode()
-	/*
-		if o.RowID == nil && o.ValueDestination == InputSingleRowBool {
-			return nil, errors.New("rowID required for single-row-boolean")
+// Validate ensures the input definition action conforms to our specification.
+func (a *Action) Validate() error {
+	if a.Frame == "" {
+		return ErrFrameRequired
+	}
+	validValues := make(map[string]bool)
```
travisturner: it may be easier to read code that uses a function like
```
func stringInSlice(a string, list []string) bool {
    for _, b := range list {
        if b == a {
            return true
        }
    }
    return false
}
```
instead of using a map like `validValues` just for comparison.

that way, you just need:
```
if !stringInSlice(a.ValueDestination, validValueDestination) {
	return fmt.Errorf(...)
}
```

handler_test.go
```diff
@@ -1299,6 +1318,13 @@ func TestHandler_GetInputDefinition(t *testing.T) {
 	} else if body := w.Body.String(); body != string(expect)+"\n" {
 		t.Fatalf("unexpected body: %s, expect: %s", body, string(expect))
 	}
+
+	// Check non existant definition
```
travisturner: `nonexistent` and a period.

PR #702, opened 2017-06-30
Set & retrieve field values.

### LINKED ISSUE #617: BSI Range-Encoding: internal support for setting field values
Given a `frame`, `columnID`, `field`, and `value`, look up the field in the schema for the frame and set the appropriate bits for the field (probably need to set both 1's and 0's)

In the example in #616, the following function:
```
frame.setFieldValue(colID uint64, field string, val int64) {}
```
called with:
```
setFieldValue(1, "age", 35)
```
Would perform the equivalent of the following (note that we are setting the `flip` of 35 in 7-bit binary (`0100011` to `1011100`):
```
SetBit(frame="f", columnID=1, rowID=0)
ClearBit(frame="f", columnID=1, rowID=1)
ClearBit(frame="f", columnID=1, rowID=2)
SetBit(frame="f", columnID=1, rowID=3)
SetBit(frame="f", columnID=1, rowID=4)
SetBit(frame="f", columnID=1, rowID=5)
ClearBit(frame="f", columnID=1, rowID=6)
SetBit(frame="f", columnID=1, rowID=7)
```

We should be aware that by having `val` be an `int64` we are limiting ourselves to 64-bit range-encoded values when technically we don't have a limit with the implementation.

A basic POC can be found at https://github.com/travisturner/bit-slice-index

benbjohnson:
## Overview

This pull request adds support for set/get range values at the `Frame`, `View`, and `Fragment` levels.

Fixes #617.

## Pull request checklist

## Code review checklist

- [x] Ensure that any changes to external docs have been included in this pull request.
- [x] If the changes require that minor/major versions need to be updated, tag the PR appropriately.
- [x] Ensure the new code is [properly commented](https://github.com/golang/go/wiki/CodeReviewComments#doc-comments) and follows [Idiomatic Go](https://dmitri.shuralyov.com/idiomatic-go).
- [x] Check that tests have been written and that they cover the new functionality.
- [x] Run tests and ensure they pass.
- [x] Build and run the code, performing any applicable integration testing.

### DIFF
--- a/fragment.go
+++ b/fragment.go
@@ -472,6 +472,69 @@ func (f *Fragment) clearBit(rowID, columnID uint64) (changed bool, err error) {
 	return changed, nil
 }
 
+func (f *Fragment) bit(rowID, columnID uint64) (bool, error) {
+	pos, err := f.pos(rowID, columnID)
+	if err != nil {
+		return false, err
+	}
+	return f.storage.Contains(pos), nil
+}
+
+// FieldValue uses a column of bits to read a multi-bit value.
+func (f *Fragment) FieldValue(columnID uint64, bitDepth uint) (value uint64, exists bool, err error) {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	// If existance bit is unset then ignore remaining bits.
+	if v, err := f.bit(uint64(bitDepth), columnID); err != nil {
+		return 0, false, err
+	} else if !v {
+		return 0, false, nil
+	}
+
+	// Compute other bits into a value.
+	for i := uint(0); i < bitDepth; i++ {
+		if v, err := f.bit(uint64(i), columnID); err != nil {
+			return 0, false, err
+		} else if !v {
+			value |= (1 << i)
+		}
+	}
+
+	return value, true, nil
+}
+
+// SetFieldValue uses a column of bits to set a multi-bit value.
+func (f *Fragment) SetFieldValue(columnID uint64, bitDepth uint, value uint64) (changed bool, err error) {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	for i := uint(0); i < bitDepth; i++ {
+		if value&(1<<i) != 0 {
+			if c, err := f.clearBit(uint64(i), columnID); err != nil {
+				return changed, err
+			} else if c {
+				changed = true
+			}
+		} else {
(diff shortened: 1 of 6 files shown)


frame_test.go
```diff
@@ -68,6 +68,153 @@ func TestFrame_SetTimeQuantum(t *testing.T) {
...
+		}
+
+		// Set value on first field.
+		if changed, err := f.SetFieldValue(100, "field0", 21); err != nil {
+			t.Fatal(err)
+		} else if !changed {
+			t.Fatal("expected change")
+		}
+
+		// Set value on on same column but different field.
```
travisturner: duplicate "on" in the comment.

fragment.go
```diff
@@ -472,6 +472,69 @@ func (f *Fragment) clearBit(rowID, columnID uint64) (changed bool, err error) {
...
+
+	return value, true, nil
+}
+
+// SetFieldValue uses a column of bits to set a multi-bit value.
+func (f *Fragment) SetFieldValue(columnID uint64, bitDepth uint, value uint64) (changed bool, err error) {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	for i := uint(0); i < bitDepth; i++ {
```
travisturner: For base-2 range-encoding, we only have to store the 0th bit vector, which ends up being the "flip" of the actual value. So if value = 20, base-2 = `10100` and base-2 range-encoded = `01011`. All that to say I think the set and clear here need to be reversed. That would change the logic in `FieldValue()` as well.

frame_test.go
benbjohnson: Fixed in 800e384.

fragment.go
benbjohnson: Right, I fixed the flip in 800e384. I don't quite understand the need for the flip though. Why not store in "unflipped" format?

travisturner: It makes the range comparisons easier. For example, if you want to find all B greater than F, starting with the most significant bit of a range-encoded bitmap (bits flipped) you can do a `NOR` operation:
```
(range-encoded case where B is definitely greater than F):
B 0 0 1 1
F 0 1 0 1
---------
  T - - -
```

If you store it as the actual value
```
(equality encoded case where B is definitely greater than F):
B 0 0 1 1
F 0 1 0 1
---------
  - - T -
```
it's not a straightforward bitwise operation.

PR #714, opened 2017-07-07
update config to match with json tag

linhvo:
Fixes #
pilosa server can run with default config
## Pull request checklist

## Code review checklist

### DIFF
--- a/config.go
+++ b/config.go
@@ -39,17 +39,17 @@ const (
 // Config represents the configuration for the command.
 type Config struct {
 	DataDir string `toml:"data-dir"`
-	Host    string `toml:"bind"`
+	Bind    string `toml:"bind"`
 
 	Cluster struct {
-		ReplicaN        int      `toml:"replicas"`
-		Type            string   `toml:"type"`
-		Hosts           []string `toml:"hosts"`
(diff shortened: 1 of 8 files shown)


linhvo:
I agree, I just updated the json tag that I think the simplest way to fix thing if we don't care about struct/json consistent name

config.go
```diff
@@ -39,14 +39,14 @@ const (
 // Config represents the configuration for the command.
 type Config struct {
 	DataDir string `toml:"data-dir"`
-	Host    string `toml:"host"`
+	Host    string `toml:"bind"`
```
jaffee: I think we ought to rename the struct fields as well as the tags.

linhvo: I saw `Host` is used by `pilosa/tools/creator`, is changing it break tools? I'm just not sure if any other app calls Config.Host directly might be broken from the changing struct's field

jaffee: Good call - I didn't think of that. I'm refactoring tools to try not to import Pilosa at all. Since it isn't open source yet, I say let's not let that stop us. If the pdk were using it, that would be a bigger issue (I haven't checked),  but since everything is vendored, changes shouldn't break anything.

linhvo: I see, I'll rename the struct fields then

PR #733, opened 2017-07-18
Implements view import

yuce:
## Overview

This PR adds support for importing raw views. This is required to be able to import data exported by `/export` endpoint. I've modified `ImportRequest` protobuf message to carry a `View` field, but I can add it as a query arg too in case changes to the protobuf code is not desired.

`/import` endpoint is not publicly documented.

## Pull request checklist

## Code review checklist


yuce:
Added the tests; this PR should be OK to review.

yuce:
PING

travisturner:
@yuce can you provide me with an example where this applies so i can understand the use case and help test it.

yuce:
@travisturner The main reason this is needed is, exporting bits (`/export` endpoint) requires a view and returns `row_id, column_id` for each bit. Importing bits on the other hand (`/import`) requires a frame and `row_id,column_id[,timestamp]` so exported data cannot be used for imports. This PR implements the missing *view* bit, so the user can export bits for a view and import the same data.

travisturner:
@yuce i guess i'm trying to understand which view would typically be specified for an import. I don't think it makes sense to specify `inverse` because an inverse view is going to be automatically created/populated when you import data from the `standard` view (we would have to suppress that functionality on `standard` import if we wanted to specifically handle `inverse` view population via an import. The other views are related to time quantums. Is that what you're trying to address here?

yuce:
I didn't notice the inverse view would be automatically created on import. We should have a way to suppress that. Yes, we should import time quantum views explicitly or update `/export` endpoint so it returns timestamps.

yuce:
Having said that, I think the ability to import views is a more future-proof solution; in case we have different kinds of views in the future.

jaffee:
closing for now - referenced in #888

PR #748, opened 2017-07-25
Attempt to set open file limit during Pilosa startup. Fixes #722.

### LINKED ISSUE #722: Investigate and implement use of syscall.Setrlimit to change open file limit.
### Description
Pilosa needs a fairly high open file limit to operate in many cases. This is a common sticking point for new users, and can also be an annoyance for experienced users. Go's syscall package provides some OS dependent functions for setting the open file limit from within the process - a small example that seems to work on OSX is: https://play.golang.org/p/j5ni_IihCw
The meat of this is 
```
lim := &syscall.Rlimit{}
err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, lim)
err = syscall.Setrlimit(syscall.RLIMIT_NOFILE, &syscall.Rlimit{Cur: lim.Max, Max: lim.Max})
```

We should try implementing this, and make sure it works on darwin and linux. We could try compiling for BSD and other systems as well, and if it doesn't work, we can use file suffixes (i.e. file_darwin.go) to exclude the functionality for those platforms. (I think that will work, but we'll need to confirm)

The logic here will be something like:
1. Check current limit - if unable to check, log a message notifying the users. If limit is high enough (200k?) we're done. Otherwise...
2. Try to set the limit higher (we've been using 262144 - alternatively use `lim.Max` as in the example. If there's an error, log out that the limit may not be high enough, and what the error was setting it.
3. Check to see what the limit was set to. If high enough, good, otherwise log out again that we weren't able to set it high enough.

Pilosa tries its best to set the open file limit high enough, logs clear messages describing the situation if it isn't able to, and gives the user something searchable to get further instructions for setting open file limits.

codysoyland:
## Overview

Attempt to set open file limit during Pilosa startup.

Fixes #722

## Pull request checklist

## Code review checklist

### DIFF
--- a/docs/administration.md
+++ b/docs/administration.md
@@ -32,7 +32,7 @@ While Pilosa does have some high system requirements it is not a best practice t
 
 ### Open File Limits
 
-Pilosa requires a large amount of open files to support its memory-mapped file storage system. Most operating systems put limits on the maximum number of files that may be opened concurrently by a process. On Linux systems, this limit is controlled by a utility called [ulimit](https://ss64.com/bash/ulimit.html). Pilosa will automatically attempt to raise the limit to `262144` during startup, but it may fail due to access limitations. If you see errors related to open file limits when starting Pilosa, it is recommended that you run `sudo ulimit -n 262144` before starting Pilosa.
+Pilosa requires a large number of open files to support its memory-mapped file storage system. Most operating systems put limits on the maximum number of files that may be opened concurrently by a process. On Linux systems, this limit is controlled by a utility called [ulimit](https://ss64.com/bash/ulimit.html). Pilosa will automatically attempt to raise the limit to `262144` during startup, but it may fail due to access limitations. If you see errors related to open file limits when starting Pilosa, it is recommended that you run `sudo ulimit -n 262144` before starting Pilosa.
 
 On Mac OS X, `ulimit` does not behave predictably. [This blog post](https://blog.dekstroza.io/ulimit-shenanigans-on-osx-el-capitan/) contains information about setting open file limits in OS X.
 
--- a/docs/administration.md
+++ b/docs/administration.md
@@ -30,6 +30,12 @@ Pilosa is designed to be a distributed application, with data replication shared
 
 While Pilosa does have some high system requirements it is not a best practice to set up a cluster with the fewest, largest machines available.  You want an evenly distributed load across several nodes in a cluster to easily recover from a single node failure, and have the resource capacity to handle a missing node until it's repaired or replaced.   Nor is it advisable to have many small machines.  The internode network traffic will become a bottleneck.  You can always add nodes later, but that does require some down time.
 
+### Open File Limits
+
+Pilosa requires a large amount of open files to support its memory-mapped file storage system. Most operating systems put limits on the maximum number of files that may be opened concurrently by a process. On Linux systems, this limit is controlled by a utility called [ulimit](https://ss64.com/bash/ulimit.html). Pilosa will automatically attempt to raise the limit to `262144` during startup, but it may fail due to access limitations. If you see errors related to open file limits when starting Pilosa, it is recommended that you run `sudo ulimit -n 262144` before starting Pilosa.
+
+On Mac OS X, `ulimit` does not behave predictably. [This blog post](https://blog.dekstroza.io/ulimit-shenanigans-on-osx-el-capitan/) contains information about setting open file limits in OS X.
+
 ### Importing and Exporting Data
 
 #### Importing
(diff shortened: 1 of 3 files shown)


alanbernstein:
Would it appropriate to add tests that call this with different limits?

holder.go
```diff
@@ -351,6 +359,42 @@ func (h *Holder) flushCaches() {
...
+	var setLimit uint64
+	if err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, lim); err != nil {
+		h.logger().Printf("ERROR checking open file limit: %s", err)
+		return
+	}
+	if lim.Max < FileLimit {
+		h.logger().Printf("WARNING: open file limit max is %d. Please consider running \"ulimit -n %d\" before starting Pilosa to avoid \"too many open files\" error.", lim.Max, FileLimit)
+	}
+	// If the soft limit is lower than both the FileLimit constant and the hard limit, we will try to change it.
+	if lim.Cur < FileLimit && lim.Cur < lim.Max {
```
jaffee: I think it is possible to try to set both the soft and hard limits, so it might be worth trying to set lim.Max if it's below our target value.

holder.go
```diff
@@ -351,6 +359,42 @@ func (h *Holder) flushCaches() {
...
+// setFileLimit attempts to set the open file limit to the FileLimit constant defined above.
+func (h *Holder) setFileLimit() {
+	lim := &syscall.Rlimit{}
+	var setLimit uint64
+	if err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, lim); err != nil {
+		h.logger().Printf("ERROR checking open file limit: %s", err)
+		return
+	}
+	if lim.Max < FileLimit {
+		h.logger().Printf("WARNING: open file limit max is %d. Please consider running \"ulimit -n %d\" before starting Pilosa to avoid \"too many open files\" error.", lim.Max, FileLimit)
```
jaffee: I think we need better help messages than `ulimit -n` especially in the case below where the OS accepted the syscall, but silently refused to set the limit - the user is probably on OSX and will need better direction to fix the problem. This post https://blog.dekstroza.io/ulimit-shenanigans-on-osx-el-capitan/ describes the arduous journey to actually changing the open file limit on el capitan, I'm not sure if it works on Sierra as well, but I think we should at least try to give the user some place to look further if they are on OSX.

codysoyland: Maybe I can add a section to the documentation about these caveats and provide the URL to that. Also, I have no idea how this works in other OSs (Windows and BSD), so just describing the problem in more detail would be good to help people along.

holder.go
codysoyland: Ok, I can give that a try. The hard limit is only mutable by root, but might as well at least try.

jaffee:
ah yea, we do need to check that this doesnt fail to compile/run on other
platforms we support

docs/administration.md
```diff
@@ -30,6 +30,12 @@ Pilosa is designed to be a distributed application, with data replication shared
 
 While Pilosa does have some high system requirements it is not a best practice to set up a cluster with the fewest, largest machines available.  You want an evenly distributed load across several nodes in a cluster to easily recover from a single node failure, and have the resource capacity to handle a missing node until it's repaired or replaced.   Nor is it advisable to have many small machines.  The internode network traffic will become a bottleneck.  You can always add nodes later, but that does require some down time.
 
+### Open File Limits
+
+Pilosa requires a large amount of open files to support its memory-mapped file storage system. Most operating systems put limits on the maximum number of files that may be opened concurrently by a process. On Linux systems, this limit is controlled by a utility called [ulimit](https://ss64.com/bash/ulimit.html). Pilosa will automatically attempt to raise the limit to `262144` during startup, but it may fail due to access limitations. If you see errors related to open file limits when starting Pilosa, it is recommended that you run `sudo ulimit -n 262144` before starting Pilosa.
```
jaffee: I would say "a large number" rather than a large amount.

jaffee:
build fails on freebsd and dragonfly due to slightly different syscall interface it looks like:
```
[11:15:04]~/g/s/g/p/pilosa git:722-ulimit ❯❯❯ GOOS=freebsd go build ./cmd/pilosa
# github.com/pilosa/pilosa
./holder.go:372: invalid operation: oldLimit.Cur < FileLimit (mismatched types int64 and uint64)
./holder.go:373: cannot use FileLimit (type uint64) as type int64 in assignment
./holder.go:375: invalid operation: oldLimit.Max < FileLimit (mismatched types int64 and uint64)
./holder.go:376: cannot use FileLimit (type uint64) as type int64 in assignment
./holder.go:404: invalid operation: oldLimit.Cur < FileLimit (mismatched types int64 and uint64)
```

codysoyland:
Good call on the FreeBSD build failure. I committed a patch that removes the type declaration from the const, allowing it to build successfully on FreeBSD. I did not, however, attempt to execute the binary on FreeBSD.

PR #773, opened 2017-08-15
Countrange bug

### LINKED ISSUE #770: bug in Bitmap.CountRange
Imported some data which only had 3 columns, but 10s of millions of rows. A few rows have 2 or 3 bits set, but most have just 1.

### Expected behavior
A TopN query will return 3 for some rows, followed by some 2s and lots of 1s.

### Actual behavior
TopN query returns 6s and 5s. Changing line (approx.) 277 in fragment.go from 

		n := f.storage.CountRange(id*SliceWidth, (id+1)*SliceWidth)
to

		n := f.row(id, false, false).Count()
stops the problem from occurring.

Todd and I have a data file (77M zipped) which can be shared if necessary.

### Information about your environment (OS/architecture, CPU, RAM, cluster/solo, configuration, etc.
Mid 2014 Macbook Pro 16GB single node.

jaffee:
## Overview

A few bugs in Bitmap.CountRange code. Both occurred in the case where `search64` returned a negative number because the container key was not found in the list of container keys. The negative number returned is `-1 * (<index where the number should be> + 1)` That  `+ 1` was not being taken into account in the CountRange method, although it is taken into account at other locations where `search64` is used in roaring.go. The reason `search64` doesn't just return the negative index is because the index might be 0 in which case there would be no way to distinguish between "value was found at position 0" and "value was not found but would be located at position 0" 

Fixes #770 and one other bug which was discovered during the process but not ticketed.

## Pull request checklist

## Code review checklist

### DIFF
--- a/roaring/roaring.go
+++ b/roaring/roaring.go
@@ -197,6 +197,7 @@ func (b *Bitmap) CountRange(start, end uint64) (n uint64) {
 	if len(b.keys) == 0 {
 		return
 	}
+
 	skey := highbits(start)
 	ekey := highbits(end)
 
@@ -233,7 +234,11 @@ func (b *Bitmap) CountRange(start, end uint64) (n uint64) {
 
 	// Count containers in between.
 	for x := i + 1; x < j; x++ {
+		if ekey < b.keys[x] {
+			break
+		}
 		n += uint64(b.containers[x].n)
+
 	}
 
 	return n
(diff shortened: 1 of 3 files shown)


alanbernstein:
@tgruben and I discussed changing the search function signature (maybe return "found" bool plus index) to avoid this kind of issue, but neither of us thought it was worth the risk/hassle of updating all the uses of the functions.

jaffee:
@alanbernstein I agree with not changing those functions, but I updated the godocs at least to make it more clear what they're doing.

Issue #623, opened 2017-06-07
BSI Range-Encoding: HTTP endpoint to add/delete a field


travisturner:
add
- `curl -X POST /index/i/frame/f/field/x -d '{"type": "int", "min": 1, "max": 120}'`
- returns an error if the field already exists

delete
- this can be a lower priority
- `curl -X DELETE /index/i/frame/f/field/x`
- internally this should mark the field as deleted, but since we can't really re-use the rows allocated to that field (unless we completely cleared them), we'll just leave the data as is

travisturner:
Because fields are stored in their own view, PR #785 is likely related to the delete portion of this issue. I believe the logic in that PR is restricted to frames with a TimeQuantum, so that would need to be changed.

PR #785, opened 2017-08-22
Delete view API

raskle:
## Overview
An API to delete views from frames with a time quantum.

## Pull request checklist

## Code review checklist


frame_test.go
```diff
@@ -307,3 +307,36 @@ func TestFrame_RowLabelValidation(t *testing.T) {
...
+	err = f.DeleteView("v")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if f.View(viewName) != nil {
+		t.Fatal("view still exists in frame")
+	}
+
+	// // Retrieve existing view.
```
travisturner: I don't understand what this portion of the test is checking. Can you clarify in the comment?

handler.go
```diff
@@ -43,6 +44,7 @@ import (
 
 	"unicode"
 
+	// Allow building Pilosa without the web UI.
```
travisturner: is this a TODO?

handler.go
```diff
@@ -789,6 +792,50 @@ func (h *Handler) handleGetFrameViews(w http.ResponseWriter, r *http.Request) {
...
+
+	// Retrieve frame.
+	f := h.Holder.Frame(indexName, frameName)
+	if f == nil {
+		http.Error(w, ErrFrameNotFound.Error(), http.StatusNotFound)
+		return
+	}
+
+	// Check the frame for time quantum.
+	if f.TimeQuantum() == "" {
```
travisturner: I'm just calling out that this restriction may be a problem for #623 since fields are stored as views.

test/holder.go
```diff
@@ -44,11 +44,8 @@ func (h *Holder) Close() error {
 }
 
 // Reopen closes the holder and instantiates and opens a new holder.
+// note that the hold most be Closed first
```
travisturner: What was the reason for removing the `Close()` from this `Reopen()` method?
With this change it no longer agrees with the original comment.

fragment.go
```diff
@@ -293,11 +293,13 @@ func (f *Fragment) close() error {
 	// Flush cache if closing gracefully.
 	if err := f.flushCache(); err != nil {
 		f.logger().Printf("fragment: error flushing cache on close: err=%s, path=%s", err, f.path)
+		return err
```
raskle: Hi Travis I spoke with @benbjohnson handling fragment `Close()` and he said he couldn't think of any reason why it shouldn't be handling the errors.  While I agree it probably does not make much difference during server shutdown, I think we should handle them during delete operations.    Unless you think otherwise I'll file tickets to handle the other `Close()` function errors.

frame.go
```diff
@@ -401,7 +401,9 @@ func (f *Frame) Close() error {
 
 	// Close all views.
 	for _, view := range f.views {
-		_ = view.Close()
+		if err := view.Close(); err != nil {
```
raskle: @benbjohnson did not recall if this was intentional and thought we should address it.

handler.go
raskle: I was just addressing a golint warning, and thought it would be helpful to comment why these blank imports exist

test/holder.go
raskle: @travisturner Thanks I will adjust function comment to reflect the changed behavior.  Previously this method was swallowing errors on `Close()`.  To accomplish the goals of the initial test I thought it was better to run `Close()` as a preceding step, handling any error.  Then make file permission changes, that would later fail on `Reopen()`

fragment.go
benbjohnson: I agree that we should handle them. I don't remember why they were previously ignored. It doesn't make a lot of difference in server shutdown but it can be helpful if we're spinning up multiple instances during tests or if someone embeds pilosa inside their application for some reason.

handler.go
raskle: The current `(f *Frame) DeleteView` function implicitly manages deleting the standard and inverse `Views`.  So the `Field` based views are a problem for the current implementation.  I'm thinking of changing the to explicitly specify the full view name i.e. standard_2016010215 and inverse_ 2016010215.  This will enable the Field delete to use the same interface.  Plus the time window retention will manage this internally so this public API will not be needed for it.

handler.go
```diff
@@ -789,6 +792,44 @@ func (h *Handler) handleGetFrameViews(w http.ResponseWriter, r *http.Request) {
...
+
+	// Retrieve frame.
+	f := h.Holder.Frame(indexName, frameName)
+	if f == nil {
+		http.Error(w, ErrFrameNotFound.Error(), http.StatusNotFound)
+		return
+	}
+
+	// Delete the view.
+	if err := f.DeleteView(viewName); err != nil {
```
travisturner: There is a potential issue here. In a multi-node cluster, when a view does not reside on all nodes then a delete request sent to a node might return `invalid view` when the view is actually valid, there just isn't data for that view on that node.

You can see this by running the following in a 3-node cluster:
```
PIL="http://localhost:10101"
curl -X POST "$PIL/index/i" -d '{"options":{"columnLabel":"col"}}'
curl -X POST "$PIL/index/i/frame/f" -d '{"options":{"rowLabel":"row"}}'
curl -X POST "$PIL/index/i/query" -d 'SetBit(frame=f, row=1, col=100)'
```
results in the following:
```
├── pil0
│   └── i
│       └── f
│           └── .schema
├── pil1
│   └── i
│       └── f
│           └── .schema
└── pil2
    └── i
        └── f
            ├── .schema
            └── views
                └── standard
                    └── fragments
                        └── 0
```
Sending a DELETE view request to either node0 or node1 will not have the expected result.

We could leave it the way it is, and just make it a requirement that DELETE view requests must be sent to all nodes in order to guarantee a delete, or we could change this error so that it just logs but doesn't stop processing (and broadcasts the delete message to other nodes).

PR #803, opened 2017-08-30
Http Connection limit

### LINKED ISSUE #760: Cap the number of concurrent http connections
### Description
Pilosa will crash all the available file handles for that process are consumed.  One consumer of those file handles is an http connection.  If Pilosa could prohibit new connections once a threshold had been exceeded, a crash could be prohibited.

Pilosa server does not crash as a result of too many connections
Returns error when capacity has been exceeded

tgruben:
## Overview

This provides a path of safely limiting connections to pilosa should it be a limited file handle environment. The limit is enabled by adding a non-zero value to max-client-connections in the config file.  If the value is 0 or not present, no limit is imposed.
Fixes #760

## Pull request checklist

## Code review checklist


limit.go
```diff
@@ -0,0 +1,75 @@
...
+	"io"
+	"net"
+	"time"
+)
+
+var (
+	ErrorTooManyConnections = errTooManyConnections{}
+)
+
+func NewBoundListener(maxActive int, l net.Listener) net.Listener {
```
travisturner: I don't think there's any rules here, but I would put `l net.Listener` as the first argument to the function.

limit.go
```diff
@@ -0,0 +1,75 @@
+package pilosa
```
travisturner: we should rename this file to something a little more specific and related to its content.
perhaps `listener.go`. what do you think?

limit.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+type boundConn struct {
+	net.Conn
+	active chan struct{}
+}
+type errTooManyConnections struct {
+}
+
+func (e errTooManyConnections) Error() string   { return "To Many Connections" }
```
travisturner: Typo in "Too".
Also, I don't think `Many Connections` should be capitalized.
i.e. `Too many connections`

limit.go
```diff
@@ -0,0 +1,75 @@
...
+	case <-l.active:
+		c, err := l.Listener.Accept()
+		if err != nil {
+			l.active <- struct{}{}
+			return nil, err
+		}
+		return &boundConn{c, l.active}, err
+	default: //out of connections
+		c, _ := l.Listener.Accept()
+		l.writeError(c, "Too Many Connections")
```
travisturner: can you use `ErrorTooManyConnections.Error()` here instead of repeating the string?

config.go
```diff
@@ -41,6 +41,9 @@ const (
 
 	// DefaultMaxWritesPerRequest is the default number of writes per request.
 	DefaultMaxWritesPerRequest = 5000
+
+	// DefaultMaxClientConnections is zero, which means no connection limit.
+	DefaultMaxClientConnections = 0
```
travisturner: Do you think there's a reason to make the default something other than 0?
I ask because if something is driving this (like open file handles), should we protect against that in a default installation?

tgruben: The only thing I could think of was that if you were not experiencing a problem, there is really no need to add the complexity.  Other than that, I have no opinion

travisturner: Ok with me. Just wanted to check.

limit.go
tgruben: hmm...wonder what value adds.

travisturner: I don't think it adds any value. It just seems that logically if you're writing a wrapper you would structure it like `ObjWrapper(obj, arg1, arg2)`, and if you later added another argument to your Wrapper, you would just add it to the end of the list. But again, other than what amounts to me being picky, I don't think it adds any value.

travisturner:
Looks good to me. 
I'll leave this PR unapproved so we can add updated documentation and tests to it.

limit_listener.go
```diff
@@ -0,0 +1,75 @@
...
+	l.active <- struct{}{}
+	return err
+}
+func (l *boundListener) writeError(w io.Writer, msg string) {
+	date := time.Now()
+	var dst []byte
+	dst = date.In(time.UTC).AppendFormat(dst, time.RFC1123)
+	copy(dst[len(dst)-3:], "GMT")
+
+	fmt.Fprintf(w, "HTTP/1.1 503 Service Unavailable\r\n"+
```
jaffee: It probably isn't safe to assume HTTP/1.1 here...?

PR #811, opened 2017-09-01
Field HTTP API

### LINKED ISSUE #623: BSI Range-Encoding: HTTP endpoint to add/delete a field

benbjohnson:
## Overview

Provides an HTTP API for creating new fields on an existing frame.

Fixes #623

## Pull request checklist

## Code review checklist

### DIFF
--- a/frame.go
+++ b/frame.go
@@ -419,6 +419,58 @@ func (f *Frame) Field(name string) *Field {
 	return nil
 }
 
+// CreateField creates a new field on the schema.
+func (f *Frame) CreateField(field *Field) error {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	// Ensure frame supports fields.
+	if f.rangeEnabled {
+		return ErrFrameFieldsNotAllowed
+	}
+
+	// Copy schema and append field.
+	schema := f.schema.Clone()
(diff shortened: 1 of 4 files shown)


benbjohnson:
@travisturner From the original issue #623 regarding `delete`:

> _internally this should mark the field as deleted, but since we can't really re-use the rows allocated to that field (unless we completely cleared them), we'll just leave the data as is`

Fields are stored in separate views so we could just drop the entire view instead of simply marking the field as deleted. Do you see an issue with that approach?

travisturner:
@benbjohnson I wrote that before I realized we would store fields in separate views. I think your approach is a good one. @raskle implemented some `DeleteView` functionality in #785, so hopefully that can be used here.

benbjohnson:
@travisturner I added support for field deletion in 8bd1ab3.

PR #814, opened 2017-09-05
Row id zero bug

### LINKED ISSUE #809: input definition assumes 0 for non-existent fields, sets unexpected bits
### Expected behavior
If a field is not present in a record sent to the input definition, for example the "stargazer_id" field, no bit would be set in the corresponding frame (stargazers).

### Actual behavior
A value of 0 is assumed for the missing field (at least in the case of stargazer_id, and the bit is set in that row.

Follow the input definition docs.

Macbook Pro single node cluster.

linhvo:
## Overview

Fixes #809 
add a test and some tests refactor using driven table test

## Pull request checklist

## Code review checklist


input_definition.go
```diff
@@ -372,7 +371,7 @@ func HandleAction(a Action, value interface{}, colID uint64, timestamp int64) (*
 		}
 		bit.RowID = uint64(v)
 	case InputSetTimestamp:
```
raskle: I think we should add a comment about where the timestamp is handled earlier, and why we need to return nil.

raskle: how about:
//  InputSetTimestamp action is used in the InputJSONDataParser Handler to append a timestamp to all bits in the frame. There are no individual rowID's to set, and the action is a no-op at this step.

PR #820, opened 2017-09-07
Reuse container storage on UnmarshalBinary to improve memory utilization

### LINKED ISSUE #804: Memory thrashing issue importing tall skinny data
I've noticed a strange issue while importing a huge number of rows (~300M) with only a few columns. The import slows down significantly, and memory usage creeps up close to 100% - attempts to profile cpu or heap time out. After some time, memory usage drops to 50% or so, and the cycle repeats. Once the memory usage drops, it's possible to profile and heap usage is almost entirely in roaring/Bitmap.UnmarshalBinary.

This is on an AWS instance with 60GB of RAM - the import completes on a larger instance without much issue.

The size of the pilosa data directory while this is happening is about 4.4GB - the finished size is 6.7GB.

I would not expect memory usage to be so much larger than the size of the dataset, and to fluctuate so wildly. Will add more detail to this issue as it comes.

tgruben:
## Overview

Fixes #804

## Pull request checklist

## Code review checklist

### DIFF
--- a/fragment.go
+++ b/fragment.go
@@ -312,7 +312,8 @@ func (f *Fragment) close() error {
 
 func (f *Fragment) closeStorage() error {
 	// Clear the storage bitmap so it doesn't access the closed mmap.
-	f.storage = roaring.NewBitmap()
+
+	//f.storage = roaring.NewBitmap()
 
 	// Unmap the file.
 	if f.storageData != nil {
--- a/fragment.go
+++ b/fragment.go
@@ -187,8 +187,9 @@ func (f *Fragment) Open() error {
 // openStorage opens the storage bitmap.
 func (f *Fragment) openStorage() error {
 	// Create a roaring bitmap to serve as storage for the slice.
-	f.storage = roaring.NewBitmap()
-
+	if f.storage == nil {
+		f.storage = roaring.NewBitmap()
+	}
 	// Open the data file to be mmap'd and used as an ops log.
 	file, err := os.OpenFile(f.path, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0666)
 	if err != nil {
--- a/roaring/roaring.go
+++ b/roaring/roaring.go
@@ -622,6 +622,10 @@ func (b *Bitmap) UnmarshalBinary(data []byte) error {
 		b.keys = make([]uint64, 0, keyN)
 		b.containers = make([]*container, 0, keyN)
 	} else if int(keyN) < len(b.keys) { //shrink
+		// nil out to allow to be GCed
+		for i := range b.containers[keyN:] {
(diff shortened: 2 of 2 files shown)


roaring/roaring.go
```diff
@@ -617,18 +617,34 @@ func (b *Bitmap) UnmarshalBinary(data []byte) error {
...
+				n:              int(binary.LittleEndian.Uint16(buf[10:12])) + 1,
+				mapped:         true,
+			})
+		} else {
+			b.keys[i] = binary.LittleEndian.Uint64(buf[0:8])
+			c := b.containers[i]
+			c.container_type = byte(binary.LittleEndian.Uint16(buf[8:10]))
+			c.n = int(binary.LittleEndian.Uint16(buf[10:12])) + 1
+			c.mapped = true
+
```
jaffee: I think in the next section where the data actually gets attached, we need to nil out the container types that aren't being used - i.e. if it is an array type, we should do `c.runs = nil, c.bitmap = nil`

fragment.go
```diff
@@ -312,7 +313,8 @@ func (f *Fragment) close() error {
 
 func (f *Fragment) closeStorage() error {
 	// Clear the storage bitmap so it doesn't access the closed mmap.
-	f.storage = roaring.NewBitmap()
+
+	//f.storage = roaring.NewBitmap()
```
jaffee: are we concerned about anything accessing the closed mmap now that we aren't doing this? Should be nil out all the container storage here to prevent this?

tgruben: nil - ing out the container storage would make it available for GC. i could add an extra pointer to the fragment to hold the reference during the invalid period that occurs after close and before open, but all that would accomplish is awareness of the unexpected act. In the current release, we are throwing away those actions that occur during that period.  I'm not even sure that can happen.

roaring/roaring.go
```diff
@@ -617,18 +617,34 @@ func (b *Bitmap) UnmarshalBinary(data []byte) error {
...
-	b.keys = make([]uint64, keyN)
-	b.containers = make([]*container, keyN)
+
+	if len(b.keys) == 0 {
+		b.keys = make([]uint64, 0, keyN)
+		b.containers = make([]*container, 0, keyN)
+	} else if int(keyN) < len(b.keys) { //shrink
+		b.keys = b.keys[:keyN]
+		b.containers = b.containers[:keyN]
+	}
```
benbjohnson: The containers in `b.containers[keyN:]` should be `nil`'d out before shrinking. Otherwise they won't get GC'd.

jaffee:
@benbjohnson thanks for the review - can you approve if todd's changes look ok?

PR #823, opened 2017-09-11
Fixes Sum call response; resolves #821, resolves #822

### LINKED ISSUE #821: Replace Sum and Average calls with SumCount call
### Description

When the response is JSON, `Average` call returns a single integer even though the result may be a float. This is problematic, since the result is not correct, and also the user has no control over the conversion of the float to integer.

Meanwhile, in the protobuf response there are two fields: `Sum` and `Count` instead of the single integer. `Sum` and `Average` returns the same result!

In order to resolve these problems, we can have a `SumReduce` call which returns `Sum` and `Count` fields. That would be easier to implement than the other option: returning floats for `Average` and returning floats (or integers) from the `Sum` call.

**EDIT**
The protobuf result has the `SumCount` field which has a `Count` field of type `int64`. The result also has the `N` field which is of type `uint64`. That is confusing. We can remove the `SumCount` field from the result, and add a `Sum` field which is `int64`. The `N` field would be used in place of `Count`.

**EDIT 2**
See: https://github.com/pilosa/pilosa/issues/822 for a related issue.

**EDIT 3**
Renamed `SumCount` call to `SumReduce`.

### LINKED ISSUE #822: Protobuf response for Average and Sum calls is broken
### Expected behavior

The protobuf results for `Average` and `Sum` queries should have their `SumCount` field set accordingly.

### Actual behavior

`SumCount` field for protobuf results for `Average` and `Sum` queries is `nil`.

This happens because `handler.encodeQueryResponse` function receives `int64` results for `Average` and `Sum` instead of `SumCount`. So the `case SumCount:` statement around line 1514 doesn't get run which results in `pb.Results[i].SumCount == nil`.

yuce:
## Overview

Fixed Sum call to return SumCount; removed Average call.

Fixes #821, #822

## Pull request checklist

## Code review checklist

### DIFF
--- a/executor_test.go
+++ b/executor_test.go
@@ -645,7 +645,7 @@ func TestExecutor_Execute_SumReduce(t *testing.T) {
 	}
 
 	t.Run("NoFilter", func(t *testing.T) {
-		if result, err := e.Execute(context.Background(), "i", test.MustParse(`SumReduce(frame=f, field=foo)`), nil, nil); err != nil {
+		if result, err := e.Execute(context.Background(), "i", test.MustParse(`Sum(frame=f, field=foo)`), nil, nil); err != nil {
 			t.Fatal(err)
 		} else if !reflect.DeepEqual(result[0], pilosa.SumCount{Sum: 200, Count: 5}) {
 			t.Fatalf("unexpected result: %s", spew.Sdump(result))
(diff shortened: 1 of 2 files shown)


yuce:
This should be ready to review. Although since Sum call can cause an overflow in the `SumCount.Sum` field, we should remove it or increase the size of `SumCount.Sum` to 128 bits (protobuf doesn't support it, so we would have to store it as 2 int64s)

executor_test.go
```diff
@@ -600,8 +600,8 @@ func TestExecutor_Execute_TopN_Attr_Src(t *testing.T) {
 	}
 }
 
-// Ensure a Sum() query can be executed.
-func TestExecutor_Execute_Sum(t *testing.T) {
+// Ensure a SumReduce() query can be executed.
```
travisturner: Should we change this test name and comment back to `Sum` instead of `SumReduce`?

PR #842, opened 2017-09-25
Roaring optimizations

travisturner:
## Overview
Various fixes that were introduced during the SSB benchmarking project.

### DIFF
--- a/roaring/roaring.go
+++ b/roaring/roaring.go
@@ -2531,15 +2531,46 @@ func differenceRunArray(a, b *container) *container {
 
 // differenceRunBitmap computes the difference of an run from a bitmap.
 func differenceRunBitmap(a, b *container) *container {
-	if a.n == 0 || b.n == 0 {
-		return a.clone()
-	}
 	// If a is full, difference is the flip of b.
 	if a.runs[0].start == 0 && a.runs[0].last == 65535 {
 		return b.flipBitmap()
 	}
-	itr := newBufBitmapIterator(newBitmapIterator(b.bitmap))
-	return differenceRunIterator(a, itr)
+	output := &container{container_type: ContainerRun}
+	output.n = a.n
+	for j := 0; j < len(a.runs); j++ {
+		run := a.runs[j]
+		for bit := a.runs[j].start; bit <= a.runs[j].last; bit++ {
+			if b.bitmapContains(bit) {
+				output.n--
+				if run.start == bit {
+					run.start++
+				} else if bit == run.last {
+					run.last--
+				} else {
+					run.last = bit - 1
+					if run.last >= run.start {
+						output.runs = append(output.runs, run)
+					}
+					run.start = bit + 1
+					run.last = a.runs[j].last
+				}
+				if run.start > run.last {
+					break
+				}
+			}
+		}
+		if run.start <= run.last {
+			output.runs = append(output.runs, run)
+
+		}
+	}
+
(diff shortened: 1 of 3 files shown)


roaring/roaring.go
```diff
@@ -2658,83 +2646,46 @@ func differenceRunRun(a, b *container) *container {
...
+	output := &container{bitmap: make([]uint64, bitmapN), container_type: ContainerBitmap}
 
-		}
+	for i := range a.bitmap {
+		v := a.bitmap[i] & (^b.bitmap[i])
+		output.bitmap[i] = v
+		output.n += int(popcount(v))
+
+	}
+	if output.n < ArrayMaxSize {
```
jaffee: Optimize?

roaring/roaring.go
```diff
@@ -2658,83 +2646,46 @@ func differenceRunRun(a, b *container) *container {
...
-		}
-		if v0 < v1 {
-			output.add(v0)
-			v0, eof0 = itr0.next()
-		} else if v0 > v1 {
-			v1, eof1 = itr1.next()
-		} else {
-			v0, eof0 = itr0.next()
-			v1, eof1 = itr1.next()
+	output := &container{bitmap: make([]uint64, bitmapN), container_type: ContainerBitmap}
```
jaffee: not related to this PR, but we super duper need to rename `container_type`

roaring/roaring_internal_test.go
```diff
@@ -669,24 +668,30 @@ func TestUnionMixed(t *testing.T) {
 	if !reflect.DeepEqual(res.runs, []interval16{{start: 5, last: 10}}) {
 		t.Fatalf("test #3 expected %v, but got %v", []interval16{{start: 5, last: 10}}, res.runs)
 	}
-	c.bitmap = []uint64{0x3}
+
```
jaffee: this test should probably be refactored to be table-driven

roaring/roaring_internal_test.go
```diff
@@ -1542,8 +1547,8 @@ func TestDifferenceBitmapRun(t *testing.T) {
 }
 
 func TestDifferenceBitmapArray(t *testing.T) {
-	a := &container{bitmap: make([]uint64, bitmapN)}
-	b := &container{}
+	a := &container{bitmap: make([]uint64, bitmapN), container_type: ContainerBitmap}
```
jaffee: This test could use more cases - also the bitmap should use a helper function to construct a full bitmap, and we should make sure we're testing the edges of the container. The helper functions bitmapEvens and bitmapOdds are good for this.

roaring/roaring_internal_test.go
```diff
@@ -2410,3 +2416,83 @@ func TestSearc64(t *testing.T) {
...
+			array:  []uint16{0, 1, 63, 120, 543, 639, 12000, 65534, 65535},
+			bitmap: bitmapEvens(),
+			exp:    []uint16{0, 120, 12000, 65534},
+		},
+	}
+
+	for i, test := range tests {
+		a.array = test.array
+		a.container_type = ContainerArray
+		for i, bmval := range test.bitmap {
```
jaffee: we should use helper functions to create full bitmaps and get rid of this gross for loop

roaring/roaring_internal_test.go
tgruben: there is also a MakeBitmap(..)  that will alloc full bitmap with passed in initialize values

PR #845, opened 2017-09-25
forgotten error check which fooled the compiler but not guru/rename

travisturner:
see https://github.com/golang/go/issues/8560

also gofmting happening

### DIFF
--- a/input_definition_test.go
+++ b/input_definition_test.go
@@ -203,14 +203,14 @@ func TestHandleAction(t *testing.T) {
 		name     string
 		value    interface{}
 		expected uint64
-		err 	 string
+		err      string
 	}{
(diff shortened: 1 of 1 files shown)


input_definition_test.go
```diff
@@ -226,6 +226,9 @@ func TestHandleAction(t *testing.T) {
 	value = true
 	action.ValueDestination = pilosa.InputSingleRowBool
 	b, err := pilosa.HandleAction(action, value, colID, timestamp)
+	if err != nil {
```
travisturner: @jaffee do we also need to handle the `err` on line 253?

jaffee: uh, yep, looks like it

travisturner: added it

PR #816, opened 2017-09-05
BSI docs

raskle:
## Overview
BSI range encoding docs.
This work is dependent on Range Encoding additions to the Getting Started project

Fixes #740 
Fixes #853 

## Pull request checklist

- [x] I have read the [contributing guide](https://github.com/pilosa/pilosa/blob/master/CONTRIBUTING.md).
- [x] I have agreed to the [Contributor License Agreement](https://cla-assistant.io/pilosa/pilosa).
- [x] I have updated the [documentation](https://github.com/pilosa/pilosa/tree/master/docs).
- [x] I have resolved any merge conflicts.
- [x] I have included tests that cover my changes.
- [x] All new and existing tests pass.

## Code review checklist
This is the checklist that the reviewer will follow while reviewing your pull request. You do not need to do anything with this checklist, but be aware of what the reviewer will be looking for.

- [ ] Ensure that any changes to external docs have been included in this pull request.
- [ ] If the changes require that minor/major versions need to be updated, tag the PR appropriately.
- [ ] Ensure the new code is [properly commented](https://github.com/golang/go/wiki/CodeReviewComments#doc-comments) and follows [Idiomatic Go](https://dmitri.shuralyov.com/idiomatic-go).
- [ ] Check that tests have been written and that they cover the new functionality.
- [ ] Run tests and ensure they pass.
- [ ] Build and run the code, performing any applicable integration testing.

PR #847, opened 2017-09-25
Implements BETWEEN for Range queries.

travisturner:
The PQL looks like:
```
Range(frame=f, field0 >< [200,610])
```

One thing I noticed while implementing this is that it doesn't seem
like `FieldRange()` is used in either `Frame` or `View`; the Executor
calls `Fragment.FieldRange()` directly. The problem with this is that
the offset logic is calculated in the Frame, but since the Executor
doesn't go through Frame, then the Executor also has to calculate
the offset before calling `Fragment.FieldRange`. We should unify this
logic somewhere. Note, this applies to both `FieldRange` and
`FieldRangeBetween`.

## Pull request checklist

## Code review checklist

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -731,21 +731,11 @@ func (e *Executor) executeFieldRangeSlice(ctx context.Context, index string, c *
 		field := f.Field(fieldName)
 		if field == nil {
 			return nil, ErrFieldNotFound
-		} else if predicates[1] < field.Min || predicates[0] > field.Max {
-			return NewBitmap(), nil
 		}
 
-		// Adjust predicates to range.
-		baseValueMin := uint64(0)
-		baseValueMax := uint64(0)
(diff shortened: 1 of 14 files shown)


alanbernstein:
The "><" syntax needs to be documented before release (#816).

yuce:
How about something like `Range(frame=f, 200 <= field0 <= 610)` (200 and 610 are included) or `Range(frame=f, 200 <= field0 < 610)` (200 is included but 610 is excluded) instead of `Range(frame=f, field0 >< [200,610])` ?

travisturner:
@yuce Yes, that would the be preferred format. We didn't implement it that way because we needed the functionality quickly for our benchmark work, and we didn't want to spend the time required to modify the parser to handle that type of format. But I agree it's probably the way we should do it.

PR #859, opened 2017-09-27
Refactor differencerunarray

travisturner:


tgruben:
What value is this adding? I see no performance advantage over arrayIterator (bitmap iterator is a different story)

jaffee:
the drive for this change was lots of heap allocation in the array iterator if I'm remembering correctly. I'm sure the code could be cleaner, I wrote it in the middle of the night last Thursday.

PR #867, opened 2017-10-04
TLS support

### LINKED ISSUE #864: TLS Support
# TLS Support

### Use URI struct instead of string addresses

Curently host addresses are kept as strings in the server code. This was fine when we had only the `http` scheme, but since we have `https` now and maybe more in the future, we should be able to keep the scheme alongside the host and port as well. In the Go client, we have the `URI` struct which keeps the scheme, host and port as separate fields and is also able to parse Pilosa URIs. This struct may be used in the server.

### New Configuration Items

There are a few additional configuration items necessary for TLS support.

#### server subcommand

* Path to the server certificate: `tls.certificate`.
* Path to the server certificate key: `tls.key`.
* Client certificate verification (default: disabled): `tls.verify-client`.
* Path to the client certificate CA which is required when using self signed certificates and client certificate verification is enabled: `tls.client-certificate`.

#### import and export subcommands

* Server certificate verification (default: enabled): `tls.verify`
* Path to the client certificate: `tls.client-certificate-path`.
* Path to the client certificate key: `tls.client-certificate-key-path`.

## Client Changes

Client libraries already use `URI` objects to keep addresses, so there are only minimal changes required. SSL support is highly programming language specific, so we don't aim for a uniform way of setting that. Instead, the client libraries should have a way for the user to specify SSL settings in `ClientOptions`. The following are some of the settings expected to configure SSL on the client side:

* Skip server certificate verification (default: false): `SkipVerify`.
* Path to CA certificate: `CACertificate`.
* Path to client certificate: `ClientCertificate`.
* Path to client certificate key: `ClientCertificateKey`.

yuce:
## Overview

Adds TLS support.

Resolves #864 

Here's how I ran a sample cluster and test it. `--tls.skip-verify` is only required if the certificate is self-signed:

1) Created a self-signed certificate: `openssl req -x509 -newkey rsa:4096 -keyout 01.pilosa.local.key -out 01.pilosa.local.crt -days 3650 -nodes -subj "/C=US/ST=Texas/L=Austin/O=Pilosa/OU=Org/CN=01.pilosa.local"`
2) `make pilosa`
3) Run the 1st server in the cluster as: `./pilosa server -b https://01.pilosa.local:10501 --tls.certificate 01.pilosa.local.crt --tls.key 01.pilosa.local.key -d pil-01  --cluster.hosts https://01.pilosa.local:10501,https://01.pilosa.local:10502 --gossip-seed 01.pilosa.local:15000 --gossip-port 15000 --tls.skip-verify`
4) Run the 2nd server in the cluster as: `./pilosa server -b https://01.pilosa.local:10502 --tls.certificate 01.pilosa.local.crt --tls.key 01.pilosa.local.key -d pil-02 --cluster.hosts https://01.pilosa.local:10501,https://01.pilosa.local:10502 --gossip-seed 01.pilosa.local:15000 --gossip-port 16000 --tls.skip-verify`
5) Create an index and frame `curl -k https://01.pilosa.loca:10501/index/i1 -d ''`, `curl -k https://01.pilosa.loca:10501/index/i1/frame/f1 -d ''` . `-k` switch is only required if the certificate is self-signed.
6) Run a SetBit query: `curl -k https://01.pilosa.local:10501/index/i1/query -d 'SetBit(frame="f1", rowID=1, columnID=1048577)'`
7) Run a Bitmap query: `curl -k https://01.pilosa.local:10502/index/i1/query -d 'Bitmap(frame="f1", rowID=1)'`
8) Run an export cmd: `./pilosa export --host https://01.pilosa.local:10501 -i i1 -f f1 --tls.skip-verify --tls.certificate 01.pilosa.local.crt --tls.key 01.pilosa.local.key`

## Pull request checklist

## Code review checklist

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -56,8 +56,8 @@ type Handler struct {
 	StatusHandler StatusHandler
 
 	// Local hostname & cluster configuration.
-	Host    *URI
-	Cluster *Cluster
+	URI           *URI
+	Cluster       *Cluster
 	ClientOptions *ClientOptions
 
 	Router *mux.Router
@@ -1167,8 +1167,8 @@ func (h *Handler) handlePostImport(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Validate that this handler owns the slice.
-	if !h.Cluster.OwnsFragment(h.Host.HostPort(), req.Index, req.Slice) {
-		mesg := fmt.Sprintf("host does not own slice %s-%s slice:%d", h.Host, req.Index, req.Slice)
+	if !h.Cluster.OwnsFragment(h.URI.HostPort(), req.Index, req.Slice) {
+		mesg := fmt.Sprintf("host does not own slice %s-%s slice:%d", h.URI, req.Index, req.Slice)
 		http.Error(w, mesg, http.StatusPreconditionFailed)
 		return
 	}
@@ -1237,8 +1237,8 @@ func (h *Handler) handlePostImportValue(w http.ResponseWriter, r *http.Request) 
 	}
 
 	// Validate that this handler owns the slice.
-	if !h.Cluster.OwnsFragment(h.Host.HostPort(), req.Index, req.Slice) {
-		mesg := fmt.Sprintf("host does not own slice %s-%s slice:%d", h.Host, req.Index, req.Slice)
+	if !h.Cluster.OwnsFragment(h.URI.HostPort(), req.Index, req.Slice) {
+		mesg := fmt.Sprintf("host does not own slice %s-%s slice:%d", h.URI, req.Index, req.Slice)
 		http.Error(w, mesg, http.StatusPreconditionFailed)
 		return
 	}
@@ -1303,8 +1303,8 @@ func (h *Handler) handleGetExportCSV(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Validate that this handler owns the slice.
-	if !h.Cluster.OwnsFragment(h.Host.HostPort(), index, slice) {
-		mesg := fmt.Sprintf("host does not own slice %s-%s slice:%d", h.Host, index, slice)
+	if !h.Cluster.OwnsFragment(h.URI.HostPort(), index, slice) {
+		mesg := fmt.Sprintf("host does not own slice %s-%s slice:%d", h.URI, index, slice)
 		http.Error(w, mesg, http.StatusPreconditionFailed)
 		return
 	}
@@ -1535,7 +1535,7 @@ func (h *Handler) handlePostFrameRestore(w http.ResponseWriter, r *http.Request)
 	// Loop over each slice and import it if this node owns it.
 	for slice := uint64(0); slice <= maxSlices[indexName]; slice++ {
 		// Ignore this slice if we don't own it.
-		if !h.Cluster.OwnsFragment(h.Host.HostPort(), indexName, slice) {
+		if !h.Cluster.OwnsFragment(h.URI.HostPort(), indexName, slice) {
 			continue
 		}
 
--- a/handler.go
+++ b/handler.go
@@ -56,7 +56,7 @@ type Handler struct {
 	StatusHandler StatusHandler
 
 	// Local hostname & cluster configuration.
-	Host    string
+	Host    *URI
 	Cluster *Cluster
 
 	Router *mux.Router
@@ -1166,7 +1166,7 @@ func (h *Handler) handlePostImport(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Validate that this handler owns the slice.
-	if !h.Cluster.OwnsFragment(h.Host, req.Index, req.Slice) {
+	if !h.Cluster.OwnsFragment(h.Host.HostPort(), req.Index, req.Slice) {
 		mesg := fmt.Sprintf("host does not own slice %s-%s slice:%d", h.Host, req.Index, req.Slice)
 		http.Error(w, mesg, http.StatusPreconditionFailed)
 		return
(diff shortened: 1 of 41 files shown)


yuce:
Here's how I ran a sample cluster. `--tls.skip-verify` is only required if the certificate is self-signed:
1) Created a self-signed certificate: `openssl req -x509 -newkey rsa:4096 -keyout 01.pilosa.local.key -out 01.pilosa.local.crt -days 3650 -nodes -subj "/C=US/ST=Texas/L=Austin/O=Pilosa/OU=Org/CN=01.pilosa.local"`
2) `make pilosa`
3) Run the 1st server in the cluster as: `./pilosa server -b https://01.pilosa.local:10501 --tls.certificate 01.pilosa.local.crt --tls.key 01.pilosa.local.key -d pil-01  --cluster.hosts https://01.pilosa.local:10501,https://01.pilosa.local:10502 --gossip-seed 01.pilosa.local:15000 --gossip-port 15000 --tls.skip-verify`
4) Run the 2nd server in the cluster as: `./pilosa server -b https://01.pilosa.local:10502 --tls.certificate 01.pilosa.local.crt --tls.key 01.pilosa.local.key -d pil-02 --cluster.hosts https://01.pilosa.local:10501,https://01.pilosa.local:10502 --gossip-seed 01.pilosa.local:15000 --gossip-port 16000 --tls.skip-verify`
5) Create an index and frame `curl -k https://01.pilosa.loca:10501/index/i1 -d ''`, `curl -k https://01.pilosa.loca:10501/index/i1/frame/f1 -d ''` . `-k` switch is only required if the certificate is self-signed.
6) Run a SetBit query: `curl -k https://01.pilosa.local:10501/index/i1/query -d 'SetBit(frame="f1", rowID=1, columnID=1048577)'`
7) Run a Bitmap query: `curl -k https://01.pilosa.local:10502/index/i1/query -d 'Bitmap(frame="f1", rowID=1)'`
8) Run an export cmd: `./pilosa export --host https://01.pilosa.local:10501 -i i1 -f f1 --tls.skip-verify --tls.certificate 01.pilosa.local.crt --tls.key 01.pilosa.local.key`

handler.go
```diff
@@ -56,8 +56,9 @@ type Handler struct {
 	StatusHandler StatusHandler
 
 	// Local hostname & cluster configuration.
-	Host    string
+	Host    *URI
```
codysoyland: Can we rename the Host variable to `URI`?

holder.go
```diff
@@ -417,8 +417,9 @@ func (h *Holder) logger() *log.Logger { return log.New(h.LogOutput, "", log.Lstd
 type HolderSyncer struct {
 	Holder *Holder
 
-	Host    string
-	Cluster *Cluster
+	Host          *URI
```
codysoyland: Also here please change Host to `URI`

server.go
```diff
@@ -143,10 +160,14 @@ func (s *Server) Open() error {
 		return fmt.Errorf("opening NodeSet: %v", err)
 	}
 
+	// Create default HTTP client
+	s.createDefaultClient()
+
 	// Create executor for executing queries.
-	e := NewExecutor()
+	e := NewExecutor(&ClientOptions{TLS: s.TLS})
```
codysoyland: I think that ClientOptions shouldn't be an argument to NewExecutor. The pattern we have been following is to call NewExecutor and set options on the resulting object (below). Perhaps a method `Executor.SetClientOptions(*ClientOptions)` would set everything up (including Scheme and Host)?

server.go
```diff
@@ -59,18 +60,23 @@ type Server struct {
 	// Cluster configuration.
 	// Host is replaced with actual host after opening if port is ":0".
 	Network string
-	Host    string
+	Host    *URI
```
codysoyland: Also change Host to `URI` here please.

uri.go
```diff
@@ -0,0 +1,207 @@
...
+// 	http://localhost
+// 	http://:10101
+// 	localhost:10101
+// 	localhost
+// 	:10101
+type URI struct {
+	scheme string
+	host   string
+	port   uint16
+	error  error
```
codysoyland: What is the purpose of `error` being a member of the URI struct? It and the `Error()` and `Valid()` methods seem to only be used by tests. Seems to add code with no benefit.

uri.go
```diff
@@ -0,0 +1,207 @@
...
+func NewURIFromAddress(address string) (*URI, error) {
+	uri, err := parseAddress(address)
+	if err != nil {
+		return &URI{error: err}, err
+	}
+	return uri, err
+}
+
+// URIFromAddress creates a URI from the given address.
+func URIFromAddress(host string) *URI {
```
codysoyland: Why does this function exist?

yuce: It is for optionally ignoring errors while creating URIs to make the code a bit less verbose when the address is most probably valid. It sets the `error` field instead of returning it, so it can be used in function calls. It's not being used, so removed that, `error` and related stuff

server.go
yuce: `ClientOptions` contains configuration for only the client which is used to access other nodes (like syncing blocks, etc.). `Host` and `Scheme` belong to the node which the executor is running, so I don't think it makes sense to put them in client options. Maybe you meant to have something like `ExecutorOptions` which would include `ClientOptions` ?

`NewExecutor` sets up the HTTP client so IMO it's better to pass client options (or executor options?) to it. Is there an advantage of passing nothing to `NewExecutor` and having separate `Set` functions ? If the user wants to initialize an executor they can always use `Executor{}`.

handler.go
yuce: ok

PR #874, opened 2017-10-05
FieldNotNull for more efficient BETWEEN queries

### LINKED ISSUE #869: Not Null query
### Description
We need a query that can return the Not Null Bitmap for a BSI `Field`.  This query will have a `Frame` and `Field` argument, and return a bitmap.
The BSI field already uses the high bit to store not null so this can be used.

Can run the is Not Null query on an 'Field' that returns a bitmap of all columns with a BSI value set.

travisturner:
## Overview

In the example use-case where a `field` represents a calendar day, and the field's value range is `[0-1439]` such that a value represents a particular minute of the day, then if we want to support a range query that spans days, we can make the query more efficient by returning the `not-null` rows for any full field (day) in the middle of the range.

For example, if we want to query a 48-hour range like `2017-10-03T10:28 to 2017-10-05T10:27`, then we effectively want this:
```
Union(
  Range(frame=f, d20171003 >= 628),
  Range(frame=f, d20171004 >< [0, 1439]),
  Range(frame=f, d20171005 < 628)
)
```
The first pass at this PR doesn't change the query language. The second Range() query in the example above will return the `not-null` row for that field. This applies to any case where the query range is either equal to or outside of the field's min/max values.

Fixes #869

## Pull request checklist

## Code review checklist

### DIFF
--- a/fragment.go
+++ b/fragment.go
@@ -730,6 +730,11 @@ func (f *Fragment) fieldRangeGT(bitDepth uint, predicate uint64, allowEquality b
 	return b, nil
 }
 
+// FieldNotNull returns the not-null row (stored at bitDepth).
+func (f *Fragment) FieldNotNull(bitDepth uint) (*Bitmap, error) {
+	return f.Row(uint64(bitDepth)), nil
+}
+
 func (f *Fragment) FieldRangeBetween(bitDepth uint, predicateMin, predicateMax uint64) (*Bitmap, error) {
 	b := f.Row(uint64(bitDepth))
 	keep1 := NewBitmap() // GTE
--- a/fragment.go
+++ b/fragment.go
@@ -619,6 +619,8 @@ func (f *Fragment) FieldRange(op pql.Token, bitDepth uint, predicate uint64) (*B
 	switch op {
 	case pql.EQ:
 		return f.fieldRangeEQ(bitDepth, predicate)
+	case pql.NEQ:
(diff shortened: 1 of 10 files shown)


travisturner:
The latest commit adds support for `field != null` in the `Range()` query.

raskle:
This looks good so far, do you want to handle != for integers in this PR?

travisturner:
@raskle If it's not too much work I'll add a commit for `!= <int>` to this PR. If it becomes more complicated, I'll open a new PR instead.

travisturner:
@raskle I added support for `field != <int>` to this PR.
The only operation left to be implemented is `field == null`, but that one is dependent upon knowing `maxColumnID`, so we'll leave it for a future task.

fragment.go
```diff
@@ -647,6 +649,22 @@ func (f *Fragment) fieldRangeEQ(bitDepth uint, predicate uint64) (*Bitmap, error
 	return b, nil
 }
 
+func (f *Fragment) fieldRangeNEQ(bitDepth uint, predicate uint64) (*Bitmap, error) {
```
raskle: The != operation does not work for negative integers.

travisturner:
@raskle I don't think it was specifically negative values, but rather values that fell outside of the valid range for the field. For those cases, NEQ should return the not-null bitmap. I pushed a commit for that.

PR #908, opened 2017-10-27
Update status schema

travisturner:
Add Schema to proto.
Update MaxSlices in proto to include both standard and inverse.
Update LocalStatus (shared in gossip) to include MaxSlices and Schema.
Encode InputDefinitions with Index.

TODO:
- make sure InputDefinitions are considered in LocalStatus merge.
- decide what to do about `/slices/max` endpoint. client.backup is using
  it.

@raskle This compiles but the tests are failing. I'm just putting this here as a WIP so you can see what direction I'm thinking with the status/schema stuff.


travisturner:
@raskle i fixed the tests.

PR #923, opened 2017-11-01
add test helper for starting a new pilosa instance

jaffee:
Pilosa runs on ephemeral ports with temporary storage.

### DIFF
--- a/test/pilosa.go
+++ b/test/pilosa.go
@@ -1,0 +1,48 @@
+package test
+
+import (
+	"bytes"
+	"io/ioutil"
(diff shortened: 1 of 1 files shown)


jaffee:
@travisturner I think you mentioned that you were working on some test helping code. I wrote this a little while ago for some integration tests in another repo, but I also need it in the PDK, so I think it makes sense to expose this functionality in the Pilosa test package.

In any case, if you've already got something like this (or something more robust) feel free to close, and/or just steal any code you need.

PR #929, opened 2017-11-03
Add Cluster resize tests.

travisturner:
Consolidate schema creation into Holder.ApplySchema.
Add view names to proto schema.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -901,9 +901,7 @@ func (j *ResizeJob) setState(state string) {
 // Run distributes ResizeInstructions.
 func (j *ResizeJob) Run() error {
 	// Set job state to RUNNING.
-	j.mu.RLock()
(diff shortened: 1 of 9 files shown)


cluster.go
```diff
@@ -921,11 +900,10 @@ func (j *ResizeJob) setState(state string) {
 
 // Run distributes ResizeInstructions.
 func (j *ResizeJob) Run() error {
-	j.mu.RLock()
-	defer j.mu.RUnlock()
-
 	// Set job state to RUNNING.
+	j.mu.RLock()
 	j.setState(ResizeJobStateRunning)
+	j.mu.RUnlock()
```
tgruben: looks like that would need to be a write lock

travisturner: good catch. i pushed a fix.

PR #945, opened 2017-11-10
Removes column/row labels for input definition. Resolves #810

### LINKED ISSUE #810: Remove input definition dependence on row/col labels
### Description

Currently the input definition feature uses row and column labels as a normal part of it's operation. Since we want to remove this feature (#752) , we need to refactor input definition to not be dependent on it

Use of input definition no longer requires row/column labels, and all mention of row/col labels is removed from input definition docs and examples.

yuce:
## Overview

The primary key name defaults to `DefaultColumnLabel`. Primary key field is optional. If not given and there are other fields, a new field with `Name=DefaultColumnLabel, PrimaryKey=true, Actions=[]` is added to the input def.

Fixes #810 

## Pull request checklist

## Code review checklist

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -1760,8 +1760,7 @@ func (h *Handler) handlePostInputDefinition(w http.ResponseWriter, r *http.Reque
 		return
 	}
 
-	// Validation the input definition with the curent index's ColumnLabel.
-	if err := req.Validate(index.ColumnLabel()); err != nil {
+	if err := req.Validate(); err != nil {
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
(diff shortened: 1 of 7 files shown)


input_definition.go
```diff
@@ -111,14 +115,27 @@ func (i *InputDefinition) LoadDefinition(pb *internal.InputDefinition) error {
...
 			PrimaryKey: field.PrimaryKey,
 			Actions:    actions,
 		}
 		i.fields = append(i.fields, inputField)
 	}
 
+	if len(pb.Fields) > 0 && !primaryKeyGiven {
+		// primary field is required if there are other fields.
+		// add it if it doesn't exist.
+		i.fields = append(i.fields, InputDefDefaultPrimaryKeyField())
```
travisturner: It seems confusing to add a field to the input definition. Why not just return an error if a primary key is not set?

yuce: It's convenient to leave out the primary key in the definition, since its name would always be the default column label (currently `columnID`). If we don't add the primary key field auto, the user has to add this to the following to `fields` for any input def:
```json
{
  "name": "columnID",
  "primaryKey": true
}
```

handler.go
```diff
@@ -1908,10 +1907,13 @@ func (h *Handler) InputJSONDataParser(req map[string]interface{}, index *Index,
 	for _, field := range inputDef.Fields() {
 		validFields[field.Name] = true
 		if field.PrimaryKey {
-			columnLabel := field.Name
-			value, ok := req[columnLabel]
+			primaryKey := field.Name
```
travisturner: This logic is not correct. We don't want to require the user to provide a `columnID` field in the input definition. The input definition can have any field marked as its `primaryKey`. No matter what that field is called, its value gets used as the column value.

input_definition.go
```diff
@@ -111,14 +115,25 @@ func (i *InputDefinition) LoadDefinition(pb *internal.InputDefinition) error {
 			})
 		}
 
+		if field.PrimaryKey {
```
travisturner: Similar to my comment in `handler.go`, this renaming of the field is not correct.

PR #985, opened 2017-11-29
Add HolderCleaner and view.DeleteFragment

travisturner:
Add HolderCleaner and view.DeleteFragment to support post-resize cleanups
Add tests for view.DeleteFragment and HolderCleaner

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -304,6 +304,8 @@ func (c *Cluster) setState(state string) {
 
 	c.State = state
 
+	// TODO: consider NOT running cleanup on an active node that has
+	// been removed.
 	// It's safe to do a cleanup after state changes back to normal.
 	if doCleanup {
 		var cleaner HolderCleaner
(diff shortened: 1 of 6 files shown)


cluster.go
```diff
@@ -284,17 +284,41 @@ func (c *Cluster) setState(state string) {
...
+	// It's safe to do a cleanup after state changes back to normal.
+	if doCleanup {
+		var cleaner HolderCleaner
+		cleaner.URI = c.URI
+		cleaner.Holder = c.Holder
+		cleaner.Cluster = c
+		cleaner.Closing = c.closing
+
+		// Clean holder.
+		if err := cleaner.CleanHolder(); err != nil {
```
tgruben: Im not sure of the correct answer here, but i'm not sure if we want to leave it in state NORMAL if the cleanup fails, though I'm not clear on what else it could be

travisturner: This is a valid concern.

I think the biggest problem would be if you resized a cluster, the cleanup failed, and then later you did another resize and those lingering fragments were  conflicting with the new topology. We need to test that scenario and see how a resize over orphaned fragments behaves.

Another option is to hold the cluster in an inactive state until manual intervention.

view.go
```diff
@@ -270,6 +274,36 @@ func (v *View) newFragment(path string, slice uint64) *Fragment {
...
 
+// DeleteFragment removes the fragment from the view.
+func (v *View) DeleteFragment(slice uint64) error {
+
+	fragment := v.fragments[slice]
+	if fragment == nil {
+		return ErrFragmentNotFound
+	}
+
+	v.logger().Printf("delete fragment: %d", slice)
```
tgruben: If you have access to the frame and view labels that would be informative for the log

PR #1077, opened 2018-02-01
Use NodeID instead of URI for node identification

travisturner:
## Overview

This PR changes the way nodes are identified in the cluster from `Cluster.URI` to `Cluster.Node.ID`. Note that `Node` is now defined as:
```
// Node represents a node in the cluster.
type Node struct {
    ID  string `json:"id"`
    URI URI    `json:"uri"`
}
```
The `URI` is maintained as part of the `Node` for use in query distribution and things of that nature, but `URI` is no longer used to determine partition location or anything like that. The node `ID` is stored in the data directory in a file called `ID`. If that file does not exist, then one is created. The means that a user can bring up the same node (i.e. data directory) with a different bind address (`URI`) and everything will still work as expected.

This changes some of the APIs. For example, removing a node now requires that you provide the node `ID`:
```
curl -X POST "http://localhost:10101/cluster/resize/remove-node" -d '{"id": "32b20e74-5be9-4476-be69-4717c6247cd3"}'
```

I realize this is a pretty extensive change, but even if you can't take in the entire PR, I would appreciate feedback on how it's implemented at a high level.

## Pull request checklist

## Code review checklist


cluster.go
```diff
@@ -227,28 +287,45 @@ func (c *Cluster) logger() *log.Logger {
...
-	return c.Static || c.Coordinator == c.URI
+	return c.Static || c.Coordinator == c.Node.URI
 }
 
 // SetCoordinator updates the Coordinator to new if it is
 // currently old. Returns true if the Coordinator changed.
-func (c *Cluster) SetCoordinator(oldURI, newURI URI) bool {
-	if c.Coordinator == oldURI && oldURI != newURI {
-		c.Coordinator = newURI
+func (c *Cluster) SetCoordinator(o, n *Node) bool {
```
jaffee: why does this need the old and new node?

cluster.go
```diff
@@ -550,57 +636,57 @@ func (c *Cluster) fragCombos(idx string, maxSlice uint64, frameViews viewsByFram
...
 // added or removed. An error is returned for any case other than where
 // exactly one node is added or removed.
-func (c *Cluster) diff(other *Cluster) (action string, uri URI, err error) {
+func (c *Cluster) diff(other *Cluster) (action string, nodeID string, err error) {
 	lenFrom := len(c.Nodes)
 	lenTo := len(other.Nodes)
 	// Determine if a node is being added or removed.
 	if lenFrom == lenTo {
-		return action, uri, errors.New("clusters are the same size")
+		return action, nodeID, errors.New("clusters are the same size")
```
jaffee: this applies elsewhere as well, but I think in general it's best to explicitly return 0 values for everything when you're returning a non-nil error. i.e. `return "", "", errors.New(...)`

gossip/gossip.go
```diff
@@ -387,14 +381,19 @@ func (g *GossipEventReceiver) listen() {
 			continue
 		}
 
-		uri, _ := pilosa.NewURIFromAddress(e.Node.Name)
-		// TODO: don't swallow this error
+		// Get the node from the event.Node meta data.
+		var n internal.Node
+		if err := proto.Unmarshal(e.Node.Meta, &n); err != nil {
+			// TODO: consider logging error
```
jaffee: is this error expected to happen ever? If not, let's panic

cluster.go
travisturner: I don't think it does. In fact, I think it's probably better not to confirm the current coordinator on remote nodes in the case where a broadcast of `SetCoordinator` is only successful on a subset of the nodes in a cluster. In that case, an error can be re-tried without leaving the cluster in a weird state.
I'll change it to require only new node.

PR #1121, opened 2018-02-19
Validate /fragment/nodes arguments. Fixes #652

### LINKED ISSUE #652: /fragment/nodes should validate query arguments
### Expected behavior

I would expect `/fragment/nodes?slice=0&db=somedb` to return an error, like: `db is not a valid query argument`

### Actual behavior

`[{"host":":10101","internalHost":""}]` was returned, instead of an error.

`curl localhost:10101/fragment/nodes?slice=0&db=myindex`

yuce:
## Overview

This PR adds query argument validation to all relevant handlers (taking query arguments).

Fixes #652 

## Pull request checklist

## Code review checklist

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -119,9 +119,7 @@ func (h *Handler) queryArgValidator(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		key := mux.CurrentRoute(r).GetName()
 		if validator, ok := h.validators[key]; ok {
-			q := r.URL.Query()
-			err := validator.validate(q)
-			if err != nil {
+			if err := validator.validate(r.URL.Query()); err != nil {
 				// TODO: Return the response depending on the Accept header
 				response := errorResponse{Error: err.Error()}
 				body, err := json.Marshal(response)
@@ -2099,13 +2097,13 @@ type defaultClusterMessageResponse struct{}
 
 type queryValidationSpec struct {
 	required []string
-	args     map[string]bool
+	args     map[string]struct{}
 }
 
 func QueryValidationSpecRequired(requiredArgs ...string) *queryValidationSpec {
-	args := map[string]bool{}
+	args := map[string]struct{}{}
 	for _, arg := range requiredArgs {
-		args[arg] = true
+		args[arg] = struct{}{}
 	}
 
 	return &queryValidationSpec{
@@ -2116,7 +2114,7 @@ func QueryValidationSpecRequired(requiredArgs ...string) *queryValidationSpec {
 
 func (s *queryValidationSpec) Optional(args ...string) *queryValidationSpec {
 	for _, arg := range args {
-		s.args[arg] = true
+		s.args[arg] = struct{}{}
 	}
 	return s
 }
--- a/handler.go
+++ b/handler.go
@@ -27,6 +27,7 @@ import (
 	"io/ioutil"
 	"log"
 	"net/http"
+	"net/url"
 	// Imported for its side-effect of registering pprof endpoints with the server.
 	_ "net/http/pprof"
 	"os"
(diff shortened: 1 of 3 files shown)


travisturner:
I like the direction of this, but I wonder if there's a more succinct way to define validators and prevent excessive code? For example, it seems like the validation code added to `handleGetFragmentNodes()` could be avoided. Perhaps creating/registering a [gorilla/mux middleware](https://github.com/gorilla/mux#middleware) validator would be a cleaner approach?

yuce:
I think that's a great idea, I'll see how I can use the gorilla middleware.

yuce:
@travisturner I've added a gorilla/mux middleware for validation (requires the latest gorilla/mux). Changed how a validator is created and attached to validators, so validators can be defined in a single line:
```go
h.validators["GET /fragment/nodes"] = QueryValidationSpecRequired("slice").Optional("index")
```

`QueryValidationSpecRequired` is still a long function name, I can rename to be shorter.

yuce:
I've changed how the key for a validator is found (otherwise it isn't possible to match handlers with dynamic paths). I've applied validation to all handlers taking query arguments.

handler.go
```diff
@@ -2057,3 +2084,41 @@ func (h *Handler) handleGetID(w http.ResponseWriter, r *http.Request) {
 }
 
 type defaultClusterMessageResponse struct{}
+
+type queryValidationSpec struct {
+	required []string
+	args     map[string]bool
```
travisturner: If you aren't going to use the `bool` value for anything, this could be `map[string]struct{}`.

handler.go
```diff
@@ -2057,3 +2084,41 @@ func (h *Handler) handleGetID(w http.ResponseWriter, r *http.Request) {
...
+}
+
+func QueryValidationSpecRequired(requiredArgs ...string) *queryValidationSpec {
```
travisturner: What do you think about splitting this into `NewQueryValidationSpec()` and `Required()`?
Then we could define something like:
```
h.validators["GET /some/path"] = NewQueryValidationSpec().Optional("index")
h.validators["GET /other/path"] = NewQueryValidationSpec().Required("frame")
```

handler.go
```diff
@@ -91,9 +95,30 @@ func NewHandler() *Handler {
...
+	h.validators["GET /fragment/nodes"] = QueryValidationSpecRequired("slice").Optional("index")
+}
+
+func (h *Handler) queryArgValidator(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		key := fmt.Sprintf("%s %s", r.Method, r.URL.Path)
+		if validator, ok := h.validators[key]; ok {
+			q := r.URL.Query()
+			err := validator.validate(q)
+			if err != nil {
```
travisturner: how about: `if err := validator.validate(r.URL.Query()); err != nil {`

handler.go
```diff
@@ -85,28 +89,67 @@ var externalPrefixFlag = map[string]bool{
...
+	h.validators["PostFrameRestore"] = QueryValidationSpecRequired("host")
+}
+
+func (h *Handler) queryArgValidator(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		key := mux.CurrentRoute(r).GetName()
+		if validator, ok := h.validators[key]; ok {
+			q := r.URL.Query()
+			err := validator.validate(q)
+			if err != nil {
```
travisturner: how about: `if err := validator.validate(r.URL.Query()); err != nil {`

handler.go
yuce: ok

handler.go
yuce: Actually it was that way before, but I've changed it to make it look a bit better:
```go
QueryValidationSpecRequired("index").Optional("frame")
```
instead of
```go
QueryValidationSpec().Required("index").Optional("frame")
```
I like the current way better, but I can change it to another style.

handler.go
yuce: ok

handler.go
yuce: ok

handler.go
travisturner: No need to change it. This was just a suggestion.

yuce:
Updated

PR #1122, opened 2018-02-19
WIP: Implements Bitmaps call; fixes #1029

### LINKED ISSUE #1029: proposal :Range query in row ids
### Description

I create a frame with unique numbers  as  column ids and some  type as row ids, it works well but sometimes I need query like  Count(Bitmap(frame="datas",  rowID >< [200,1234])) .
an workaround is use Union   : Count(Bitmap(frame="datas",rowID=200),Bitmap(frame="datas",rowID=201)... Bitmap(frame="datas",rowID=1234)) . 
Not simple at all.

support  > , < ,   >= ,  <= ,   >< operator  on rowID for range query .

yuce:
## Overview

Adds `BitmapRange` call and `~` operator.
Called like: `Union(BitmapRange(frame="data", rowID=[29~32]))`

Fixes #1029 

Currently works only together with `Union` calls, but it's trivial to add it to other bitmap calls. I'll do that once this approach is confirmed.

TODO: Misses tests.

## Pull request checklist

## Code review checklist

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -821,8 +821,8 @@ func (e *Executor) executeFieldRangeSlice(ctx context.Context, index string, c *
 func (e *Executor) executeUnionSlice(ctx context.Context, index string, c *pql.Call, slice uint64) (*Bitmap, error) {
 	other := NewBitmap()
 	for i, input := range c.Children {
-		if input.Name == "BitmapRange" {
-			err := e.executeBitmapRange(ctx, index, input, slice, func(bm *Bitmap, first bool) {
+		if input.Name == "Bitmaps" {
+			err := e.executeBitmaps(ctx, index, input, slice, func(bm *Bitmap, first bool) {
 				if i == 0 && first {
 					other = bm
 				} else {
@@ -848,7 +848,7 @@ func (e *Executor) executeUnionSlice(ctx context.Context, index string, c *pql.C
 	return other, nil
 }
 
-func (e *Executor) executeBitmapRange(ctx context.Context, index string, c *pql.Call, slice uint64, callback func(bm *Bitmap, first bool)) error {
+func (e *Executor) executeBitmaps(ctx context.Context, index string, c *pql.Call, slice uint64, callback func(bm *Bitmap, first bool)) error {
 	const columnLabel = "columnID"
 	const rowLabel = "rowID"
 
--- a/executor.go
+++ b/executor.go
@@ -821,19 +821,85 @@ func (e *Executor) executeFieldRangeSlice(ctx context.Context, index string, c *
 func (e *Executor) executeUnionSlice(ctx context.Context, index string, c *pql.Call, slice uint64) (*Bitmap, error) {
 	other := NewBitmap()
 	for i, input := range c.Children {
-		bm, err := e.executeBitmapCallSlice(ctx, index, input, slice)
-		if err != nil {
-			return nil, err
-		}
-
-		if i == 0 {
-			other = bm
+		if input.Name == "BitmapRange" {
+			err := e.executeBitmapRange(ctx, index, input, slice, func(bm *Bitmap, first bool) {
+				if i == 0 && first {
+					other = bm
+				} else {
+					other = other.Union(bm)
+				}
+			})
+			if err != nil {
+				return nil, err
+			}
 		} else {
-			other = other.Union(bm)
+			bm, err := e.executeBitmapCallSlice(ctx, index, input, slice)
+			if err != nil {
+				return nil, err
+			}
+			if i == 0 {
+				other = bm
+			} else {
+				other = other.Union(bm)
+			}
 		}
 	}
 	other.InvalidateCount()
 	return other, nil
+}
+
+func (e *Executor) executeBitmapRange(ctx context.Context, index string, c *pql.Call, slice uint64, callback func(bm *Bitmap, first bool)) error {
+	const columnLabel = "columnID"
+	const rowLabel = "rowID"
+
+	frame, _ := c.Args["frame"].(string)
+	if frame == "" {
+		frame = DefaultFrame
+	}
+
+	// Return an error if both the row and column label are specified.
+	rowIDs, rowOK := c.Args[rowLabel]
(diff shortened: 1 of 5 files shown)


travisturner:
What do you think about introducing something that transforms/optimizes a `Call` so that we don't require so much range-specific code in the executor?

It seems to me that if we had a function which converts the contents of a `Call`, then everything downstream could work just the same.

In that case, 
```
Union(Bitmaps(frame="data", rowID=[7~9,12])
```
might work like this:
```
c := Call{
    Name: "Union",
    Children: [
        Call{
            Name: "Bitmaps",
            Args: RowID=[7,8,9,12],
        },
    ],
}

c = optimize(c)
```
where `optimize()` explodes all instances of `Bitmaps` calls into a slices of `Bitmap` calls. In that case, `c` now looks like:
```
Call{
    Name: "Union",
    Children: [
        Call{
            Name: "Bitmap",
            Args: RowID=7,
        },
        Call{
            Name: "Bitmap",
            Args: RowID=8,
        },
        Call{
            Name: "Bitmap",
            Args: RowID=9,
        },
        Call{
            Name: "Bitmap",
            Args: RowID=12,
        },
    ],
}
```

Caveats here being that I don't know how this would affect performance, and I don't think `optimize` is a good name for the function. I also used `Bitmaps` instead of `BitmapRange` in my example in order to avoid yet another use of `range` in Pilosa, but I think that is open to discussion.

yuce:
I also think we overloaded "range" a lot and `Bitmaps` is a good name, I'll rename that call.

I think current code is more efficient and simple; in case the user doesn't use `Bitmaps`, s/he pays the price for a single `if` (which should be close to 0?). Another advantage is, I think the required memory is lower, since a single call is kept in memory instead of a lot more. The disadvantage with the current code is it should be introduced for each call and it can't be used standalone, like `Bitmaps(frame="data", rowID=[7~9,12])`. Not sure whether that is very useful though.

tgruben:
I love the concise feel of the range operator, very powerful. My preference to the implementation is the route @travisturner has suggested, for it would avoid touching existing functions and just adding a single operation, which in my opinion is easier to manage.  Since this feels kinda like a macro expansion to me, it might be appropriate to manipulate the query at the parser level.  I'm thinking split up the parse function to be something like

tgruben:
Side note and probably for another ticket, the range pattern expansion could provide a more concise expression on mutations as well. For example
`SetBit(row=[10~30,41,55],frame="data",col=[1,3,5,8])`

jaffee:
I kinda think this should be implemented within the standard Bitmap call. We change the existing logic around this call to always treat it as being a list of rows (any existing valid call would return a list of length 1), and then the ability to support ranges and such falls out of that.

travisturner:
I'm ok with modifying the `Bitmap` call to support a list of rows. 

There are a few decisions that need to be made about how this is ultimately implemented:
- Do `Bitmap()` calls always return a list of bitmaps to the caller? For example, if I perform the PQL `Bitmap(frame=f, rowID=[10~20]), do we return all (potentially) 11 bitmaps to the client?
- Where does the row list first get considered/exploded?
   - This PR currently has that happening in every `execute*Slice()` method, which feels like it could become messy.
   - We could explode the list by passing every `Call` through a function which looks for `Bitmap` with a list of rows, which is what I proposed. The downside here is that that check may affect performance (for example a stream of `SetBit` queries which will never contain `Bitmap`) but still get passed through the function.
   - Modify `roaring.Union` (and similar functions) to take a slice of Bitmaps
   - Add a new function like `roaring.Union` which takes a slice of Bitmaps, and call that in `executeUnionSlice` in the case where `Bitmap` list contains more than one row. In this case, again we would have a conditional in every `execute*Slice`, but at least the exploding wouldn't be happening there.

alanbernstein:
How about `29:32` instead of `29~32`? It matches conventional slice syntax, instead of using `~` in an unconventional way. Also if we ever have a reason to indicate negation in PQL syntax, `~` is a good option (`!` is also good for that, but having both of those, and meaning different things, might be confusing). We might then want to make the upper limit noninclusive to match convention.

jaffee:
@alanbernstein I was gonna suggest we use dash instead of tilde, but I like your idea even better. After some reflection, I don't think we need to make the upper limit noninclusive - these are not array or slice indices, it is simply a specification of some numbers. Noninclusivity has some nice properties when you're talking about indices - especially if the thing is 0 indexed, but I don't think most of them apply here.

@travisturner  There are definitely some things to think through here. We're adding a feature where a user can specify a whole lot of work in just a few characters. Exploding the query at the PQL or AST level would be a pretty simple implementation that would get us off the ground. I'd advocate for this at the moment - I think we could build it into the parser in a way that wouldn't affect parsing performance for existing queries too much.

The downside to that approach becomes apparent when a user does a query like `[1:10000]`. A more sophisticated implementation would probably fetch the bitmaps in batches and optimize the order in which they were processed by intersects/unions or something, but frankly I don't think we should get into it right now. Let's do the straightforward thing and see how it feels.

To answer the first question, I think that Bitmap calls always expand, so a call to Bitmap([1 2]) at the top level looks like Bitmap(1)Bitmap(2). This will be a little wonky for software which is appending multiple PQL calls to the same request and then trying to process the results, but works fine as long as you aren't mixing and matching different types of calls in the same request. Really it works fine always, the software just has to be aware of how many calls a Bitmap call expands to if it's trying to match results to calls by index.

yuce:
I've first tried to implement `BitmapRange` in the parser, but that requires making the parser aware of `rowID`, `columnID` and validate them (e.g., they are mutually exclusive) which I thought was the job of the interpreter.

tgruben:
I believe a new protobuf message would have to be added, possibly named BitmapSet, if we opted for the list of bitmaps option, for currently we can only return Bitmaps,TopN(pairs), and a Count/Sums across the cluster. Doing so at the parser would have the benefit of not requiring that additional handling.  I'm still note sold either way but good stuff here.

jaffee:
I think we should close this now - new work on #1482 and related tickets should be informed by the needs of #1029 though.

PR #1145, opened 2018-02-23
add request/response tags to the docs

travisturner:


docs/api-reference.md
```diff
@@ -45,19 +39,10 @@ Response:
...
-
-Request:
-```
-curl localhost:10101/index/user \
-     -X POST \
-     -d '{"options": {"timeQuantum": "YMDH"}}'
-```
-
-Response:
+`` request
```
alanbernstein: missing a \` here

PR #1153, opened 2018-03-01
[WIP] [POC] Added API struct; moved query logic to API

### LINKED ISSUE #1151: Refactor http handler
### Description
Currently there is a lot of Pilosa logic embedded in handler.go. If we want to be able to do things like embedding pilosa #1057 or having alternate interfaces such as GRPC #642, we need to move common logic out of handler.go. 

For example, let's take `handlePostIndex`:
1. gets index name out of the http path
2. decodes the request body
3. calls holder.CreateIndex
4. calls Broadcaster.SendSync
5. encodes a response
6. increments stats

If we just make the handler responsible for HTTP, we might have a function which looks like `func CreateIndex(index string, body io.Reader) error` This function would decode the request body (content type info may need to be passed along, but the encoding (pbuf or json) should be distinct from the transport (http in this case)). Then it would create the index, call broadcaster.sendsync, and update the stats. The postIndexResponse is empty, but one could imagine this function returning that along with the error. The other missing piece is making sure that the correct HTTP error code is returned in the case of an error. This will depend on us implementing some notion of error codes in Pilosa and then having a mapping of Pilosa errors onto HTTP errors (or any other transport).

I think this work can be done a few endpoints at a time to avoid giant PRs full of merge conflicts.

Once it's done, we can move all http handling code into a subpackage, in the spirit of #1093 (and as suggested by Ben Johnson's package management strategy).

handler.go is responsible for http handling and little else. It no longer imports encoding/json, protobuf, pilosa/pql, pilosa/internal, etc.

yuce:
## Overview

This shows an approach to implement https://github.com/pilosa/pilosa/issues/1151

Fixes #1151 

## Pull request checklist

## Code review checklist

### DIFF
--- a/api.go
+++ b/api.go
@@ -1,0 +1,101 @@
+// Copyright 2017 Pilosa Corp.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pilosa
+
+import (
+	"context"
+	"strings"
+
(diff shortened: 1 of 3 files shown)


jaffee:
I am reviewing this and rebasing it onto master as I go.

PR #1161, opened 2018-03-06
Creates a Diagnostics interface to remove dependency from pilosa package.

### LINKED ISSUE #1094: Invert diagnostics dependency
See parent https://github.com/pilosa/pilosa/issues/1093

travisturner:
## Overview

Creates a Diagnostics interface to remove dependency from pilosa package.

Fixes #1094

## Pull request checklist

## Code review checklist

### DIFF
--- a/diagnostics.go
+++ b/diagnostics.go
@@ -19,6 +19,9 @@ import (
 	"time"
 )
 
+// Ensure nopDiagnostics implements interface.
+var _ Diagnostics = &nopDiagnostics{}
+
 // Diagnostics represents an interface for sending diagnostics to a consumer.
 type Diagnostics interface {
 	SetVersion(string)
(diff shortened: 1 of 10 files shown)


diagnostics.go
```diff
@@ -0,0 +1,73 @@
...
+func init() {
+	NopDiagnostics = &nopDiagnostics{}
+}
+
+// NopDiagnostics represents a Diagnostics that doesn't do anything.
+var NopDiagnostics Diagnostics
+
+type nopDiagnostics struct{}
+
+// SetVersion is a no-op implemenetation of Diagnostics SetVersion method.
```
codysoyland: Check spelling (implemenetation) -- several instances of this.

diagnostics/diagnostics.go
```diff
@@ -27,28 +27,30 @@ import (
...
 	"time"
 
+	"github.com/pilosa/pilosa"
 	"github.com/shirou/gopsutil/host"
 	"github.com/shirou/gopsutil/mem"
 	"github.com/sony/gobreaker"
 )
 
+// Ensure ActiveDiagnostics implements interface.
+var _ pilosa.Diagnostics = &ActiveDiagnostics{}
```
codysoyland: Apply this check to the NopDiagnostics too.

travisturner: applied in 2f74fd4aad3a4e247d166bdd76b6181662be5443

diagnostics/diagnostics.go
```diff
@@ -27,28 +27,30 @@ import (
...
 
 type versionResponse struct {
 	Version string `json:"version"`
 	Message string `json:"message"`
 }
 
-// Diagnostics represents a client to the Pilosa cluster.
-type Diagnostics struct {
+// ActiveDiagnostics represents a client to the Pilosa cluster.
+type ActiveDiagnostics struct {
```
jaffee: this naming breaks the stuttering rule (since the package is now diagnostics, when pilosa/server imports and uses it, we get `diagnostics.NewActiveDiagnostics`. The solution here is not to break up our packages by functionality, but by dependency. So the subpackage `diagnostics` should actually be named `gopsutil`, and contain a struct called `Diagnostics` and a `NewDiagnostics` constructor (if necessary). This, of course, is not my original idea, but @benbjohnson's (or at least that's where I learned it), but it seems to work quite well in practice, and I've had some success with it in the PDK. Then if we wanted another implementation of diagnostics, it would live in another package named by dependency, or if it had no external dependencies (i.e. if it was a stub or mock implementation, it could live in `pilosa/diagnostics.go` alongside the interface definition.

codysoyland:
I find it confusing to name our package `gopsutil`. I've never heard of a naming convention in which a package is named after it's dependencies.

I'm also not familiar with the stuttering rule, but it seems hard to avoid having the same name appear near each other in code. I mean just look at our import path.

If it's important that the name of the package and the name of the constructor be different, why not rename the Diagnostics struct to something like "Sender": `diagnostics.NewSender()`?

jaffee:
The stuttering rule comes from: https://talks.golang.org/2014/names.slide#12 and is fairly widely accepted by Go code review guides, etc.

The package layout convention comes from https://medium.com/@benbjohnson/standard-package-layout-7cdbc8391fc1 and is also pretty widely known although less widely adopted in my experience. If you ask about package layout on gopher slack, you're almost guaranteed to get a link to it.

In this case, it probably doesn't matter if we call the package diagnostics or something else (I don't expect us to have other implementations of it), but if we did, then things might get confusing. I've found this pattern to be very helpful in the PDK where we have multiple implementations of (e.g.) Translator using various key value stores.

codysoyland:
Thanks for the links. I'm aware of the general patterns you're referencing, but I don't think they apply well with this particular interface abstraction.

I think that packages should be functional groupings of code, with a name that can be seen as a category. I don't like the idea of diagnostics code being inside a package called gopsutil because diagnostics has nothing to do with that package name.

The motivation of this PR is not to build an interface around diagnostics; it is to remove gopsutil and gobreaker dependencies from pilosa core. One way we can solve this is by wrapping the functionality we get from `gopsutil` and `gobreaker` into small interfaces whose implementations are in subpackages, and move the main `Diagnostics` struct back into pilosa core (I also don't expect to have more than one implementation of it). I think we can remove the gobreaker dep actually -- it doesn't give us much. The gopsutil stuff could go in a wrapper called `SystemInfoGetter` or something. We can then reuse that for providing that data to the WebUI in the future.

jaffee:
I think you're right @codysoyland - diagnostics does more than just the gopsutil stuff - keeping it in `pilosa/pilosa`, and just wrapping `gopsutil` as a provider of `SystemInfo` makes a lot of sense.

codysoyland:
I implemented my proposal here: https://github.com/pilosa/pilosa/pull/1166

travisturner:
closing in favor of #1166

PR #1178, opened 2018-03-22
Clean up logger; make it honor --log-path flag.

### LINKED ISSUE #631: Some logs respect --log-path option and some do not
### Expected behavior
If `--log-path=filename` is passed, all log output will go to `filename` and not stdout or stderr.

### Actual behavior
This goes to the log file:
```
2017/06/12 15:14:38 holder sync monitor initializing (10m0s interval)
```

This goes to stdout or stderr:
```
Pilosa v0.4.0-12-g74553d8, build time 2017-06-12T15:14:33+0000
Using data from: /home/ubuntu/pilosa/data1
Listening as http://node1.jaffee28.sandbox.pilosa.com:10101
```

`pilosa server --log-path=somefile` 
Then see what goes to the file, and what goes to your terminal.

Happens on linux and osx, both cluster and single node.

travisturner:
## Overview

This PR adds a `Logger` interface so all objects share the same logger.
It also adds functional options to `NewGossipMemberSet` to support supplying a `log.Logger` and/or a `Transport`.

Fixes #631 

## Pull request checklist

## Code review checklist

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -43,7 +43,7 @@ func NewServeCmd(stdin io.Reader, stdout, stderr io.Writer) *cobra.Command {
 		Long: `pilosa server runs Pilosa.
 
 It will load existing data from the configured
-directory, and start listening client connections
+directory and start listening for client connections
 on the configured port.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			logOutput, err := server.GetLogWriter(Server.Config.LogPath, stderr)
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -17,7 +17,6 @@ package cmd
 import (
 	"fmt"
 	"io"
-	"log"
 	"os"
 	"os/signal"
 	"runtime/pprof"
@@ -46,11 +45,11 @@ It will load existing data from the configured
 directory and start listening for client connections
 on the configured port.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			logOutput, err := server.GetLogWriter(Server.Config.LogPath, stderr)
-			if err != nil {
-				return err
+			// Set up the logger.
+			if err := Server.SetupLogger(); err != nil {
+				return fmt.Errorf("error setting up the logger: %v", err)
 			}
-			logger := log.New(logOutput, "", log.LstdFlags)
+			logger := Server.Server.Logger
 			logger.Printf("Pilosa %s, build time %s\n", pilosa.Version, pilosa.BuildTime)
 
 			// Start CPU profiling.
--- a/fragment.go
+++ b/fragment.go
@@ -26,7 +26,6 @@ import (
 	"hash"
 	"io"
 	"io/ioutil"
-	"log"
 	"net/http"
 	"os"
 	"sort"
@@ -103,8 +102,8 @@ type Fragment struct {
 	// so that they can be mmapped and heap utilization can be kept low.
 	MaxOpN int
 
-	// Writer used for out-of-band log entries.
-	LogOutput io.Writer
+	// Logger used for out-of-band log entries.
+	Logger Logger
 
 	// Row attribute storage.
 	// This is set by the parent frame unless overridden for testing.
@@ -124,8 +123,8 @@ func NewFragment(path, index, frame, view string, slice uint64) *Fragment {
 		CacheType: DefaultCacheType,
 		CacheSize: DefaultCacheSize,
 
-		LogOutput: ioutil.Discard,
-		MaxOpN:    DefaultFragmentMaxOpN,
+		Logger: NopLogger,
+		MaxOpN: DefaultFragmentMaxOpN,
 
 		stats: NopStatsClient,
 	}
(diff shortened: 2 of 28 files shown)


logger.go
```diff
@@ -0,0 +1,85 @@
...
+	NopLogger = &nopLogger{
+		logger: log.New(ioutil.Discard, "", log.LstdFlags),
+	}
+}
+
+// NopLogger represents a Logger that doesn't do anything.
+var NopLogger Logger
+
+type nopLogger struct {
+	logger *log.Logger
```
jaffee: why does nopLogger need this?

server.go
```diff
@@ -111,9 +111,9 @@ func NewServer() *Server {
 		MetricInterval:      0,
 		DiagnosticInterval:  0,
 
-		LogOutput: os.Stderr,
+		LogOutput: ioutil.Discard,
```
jaffee: can we remove LogOutput entirely now?

travisturner: I left that there because `func (m *Command) Close()` uses it to close the writer. I didn't want to make `Close()` part of the `Logger` interface, but I'm not clear on the best way to close that writer.

logger.go
travisturner: good catch. it doesn't need that anymore since i removed `Logger()` from the interface.

server.go
jaffee: I'm not sure why that's there - it wouldn't be used normally since Stdout/Stderr don't implement io.Closer (afaik). What we could do is see if Server.Logger is an io.Closer and if so, call Close on it, though I'd prefer to remove it entirely. @yuce Do you know why Command.Close tries to call Close on Server.LogOutput?

```go
// Close shuts down the server.
func (m *Command) Close() error {
	var logErr error
	serveErr := m.Server.Close()
	logOutput := m.Server.LogOutput
	if closer, ok := logOutput.(io.Closer); ok {
		logErr = closer.Close()
	}
...
```

travisturner: That makes sense to me, and I just tested it with an implementation of `Logger` that also implements `io.Closer()` and that works as expected. So the idea being if someone needed their logger to close, it would get called here. I'll push a commit for that change.

travisturner: I made that change in ff25e2e01bb7c17d71b7d9dff6b68d1ccd02f77a

cmd/server.go
```diff
@@ -43,7 +43,7 @@ func NewServeCmd(stdin io.Reader, stdout, stderr io.Writer) *cobra.Command {
 		Long: `pilosa server runs Pilosa.
 
 It will load existing data from the configured
-directory, and start listening client connections
+directory and start listening for client connections
 on the configured port.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			logOutput, err := server.GetLogWriter(Server.Config.LogPath, stderr)
```
jaffee: so, I think that if this actually opens a file, that file won't get closed when Pilosa exits. GetLogWriter is also called inside Command.Run which is kinda weird - that file might get opened twice? I don't think it gets closed either time right now. 

I think GetLogWriter should save the logger on the server.Command, and be responsible for closing any file it may have opened when Command.Close is called.

fragment.go
```diff
@@ -124,8 +123,8 @@ func NewFragment(path, index, frame, view string, slice uint64) *Fragment {
 		CacheType: DefaultCacheType,
 		CacheSize: DefaultCacheSize,
 
-		LogOutput: ioutil.Discard,
-		MaxOpN:    DefaultFragmentMaxOpN,
+		Logger: NopLogger,
```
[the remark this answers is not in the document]
travisturner: @tgruben I'm not sure I understand the context of the qustion. Do you mean something like this?
```
f := NewFragment(".", "i", "f", "v", 0)
f.Logger = NewMyCustomLogger()
```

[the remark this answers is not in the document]
tgruben: no, it just looks to me like if I turn on logging in the config, nothing is going to be logged, for it is set to Nop, it is just unclear to me the path from config to instantiation.  Perhaps just a documentation issue?    The use case is, I want to log to use the log destination defined by the config.  Should I read the config in the fragment or use some convenience method to reference an already configured logger destination.

view.go
```diff
@@ -275,7 +269,7 @@ func (v *View) newFragment(path string, slice uint64) *Fragment {
 	frag := NewFragment(path, v.index, v.frame, v.name, slice)
 	frag.CacheType = v.cacheType
 	frag.CacheSize = v.cacheSize
-	frag.LogOutput = v.LogOutput
+	frag.Logger = v.Logger
```
jaffee: @tgruben The fragment's logger gets set to the view's logger on fragment creation.

tgruben: Please ignore the crazy man behind the curtain, i was looking at wrong version of view.go.  Nothing to see here, move on :)

travisturner:
The latest commit 17f58282a81b184a683881faff8efe507e561568 addresses the log file handling by moving the `Close()` responsibility to the `Logger` implementations.

tgruben:
It seems my subconscious self just wants our standard logger to be buffered, probably not a big deal but I would prefer buffered file io, just because I had many cpu issues ages ago writing directly to the file in logging type scenarios, granted we're not logging that much.  Just my 2cents, no biggie either way

PR #1179, opened 2018-03-23
Default gossip seed should be empty instead of local bind address

### LINKED ISSUE #1172: Failure to connect to gossip seed with default config on single-node cluster (no route to host)
### Expected behavior

Starting "pilosa server" with no arguments should bring up a running single-node cluster without logged errors.

### Actual behavior

Since default gossip seed is localhost:14000, it tries to connect to that even with a single node cluster. The following error is logged (does not seem to affect performance):
```
...
2018/03/19 15:00:04 [DEBUG] memberlist: Failed to join fe80::1: dial tcp [fe80::1]:14000: connect: no route to host
...
```

Run `pilosa server` with no config.

travisturner:
## Overview

There is no need to set the gossip seed to the local node's bind address (e.g. in the case of a one-node cluster). This PR removes that default, and moves the logic to the `Seeds()` method which is actually only used for testing.

Fixes #1172 

## Pull request checklist

## Code review checklist

### DIFF
--- a/gossip/gossip.go
+++ b/gossip/gossip.go
@@ -61,12 +61,11 @@ func (g *GossipMemberSet) Start(h pilosa.BroadcastHandler) error {
 	return nil
 }
 
-// Seeds returns the gossipSeeds determined by the config.
-func (g *GossipMemberSet) Seeds() []string {
-	if len(g.config.gossipSeeds) == 0 {
-		return []string{fmt.Sprintf("%s:%d", g.config.memberlistConfig.BindAddr, g.config.memberlistConfig.BindPort)}
-	}
-	return g.config.gossipSeeds
+// GetBindAddr returns the gossip bind address based on config and auto bind port.
+// This method is currently only used in a test scenario where a second node needs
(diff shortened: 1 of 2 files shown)


gossip/gossip.go
```diff
@@ -63,6 +63,9 @@ func (g *GossipMemberSet) Start(h pilosa.BroadcastHandler) error {
 
 // Seeds returns the gossipSeeds determined by the config.
```
codysoyland: Might be good to clarify the function comment to something like: "Seeds returns a list of usable gossip seeds for this cluster. If no seeds are configured, return the gossip listener of this server."

Or otherwise comment on the `if` statement to explain it.

travisturner: Thanks @codysoyland, that was definitely confusing. I pushed a commit which renames that method and doesn't use `gossipSeeds` at all. It's really only there for one test. Ideally we could get ride of that logic completely, but for now I just tried to make it more clear what was happening.

codysoyland: The latest commit is great 👍

PR #1180, opened 2018-03-26
Remove rowcol labels

### LINKED ISSUE #752: Proposal: Remove row and column labels
### Description

Anecdotally, arbitrary labelling of rows and columns seems to be more confusing than useful.
 I'm talking about when you create an index and frame, and you can optionally specify a column label an row label respectively to be used in queries in place of "rowID" and "columnID". Normally to retrieve a row or column, you'd use `Bitmap(rowID=<id>, frame=<frame>)` or `Bitmap(columnID=<id>, frame=<frame>)`, but if you specified row and column labels, you use those in place of "rowID" and "columnID".

The only value this provides is making queries read in a more relatable way to the data in Pilosa. On the flipside, it increases code complexity (both client and server side), makes it difficult to understand whether a query is accessing a row or column at a glance, and increases cognitive load in that you have to remember which label was used for each frame and index. It also may fragment the language which different groups use to discuss Pilosa and inhibit understanding between different users.

Therefore, in order to simplify Pilosa, and reduce our API surface area, I propose that we deprecate and eliminate the ability to specify arbitrary labels on rows and columns.

The purpose of this ticket is to gather community feedback to determine the level of disruption which this change would cause, and whether we don't understand any important uses of row and column labels.

Row and column labels have been removed from Pilosa, the documentation, and any examples or tools which were using them.

travisturner:
## Overview

This PR removes support for `RowLabel` and `ColumnLabel` and instead uses `row` and `col` in PQL.

Fixes #752 

## Pull request checklist

## Code review checklist


frame.go
```diff
@@ -32,7 +32,7 @@ import (
 
 // Default frame settings.
 const (
-	DefaultRowLabel       = "rowID"
+	DefaultRowLabel       = "row"
```
jaffee: shouldn't this setting not exist anymore?

index.go
```diff
@@ -31,7 +31,7 @@ import (
 
 // Default index settings.
 const (
-	DefaultColumnLabel = "columnID"
+	DefaultColumnLabel = "col"
```
jaffee: again, this goes away, right?

internal/private.proto
```diff
@@ -3,12 +3,10 @@ syntax = "proto3";
 package internal;
 
 message IndexMeta {
-	string ColumnLabel = 1;
```
jaffee: we'll need to try to synchronize client updates as well when we merge this, right?

frame.go
travisturner: Well, the constant needs to be defined somewhere (unless we want the string "row" scattered throughout the code). Is it that the name of the constant is confusing?  Perhaps changing it to `RowLabel` or something else makes more sense.

travisturner: I could also move those constants to `executor.go` since that's the only place they're used anymore.

jaffee: +1 on name change and moving

travisturner: moved and renamed in e3cd03d902174524de1082cff72c609b9342b8f6

jaffee:
nice finds @alanbernstein

travisturner:
Thanks @alanbernstein 
I removed most of those in ce4fb3c4d347eb66bc0682ca27592e73c2708c87
There was one case where those terms are still valid (in `ast.go`, in the function that determines if the query is meant for the inverse view), so I left that. Also, the instances in `handler_internal_test.go` where ok because they were meant to check for an invalid argument (which `rowLabel` and `columnLabel` are now), but I just changed the string in the test to `badKey` to avoid confusion.

internal/private.proto
jaffee: my fear is unfounded here... this is just internal comms

PR #1190, opened 2018-04-09
1151 http refactor

### LINKED ISSUE #1151: Refactor http handler

jaffee:
## Overview

Rebased yuce's 1151 work onto master after cluster-resize was merged. Lot's of conflicts and work needed to be done to get tests passing. Concerned that untested functionality may have been missed. 

Examples:
- Recalculate caches gets broadcast to all nodes - this got stomped by the rebase, had to re-add.
- I left the InputJSONDataParser handler commented out, and no tests complain.
- maxslices endpoint changed - had to be redone.
- status/schema endpoints

Fixes #1151 

Update: I'm satisfied with this now, I reviewed the API against all changed to handler.go since v0.8.8 and I don't think we missed anything - let's get it merged ASAP to avoid more merge conflicts. I've identified some follow-on work around errors, and I think /version and /debug/vars need to be pulled into API as well, but I'm of the opinion that we can do that separately.

## Pull request checklist

## Code review checklist


yuce:
I get `rowID required` errors when I go client tests against this branch. Should the `master` merged into this branch?

yuce:
Merging `master` into this branch fixes `rowID required` errors. There wasn't any merge conflicts, all Pilosa server and Go client tests pass.

jaffee:
thanks yuce - I merged master in and pushed that up

test/handler.go
```diff
@@ -40,10 +40,12 @@ func NewHandler() *Handler {
 	h := &Handler{
 		Handler: pilosa.NewHandler(),
 	}
-	h.Handler.Executor = &h.Executor
+	h.API = pilosa.NewAPI()
+	h.Handler.API = pilosa.NewAPI()
```
travisturner: should this be `h.Handler.API = h.API` instead? perhaps it doesn't matter, but I'm wondering if having those separate might cause confusion in the future.

handler.go
```diff
@@ -2133,83 +1734,82 @@ type clusterResizeAbortResponse struct {
...
-				return nil, fmt.Errorf("error handling action: %s, err: %s", action.ValueDestination, err)
-			}
-			if bit != nil {
-				setBits[frame] = append(setBits[frame], bit)
-			}
-		}
-	}
-	return setBits, nil
-}
+// // InputJSONDataParser validates input json file and executes SetBit.
```
travisturner: any reason to keep this commented code?

jaffee: Ah yes -  are we OK with aggressively deprecating input definition? This is basically just me being lazy, although the fact that all our tests pass without this code is probably indicative of something...

test/handler.go
jaffee: I think you're right

handler.go
travisturner: it looks like that method became `inputJSONDataParser` in `api.go`

jaffee: ohh, I see what happened now. For some reason I thought that the handler just hadn't been ported, but it was a helper function sneaking back in. I removed it.

PR #1207, opened 2018-04-16
Remove SecurityManager. Implement api restrictions in api package.

### LINKED ISSUE #1193: move SecurityManager functionality into API
### Description

Having merged support for #1151. The Handler is becoming a fairly thin wrapper which is simply the HTTP interface to Pilosa's actual API. Continuing this work is #1152 which will actually move that http wrapping functionality into a subpackage, and allow us to write other wrappers (like GRPC) which will be more performant and open Pilosa up for new use cases.

The current handler implements pilosa.SecurityManager which is used by pilosa.Cluster to control access to different API calls when the clusters is in restricted states (e.g. resizing). It uses features of the http router to implement this functionality. Because every API wrapper we write will need to have this exact functionality, the HTTP handler is no longer a reasonable place to do this. We need to move this into the API layer to simplify wrappers and reduce duplication.

As a side note, there is a circular reference situation going on here, where Handler has a reference to a Cluster, and the Cluster holds a reference back to the handler via the prefect field (SecurityManager). 

Access control based on cluster state is handled by API or lower, and external interfaces don't need to concern themselves with it.

travisturner:
## Overview

This PR removes the `SecurityManager` interface and moves the api restrictions (based on cluster state) out of the handler and into the api.

I don't love the way that `validate()` is called inside every api function. If there's a cleaner way to do this I'd appreciate the feedback.

Fixes #1193 

## Pull request checklist

## Code review checklist


api.go
```diff
@@ -59,8 +59,30 @@ func NewAPI() *API {
 	}
 }
 
+// functionStates specifies the api functions that are valid for each
+// cluster state.
+var functionStates = map[string][]int{
```
travisturner: This is a terrible name. Plus, those are methods, not functions. I'll clean up the naming conventions here if the overall approach is acceptable.

handler_test.go
```diff
@@ -160,7 +160,7 @@ func TestHandler_ClusterResizeAbort(t *testing.T) {
 	t.Run("No resize job", func(t *testing.T) {
 		h := test.NewHandler()
 		h.API.Cluster = test.NewCluster(1)
-		h.SetRestricted()
+		h.API.Cluster.SetState(pilosa.ClusterStateResizing)
```
jaffee: whoa, gross - Cluster should not be exported on API. Probably, this test should be moved into api_test.go and be in `package pilosa` so that it can set the unexported state.

api.go
```diff
@@ -992,3 +1142,91 @@ func (api *API) State() string {
...
+	apiClusterMessage,
+	apiMarshalFragment,
+	apiSetCoordinator,
+}
+
+var functionResizing = []int{
+	apiResizeAbort,
+}
+
+var functionNormal = []int{
```
jaffee: did you consider a `map[int]struct{}` for these? Perf difference is probably negligible, but it might be a bit more obvious what it's for at a glance.

travisturner: I don't quite understand this one. What does the map represent?

jaffee:
Just use a map as a set and get O(1) lookups instead of looping

handler_test.go
travisturner: I suggest we make this a new ticket.

jaffee: yep - will do

api.go
```diff
@@ -59,8 +61,39 @@ func NewAPI() *API {
...
+	}
+	for k, v := range b {
+		r[k] = v
+	}
+	return r
+}
+
+func (api *API) validate(f apiFunc) error {
+	state := api.Cluster.State()
+	if _, ok := functionStates[state][f]; ok {
```
jaffee: I think I'd rename `functionStates` to `validAPIMethods` or `validAPIFuncs` if we're going to stick with function terminology everywhere (e.g. the `apiFunc` type)

travisturner: oops. yeah, I forgot to change `func` to `method`. thanks for the reminder. I'll rename this while I'm at it.

PR #1242, opened 2018-05-02
Make sure ~ is expanded in NewServer; BroadcastReceiver uses temp path

yuce:
## Overview

Fixes #

## Pull request checklist

## Code review checklist

### DIFF
--- a/broadcast_test.go
+++ b/broadcast_test.go
@@ -56,9 +56,14 @@ func testMessageMarshal(t *testing.T, m proto.Message) {
 
 // Ensure that BroadcastReceiver can register a BroadcastHandler.
 func TestBroadcast_BroadcastReceiver(t *testing.T) {
+	path, err := ioutil.TempDir("", "pilosa-")
+	if err != nil {
+		panic(err)
+	}
 	com := server.NewCommand(bytes.NewBuffer([]byte{}), ioutil.Discard, ioutil.Discard)
 	com.Config.Bind = "localhost:0"
-	err := com.SetupServer() // this test shouldn't need to import pilosa/server just to set up the Server, but it really shouldn't need to setup the Server at all. The Server should not be the implementation of Broadcast* TODO
+	com.Config.DataDir = path
+	err = com.SetupServer() // this test shouldn't need to import pilosa/server just to set up the Server, but it really shouldn't need to setup the Server at all. The Server should not be the implementation of Broadcast* TODO
 	if err != nil {
 		t.Fatalf("setting up server: %v", err)
 	}
--- a/server.go
+++ b/server.go
@@ -21,6 +21,7 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"runtime"
 	"strconv"
 	"strings"
@@ -230,6 +231,11 @@ func NewServer(opts ...ServerOption) (*Server, error) {
 		}
 	}
 
+	err := s.expandDataDirName()
+	if err != nil {
+		return nil, err
+	}
+
 	s.Holder.Logger = s.logger
 	s.Holder.Stats.SetLogger(s.logger)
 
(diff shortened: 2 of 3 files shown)


server.go
```diff
@@ -260,6 +266,18 @@ func NewServer(opts ...ServerOption) (*Server, error) {
 	return s, nil
 }
 
+func (s *Server) expandDataDirName() error {
```
travisturner: It seems like making this method specific to `s.Holder.Path` is not ideal.
What do you think about making it more generic (so that it simply expands a `dir` that is provided).
Then, for example, it could be applied to the `dir` argument supplied to `OptServerDataDir()`. And that way it would be correct for both `s.Holder.Path` and `s.Cluster.Path`. Does that seem right to you?

yuce: That makes sense. Updated the PR.

server.go
```diff
@@ -230,9 +231,21 @@ func NewServer(opts ...ServerOption) (*Server, error) {
 		}
 	}
 
+	path, err := expandDirName(s.Holder.Path)
```
travisturner: I thought we could call `expandDirName()` inside `OptServerDataDir()` like this:
```
 func OptServerDataDir(dir string) ServerOption {
 	return func(s *Server) error {
 		path, err := expandDirName(dir)
 		if err != nil {
 			return err
 		}
 		s.Cluster.Path = path
 		s.Holder.Path = path
 		return nil
 	}
 }
```
Does that work?

yuce: What if `OptServerDataDir` is never called?

travisturner: That's a good point. I forgot that our default data-dir has a `~` in it.

travisturner: It looks like `OptServerDataDir()` is always called when setting up a server, otherwise that value will be an empty string, so I think it's ok to have `expandDirName()` inside `OptServerDataDir()`.

yuce: We may later have another `Opt*` function which sets the path, then we have to duplicate the expansion code (or just forget to add it). I think it's best to do the expansion after all `Opt*` functions run unless there's a reason not to do so.

broadcast_test.go
```diff
@@ -56,9 +56,14 @@ func testMessageMarshal(t *testing.T, m proto.Message) {
 
 // Ensure that BroadcastReceiver can register a BroadcastHandler.
 func TestBroadcast_BroadcastReceiver(t *testing.T) {
+	path, err := ioutil.TempDir("", "pilosa-")
```
jaffee: setting up new servers/clusters/etc. for testing should be handled by the `pilosa/test` package. It may need to be modified slightly to support this case, but let's not have pretty generic setup code like this in tests.

server.go
jaffee: either we need to add a `dataDir string` field to Server, and do the expansion, `Cluster.Path`, and `Holder.Path` in `NewServer`, or we can handle everything in the `Opt` as travis suggested. I don't think it's too likely that we'll need another option that touches the data dir, but I've been surprised before.

broadcast_test.go
yuce: It is possible that the holder dir and cluster dir is different; e.g., in `test/holder.go` holder path is set directly. So I think the implementation in this PR is the safest one. Is there anything wrong with expanding the holder and cluster paths in `NewServer` ?

jaffee: There's only one path - the holder path and cluster path should never be different - the fact that there are two separate objects for holder and cluster that each have a path is an unfortunate implementation detail. We should only be calling expandDir once on the path that is passed to Server via OptServerDataDir, and setting it on the holder and cluster should be fully opaque to the user. tests may be a special case, but I'm not too concerned about that.

I don't care if we do the setup in the Opt or in NewServer.

server.go
yuce: Updated

yuce:
great!

yuce:
Github doesn't let me merge this PR without @travisturner 's approval.

PR #1243, opened 2018-05-02
WIP- removed mmap and cloning from fragment

tgruben:
## Overview

Just a quick build to see the performance and memory overhead of cloning and memory mapping

### DIFF
--- a/fragment.go
+++ b/fragment.go
@@ -219,10 +219,10 @@ func (f *Fragment) openStorage() error {
 			return err
 		}
 	}
-	f.storageData, err = ioutil.ReadAll(f.file)
+	data, err := ioutil.ReadAll(f.file)
 
 	// Attach the mmap file to the bitmap.
-	data := f.storageData
+	//data := f.storageData
 	if err := f.storage.UnmarshalBinary(data); err != nil {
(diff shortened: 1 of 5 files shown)


fragment.go
```diff
@@ -214,22 +214,12 @@ func (f *Fragment) openStorage() error {
...
-	if err != nil {
-		return fmt.Errorf("mmap: %s", err)
-	}
-	f.storageData = storageData
-
-	// Advise the kernel that the mmap is accessed randomly.
-	if err := madvise(f.storageData, syscall.MADV_RANDOM); err != nil {
-		return fmt.Errorf("madvise: %s", err)
-	}
+	f.storageData, err = ioutil.ReadAll(f.file)
```
travisturner: probably want to handle this error

fragment.go
```diff
@@ -320,9 +310,6 @@ func (f *Fragment) closeStorage() error {
 
 	// Unmap the file.
 	if f.storageData != nil {
```
travisturner: might be clear just to remove this if statement and always set `f.storageData = nil`

tgruben:
Closing this PR for now, Performance was improved at the cost of slower startup times and higher memory footprint.

PR #1254, opened 2018-05-07
Unexport addnode

tgruben:
## Overview

API cleanup


tgruben:
it wasn't deleted tests, it was just some mock plumbing for testing that is not used @jaffee

PR #1264, opened 2018-05-09
Minor Import optimization

tgruben:
## Overview

In situations where fragments contain large amounts of data, efficiency can be gained by building a small working bitmap and merging it as opposed to simply operating on the larger bitmap file.


tgruben:
I'm moving this to a pr for standard roaring import

PR #1326, opened 2018-05-24
Change bits terminolgy to column

### LINKED ISSUE #1315: Stop using "bits" in query results
Some queries return an attribute called `bits`.
This is generally a list of columns, and having them labeld as bits seems confusing.
Decide if `cols`, `columns`, or something else makes sense, and update all instances of `bits`.

```
{"results":[{"attrs":{},"bits":[100,200,300]}]}
```

There may be other places where we refer to `bits` and these should be addressed with this ticket.
Including documentation.

tgruben:
## Overview

Converting all our references from bit(s) to column(s) in order to clearify our message.

Fixes #1315


client.go
```diff
@@ -1140,7 +1140,7 @@ func (c *InternalHTTPClient) SendMessage(ctx context.Context, uri *URI, pb proto
 	return nil
 }
 
-// Bit represents the location of a single bit.
+// Bit represents the location of a single column.
 type Bit struct {
```
jaffee: i assume this was left as Bit intentionally?

fragment.go
```diff
@@ -478,30 +478,30 @@ func (f *Fragment) clearBit(rowID, columnID uint64) (changed bool, err error) {
...
 	pos, err := f.pos(rowID, columnID)
 	if err != nil {
 		return false, err
 	}
 	return f.storage.Contains(pos), nil
 }
 
-// FieldValue uses a column of bits to read a multi-bit value.
-func (f *Fragment) FieldValue(columnID uint64, bitDepth uint) (value uint64, exists bool, err error) {
+// FieldValue uses a column of columns to read a multi-column value.
```
jaffee: this makes no sense at all

fragment.go
```diff
@@ -478,30 +478,30 @@ func (f *Fragment) clearBit(rowID, columnID uint64) (changed bool, err error) {
...
+		return 0, false, errors.Wrap(err, "getting existence column")
 	} else if !v {
 		return 0, false, nil
 	}
 
-	// Compute other bits into a value.
-	for i := uint(0); i < bitDepth; i++ {
-		if v, err := f.bit(uint64(i), columnID); err != nil {
-			return 0, false, errors.Wrapf(err, "getting value bit %d", i)
+	// Compute other columns into a value.
```
jaffee: yeah, changing bit->column for bsi fields doesn't make sense. we need to revert

view_test.go
```diff
@@ -72,18 +72,18 @@ func (v *View) Reopen() error {
...
 	for _, columnID := range columnIDs {
 		if _, err := v.SetBit(rowID, columnID); err != nil {
 			panic(err)
 		}
 	}
 }
 
-// MustClearBits clears bits on a row. Panic on error.
-func (v *View) MustClearBits(rowID uint64, columnIDs ...uint64) {
+// MustClearColumns clears columns on a row. Panic on error.
```
jaffee: I don't like this name change - sounds like we're going to clear a bunch of columns vertically, but we're just clearing bits in a row.

view_test.go
```diff
@@ -72,18 +72,18 @@ func (v *View) Reopen() error {
 	return v.Open()
 }
 
-// MustSetBits sets bits on a row. Panic on error.
+// MustSetColumns sets columns on a row. Panic on error.
 // This function does not accept a timestamp or quantum.
-func (v *View) MustSetBits(rowID uint64, columnIDs ...uint64) {
+func (v *View) MustSetColumns(rowID uint64, columnIDs ...uint64) {
```
jaffee: setting bits makes more sense than setting columns I think

time_test.go
```diff
@@ -65,7 +65,7 @@ func TestViewByTimeUnit(t *testing.T) {
 	})
 }
 
-// Ensure all applicable frame names can be generated when mutating a time bit.
+// Ensure all applicable frame names can be generated when mutating a time column.
```
jaffee: time  column doesn't make sense

test/frame.go
```diff
@@ -75,7 +75,7 @@ func (f *Frame) Reopen() error {
 	return nil
 }
 
-// MustSetBit sets a bit on the frame. Panic on error.
+// MustSetBit sets a column on the frame. Panic on error.
```
jaffee: I don't think this should be changed.

executor.go
```diff
@@ -1319,7 +1319,7 @@ func (e *Executor) executeSetRowAttrs(ctx context.Context, index string, c *pql.
 	if err := frame.RowAttrStore().SetAttrs(rowID, attrs); err != nil {
 		return err
 	}
-	frame.Stats.Count("SetBitmapAttrs", 1, 1.0)
+	frame.Stats.Count("SetColumnAttrs", 1, 1.0)
```
jaffee: isn't this `SetRowAttrs`?!

PR #1337, opened 2018-05-26
[WIP] ID Translation

benbjohnson:
## Overview

This pull request adds id-to-key translation to make it easier for users
to provide non-integer identifiers for rows & columns.

_NOTE: This is a work-in-progress. Do not merge._

### TODO

- [x] Translator
- [x] TranslateStore
- [ ] Executor integration
- [ ] `main` integration

## Pull request checklist

## Code review checklist


benbjohnson:
@travisturner I refactored out the `Translator` and merged it into `Executor`. It seems to fit a lot better there. However, we need some kind of flag regarding whether to translate a query or not. Are we adding a flag per-frame or per-index to specify if string keys should be used?

jaffee:
For columns it must be per index. For rows, I think it should be per-frame.

benbjohnson:
@jaffee 👍 That makes sense. Any suggestions for what that flag should be called? Maybe just `Keys` or `UseKeys`?

jaffee:
I like "Keys" I think.

We were talking about this yesterday, and I think the consensus is that in PQL, if you are using keys, the arguments always must be quoted, and if you are providing your own sequential integers then the arguments MUST be unquoted. This will allow for the case of using integers as keys, but mapping them to sequential integers internally.

What I'm wondering now is if we can support both using keys and optionally not using them in the same frame/index. It might be nice for performance purposes if a client could decide to cache the whole mapping locally and use the integer API instead of the Key API for performance purposes. To be clear, I don't think we should pursue this initially as it adds a lot of complexity and corner cases, but I just want to keep it in the back of our heads as we design this API.

benbjohnson:
I added the schema support to `Index` & `Field`, fixed up the rebase, integrated the HTTP async replication, added CLI/config support, and did integration testing. There are still a few issues with some of the call types and how they interact with the bits/rows but several calls like `SetBit()` & `Bitmap()` are working.

Can this PR get merged in so I can do smaller PRs against `develop` from now on? :)

jaffee:
I'll look at it asap.

benbjohnson:
Thanks @jaffee

server.go
```diff
@@ -241,6 +251,14 @@ func NewServer(opts ...ServerOption) (*Server, error) {
 	s.Cluster.Logger = s.logger
 	s.Cluster.Holder = s.Holder
 
+	// Initialize translation database.
+	s.TranslateFile = NewTranslateFile()
+	s.TranslateFile.Path = filepath.Join(path, "keys")
+	s.TranslateFile.PrimaryTranslateStore = s.primaryTranslateStore
+	if err := s.TranslateFile.Open(); err != nil {
```
jaffee: I feel like this belongs in `Server.Open` rather than here in `NewServer`. Probably OK either way though.

statik/statik.go
```diff
@@ -0,0 +1,10 @@
+package statik
```
travisturner: @codysoyland removed `statik` in #1363, so this file shouldn't be here.
I do however see that `github.com/rakyll/statik` is still in `Gopkg.lock`, so we need to get rid of that as well.

server.go
```diff
@@ -241,6 +251,14 @@ func NewServer(opts ...ServerOption) (*Server, error) {
 	s.Cluster.Logger = s.logger
 	s.Cluster.Holder = s.Holder
 
+	// Initialize translation database.
+	s.TranslateFile = NewTranslateFile()
+	s.TranslateFile.Path = filepath.Join(path, "keys")
```
travisturner: This `keys` file in the root of the data directory prevents having an index called `keys`. We could continue with the `.keys` pattern, but making everything a hidden file seems wrong. A better solution may be to have indexes reside in a subdirectory to avoid collisions. Just noting here that we should ticket this to be addressed later.

travisturner:
@benbjohnson this functionality is great!

@jaffee mentioned this in his comment on June 5, but I do think we'll need to address when to consider values to be translated. For example, using this series of queries:
```
/index/k -d '{"options": {"keys": true}}'
/index/k/field/f -d '{"options": {"keys": true}}'
/index/k/query -d 'SetBit(field=f, row="row1", col="col100")'
{"results":[true]}
/index/k/query -d 'SetBit(field=f, row="row1", col="col100")'
{"results":[false]}
/index/k/query -d 'SetBit(field=f, row="row1", col="col101")'
{"results":[true]}
/index/k/query -d 'SetBit(field=f, row="row1", col="col102")'
{"results":[true]}
/index/k/query -d 'Bitmap(field=f, row="row1")'
{"results":[{"attrs":{},"columns":[],"keys":["col100","col101","col102"]}]}
/index/k/query -d 'SetBit(field=f, row="row1", col=200)'
{"results":[true]}
/index/k/query -d 'Bitmap(field=f, row="row1")'
{"results":[{"attrs":{},"columns":[],"keys":["col100","col101","col102",""]}]}
/index/k/query -d 'SetBit(field=f, row="row1", col=201)'
{"results":[true]}
/index/k/query -d 'Bitmap(field=f, row="row1")'
{"results":[{"attrs":{},"columns":[],"keys":["col100","col101","col102","",""]}]}
```
Notice how setting `col` to an integer value results in mapping to a unique, empty string. Technically it's doing the right thing, but that's going to be confusing. I like @jaffee's suggestion that the value type would be required to match the `keys` option. In that case, the example query above:  `SetBit(field=f, row="row1", col=200)` would have resulted in an error.

PR #1365, opened 2018-06-07
WIP count optimization

tgruben:
## Overview
by circumventing the generic container iteration, i'm able to improve
 the speed  TestBitmap_Quick_LargeValue by 4x


roaring/roaring.go
```diff
@@ -93,6 +93,7 @@ type Containers interface {
 	// return the first container at or after key. found will be true if a
 	// container is found at key.
 	Iterator(key uint64) (citer ContainerIterator, found bool)
+	Count() uint64
```
travisturner: At first I was a little confused by the name of this method; "does this count the number of containers?". But I suppose as long as the comment specifies that `Count returns the number of bits set in the containers.` then it's ok.

travisturner: I've been trying to figure out which fields we made available for the b+tree stuff.
It looks like that's in commit 4282e90f where we added a `Mapped()` getter method and an `Update()` method. Would adding a `N()` getter method work for you?

tgruben: Works for me, like it @codysoyland ?

codysoyland: Seems worth the export 👍

PR #1378, opened 2018-06-13
add supporting functions for functional pql.Query construction

tgruben:
## Overview

Provides a functional method for building the parse tree ( pql.Query). This could be useful for testing parser results and experimenting with new pql features without having to bother with the details of the parser.


tgruben:
i was slightly concerned about potential name clashes if placed in pilosa. I'm give internal a go

PR #1381, opened 2018-06-15
split FieldOptions into multiple FieldTypeOptions (Set, Int, Time)

### LINKED ISSUE #1361: Improve validation and representation of FieldOptions
### Description
Based on the work from #1345 where `FieldOptions` were defined as:
```
CREATE field type='default', cacheType='ranked', cacheSize=10000
CREATE field type='int', min=0, max=1023
CREATE field type='time', quantum='YMD'
```
We need to decide how best to validate options on field creation.
For example, should the following raise an error:
```
{"options":{"type":"int", "cacheType": "ranked"}}
```
Field type `int` doesn't take a `cacheType`. Currently, we just ignore this, but perhaps we should raise an error.

Also, when fields are displayed in the `schema`, it can look like this:
```
"options":{"type":"int","cacheType":"none","max":10000}
```
Two issues here: 
1. `cacheType`, even though it's none, doesn't apply to `int` fields
2. `min` is excluded because it is 0

We have defined and implemented how best to validate and display field options.

travisturner:
## Overview

This PR replace the single `FieldOptions` struct with three, field type specific structs: `FieldTypeOptionsSet`, `FieldTypeOptionsInt`, and `FieldTypeOptionsTime`. There's a bit of custom JSON marshaling added to support this. Any feedback on how I handled that would be appreciated.

I'm going to follow this up with some additional tests around the new marshaling and validation logic, but I wanted to get this in review so it's not blocking anyone.

Fixes #1361

## Pull request checklist

## Code review checklist

### DIFF
--- a/api.go
+++ b/api.go
@@ -216,7 +216,7 @@ func (api *API) DeleteIndex(ctx context.Context, indexName string) error {
 }
 
 // CreateField makes the named field in the named index with the given options.
-func (api *API) CreateField(ctx context.Context, indexName string, fieldName string, options FieldOptions) (*Field, error) {
+func (api *API) CreateField(ctx context.Context, indexName string, fieldName string, options FieldTypeOptions) (*Field, error) {
 	if err := api.validate(apiCreateField); err != nil {
 		return nil, errors.Wrap(err, "validating api method")
 	}
@@ -238,7 +238,7 @@ func (api *API) CreateField(ctx context.Context, indexName string, fieldName str
 		&internal.CreateFieldMessage{
 			Index: indexName,
 			Field: fieldName,
-			Meta:  options.Encode(),
+			Meta:  encodeOptions(options),
 		})
 	if err != nil {
 		api.Logger.Printf("problem sending CreateField message: %s", err)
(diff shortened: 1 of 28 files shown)


travisturner:
I agree that this approach results in less than desirable code. With that said, I don't love having multiple methods on the API. As I understand it, that would mean we would have:
```
API.CreateFieldSet(indexName string, fieldName string, options FieldSetOptions)
API.CreateFieldInt(indexName string, fieldName string, options FieldIntOptions)
API.CreateFieldTime(indexName string, fieldName string, options FieldTimeOptions)
```
I'm trying to figure out what `Index.CreateField()` looks like in that case. Would we have multiple `CreateField*` methods on `Index` as well? Or would the API convert the field type-specific options into a general `FieldOptions` object which contains fields for all possible types, even though some aren't used? That would probably be reasonable if `Index.CreateField` wasn't also exported and therefore part of the API surface (as is `NewField()`). In other words, it seems like we would also need:
```
Index.CreateFieldSet()
Index.CreateFieldInt()
Index.CreateFieldTime()
```
I'll look through it some more and see if I can come up with a better option.

jaffee:
I was thinking more like
```go
CreateSetField(index, field string, cacheSize uint)
CreateIntField(index, field string, min, max int64)
CreateTimeField(index, field, quantum string)
```
but that may be premature - I don't think it will evolve well as our understanding of what we want to support with fields improves.

I think decoding into a private fieldoptions struct with pointer fields and then validating and copying to the public struct might be the cleanest.

travisturner:
I'll give that a shot. I'm going to base the new work off of #1337 with the assumption that that will be merged soon.

jaffee:
closed in favor of #1409

PR #1399, opened 2018-06-20
Accept json

### LINKED ISSUE #1383: HTTP handlers need to all check Accept: headers
### Description
Some of our HTTP handlers just return JSON with nary a care in the world about what the requester actually wanted.

In case we want to support multiple protobuf or other encodings for these endpoints in the future, we should explicitly deny requests that specify anything other than JSON.

When asked explicitly for a certain encoding, none of PIlosa's handlers should respond with anything other than the encoding requested, or an error.

tgruben:
## Overview
Adds check for Accept headers to confirm clients accepts application/json responses.  Note this might require a fix to all the clients @yuce .   Just make sure they contain the "Accept:application/json" header for those requests that you are expecting json

Fixes #1383

### DIFF
--- a/http/handler.go
+++ b/http/handler.go
@@ -164,11 +164,12 @@ func (h *Handler) queryArgValidator(next http.Handler) http.Handler {
 func NewRouter(handler *Handler) *mux.Router {
 	router := mux.NewRouter()
 	router.HandleFunc("/", handler.handleHome).Methods("GET")
-	router.HandleFunc("/cluster/message", handler.handlePostClusterMessage).Methods("POST")
-	router.HandleFunc("/cluster/resize/set-coordinator", handler.handlePostClusterResizeSetCoordinator).Methods("POST")
 	router.PathPrefix("/debug/pprof/").Handler(http.DefaultServeMux).Methods("GET")
 	router.Handle("/debug/vars", expvar.Handler()).Methods("GET")
+
 	router.HandleFunc("/schema", handler.handleGetSchema).Methods("GET")
+	router.HandleFunc("/cluster/message", handler.handlePostClusterMessage).Methods("POST")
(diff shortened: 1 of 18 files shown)


test/pilosa_test.go
```diff
@@ -32,12 +33,21 @@ func TestNewCluster(t *testing.T) {
...
+		strings.NewReader(""),
+	)
 
-	response, err := http.Get("http://" + cluster[0].Server.Addr().String() + "/status")
+	req.Header.Set("Accept", "application/json")
+
+	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
-		t.Fatalf("getting schema: %v", err)
+		panic(err)
```
jaffee: why panic?

test/pilosa.go
```diff
@@ -263,10 +263,15 @@ func (m *Main) RecalculateCaches() error {
...
-	if err != nil {
-		panic(err)
-	}
+	req, err := gohttp.NewRequest(
+		method,
+		urlStr,
+		strings.NewReader(body),
+	)
+
+	req.Header.Set("Content-Type", "application/json")
```
jaffee: is MustDo always used for json requests? maybe we should not set headers in here at all.

tgruben: you don't have access to headers outside of the `MustDo` function.  its not that big a deal now that the Accept is not required.  I can remove if you like, i don't see any value either way

http/handler.go
```diff
@@ -257,8 +258,31 @@ func (h *Handler) handleHome(w http.ResponseWriter, r *http.Request) {
...
+	sendError := false
+	if found {
+		sendError = true
+		for _, v := range v {
+			if v == "application/json" {
+				sendError = false
+
+			}
+		}
+
```
jaffee: some needless xtra whitespace in this func

test/pilosa.go
jaffee: all good

PR #1407, opened 2018-06-22
Wip api refactor

jaffee:
## Overview

This pull request simplifies the dependency/reference diagram of various Pilosa components.

The pilosa.Server object no longer has any reference to Handler or listener. The order of dependence is as follows.

Handler (containing a net.Listener) -> API -> Server -> Holder,Cluster,etc.

By themselves, these changes are fairly straightforward, but because our test logic and helpers often abused the cyclical reference behavior, much refactoring was needed in the tests.

There are also some unrelated changes which got rolled up into this PR as we continually merged develop back into it. I will do my best to annotate these in the PR, so it's less confusing.

@codysoyland and I tag-teamed this effort, so another reviewer would probably be good.

Relates to several tickets.

## Pull request checklist

## Code review checklist


cluster.go
```diff
@@ -914,7 +914,7 @@ func (c *Cluster) open() error {
 			return fmt.Errorf("sending restart NodeJoin: %v", err)
 		}
 
-		c.Logger.Printf("wait for joining to complete")
+		c.Logger.Printf("%v wait for joining to complete", c.Node.ID)
```
jaffee: was trying to debug a deadlock... I think this log is slightly more helpful though

executor.go
```diff
@@ -1618,6 +1618,9 @@ func (e *Executor) translateCall(index string, idx *Index, c *pql.Call) error {
 	// Translate row key, if field is specified & key exists.
 	if fieldName != "" {
 		field := idx.Field(fieldName)
+		if field == nil {
```
jaffee: this and the following block fix a bug introduced by the newpql work where the field could be nil.

executor_test.go
```diff
@@ -1042,6 +1042,8 @@ func TestExecutor_Execute_BSIGroupRange(t *testing.T) {
 
 // Ensure a remote query can return a row.
 func TestExecutor_Execute_Remote_Row(t *testing.T) {
+	t.Skip() // Until test.NewServer() works
```
jaffee: most of the tests were fixed, but several that had more complicated/invasive behavior are still skipped. We thought the current changes were impactful enough that it would be better to get them merged before improving the organization and test tooling to the point where we could rewrite these tests.

fragment_internal_test.go
```diff
@@ -213,39 +213,6 @@ func TestFragment_SetValue(t *testing.T) {
 			t.Fatal(err)
 		}
 	})
-	t.Run("Crash", func(t *testing.T) {
```
jaffee: todd added this test when trying to fix that crash bug - it didn't turn out to be useful, so I've removed it here.

server.go
```diff
@@ -400,12 +364,21 @@ func (s *Server) LoadNodeID() string {
...
+	return "tcp"
+}
+
 // Addr returns the address of the listener.
 func (s *Server) Addr() net.Addr {
-	if s.ln == nil {
-		return nil
-	}
-	return s.ln.Addr()
+	return pilosaAddr(s.URI)
```
jaffee: little hack so that we can still return a net.Addr even though Server no longer has a net.Listener

test/handler.go
```diff
@@ -84,22 +88,23 @@ type Server struct {
...
-
-	// Handler test messages can no-op.
-	s.Handler.API.Broadcaster = pilosa.NopBroadcaster
-	// Create a default cluster on the handler
-	s.Handler.API.Cluster = NewCluster(1)
-	s.Handler.API.Cluster.Nodes[0].URI = s.HostURI()
-
-	return s
+	return &Server{}
+	//handler, err := pilosa.NewHandler()
```
jaffee: test.NewServer will probably be deleted, but we'll keep the commented implementation here in case we need to reason about it while we fix the last few tests that use it.

codysoyland:
Note to reviewers that a0337714120769cdd520962c119db4bca9931a91 moved a lot of tests from `http` to `server`, so it's a lot easier to review some of those tests by looking at individual commits.

test/pilosa.go
```diff
@@ -221,6 +229,13 @@ func (m *Main) RunWithTransport(host string, bindPort int, joinSeeds []string) (
 
 	m.Server.Cluster.Static = false
 
+	go func() {
+		err := m.Handler.Serve()
+		if err != nil {
+			log.Printf("Handler serve error: %v", err)
```
codysoyland: We should exit the process if this error happens. Since it happens in a goroutine, we will probably want to add a channel to the Command struct, send to it in the error, and listen for it in the `select{}` in `Wait()`. We can do that after merging if you want.

test/handler.go
```diff
@@ -45,7 +45,11 @@ func NewHandler(opts ...http.HandlerOption) (*Handler, error) {
 	h := &Handler{
 		Handler: handler,
 	}
-	h.API = pilosa.NewAPI()
+
+	//h.API, err = pilosa.NewAPI(OptAPIServer(s))
```
travisturner: address this commented code

api.go
```diff
@@ -46,16 +46,43 @@ type API struct {
 	Cluster          *Cluster
 	TranslateStore   TranslateStore
 	Logger           Logger
+	server           *Server
+}
+
+// APIOption is a functional option type for pilosa.API
+type APIOption func(s *API) error
```
travisturner: very picky, but odd that this is `s`

ctl/import_test.go
```diff
@@ -95,18 +86,12 @@ func TestImportCommand_RunValue(t *testing.T) {
...
-	defer hldr.Close()
-	s := test.NewServer()
-	defer s.Close()
-
-	s.Handler.API.Cluster = test.NewCluster(1)
-	s.Handler.API.Cluster.Nodes[0].URI = s.HostURI()
-	s.Handler.API.Holder = hldr.Holder
-	cm.Host = s.Host()
+	cmd := test.MustRunMainWithCluster(t, 1)[0]
+	hostport := cmd.Server.URI.HostPort()
```
travisturner: this pattern of `hostport` is different from the pattern later in this test where you set `cm.Host` directly and then use that going forward.

executor_test.go
travisturner: just a note to make sure we address these `Skip()`s

PR #1409, opened 2018-06-25
prepare to un-export pilosa.FieldOptions

### LINKED ISSUE #1361: Improve validation and representation of FieldOptions

travisturner:
## Overview

This PR is meant to get feedback on this particular approach to field options.
The initial commit  moves `FieldOption` functional options to the `api.CreateField` method, and moves `fieldOptions` unmarshaling to the http handler since `http` is the only package receiving field options as JSON. It also provides custom JSON marshaling (in the `pilosa` package) for schema output as JSON.

Assuming this approach is reasonable, I will address the various TODOs, including field option validation. We also need to determine if field options should be supported (as flags?) in the `pilosa import` subcommand.

Fixes #1361


codysoyland:
Seems like a good direction to me 👍

http/handler_internal_test.go
```diff
@@ -59,31 +60,121 @@ func TestPostIndexRequestUnmarshalJSON(t *testing.T) {
...
+	return &s
+}
+
+func int64Ptr(i int64) *int64 {
+	return &i
+}
+
+// Test fieldOption validation.
+func TestFieldOptionValidation(t *testing.T) {
+	//foo := "foo"
```
travisturner: remove these comments

api.go
```diff
@@ -244,19 +244,27 @@ func (api *API) DeleteIndex(ctx context.Context, indexName string) error {
 }
 
 // CreateField makes the named field in the named index with the given options.
-func (api *API) CreateField(ctx context.Context, indexName string, fieldName string, options FieldOptions) (*Field, error) {
+func (api *API) CreateField(ctx context.Context, indexName string, fieldName string, opts ...FieldOption) (*Field, error) {
```
jaffee: I think we should allow only a single option for now... we can the "..." back later if there we add more options that can be added together.

PR #1424, opened 2018-06-27
Clearbit for time fields

### LINKED ISSUE #1328: remove timestamp argument from clearbit
### Description

Sending a timestamp to clearbit doesn't really make sense because it's tough to know when you can remove the higher order aggregated bits (like years), since multiple timestamps can be set on a  bit. Clearbit should always remove everything and not have a timestamp argument.

Make sure the timestamp argument is removed from clearbit in code and docs, and that clearbit behaves appropriately.

tgruben:
## Overview

removes a column from all time relevant time fields.

Fixes #1328


field.go
```diff
@@ -708,29 +712,78 @@ func (f *Field) ClearBit(rowID, colID uint64, t *time.Time) (changed bool, err e
...
+	if len(f.views) == 1 { // assuming no time views
 		return changed, nil
 	}
-
-	// If a timestamp is specified then clear bits across all views for the quantum.
-	for _, subname := range viewsByTime(viewName, *t, f.TimeQuantum()) {
-		view, err := f.CreateViewIfNotExists(subname)
-		if err != nil {
-			return changed, errors.Wrapf(err, "creating view %s", subname)
+	lastLevel := 0
```
travisturner: the names of these three variables made this confusing to read.
some comments would probably help, but also renaming; particularly the two "level" variables since they don't actually mean the same thing.
- `lastLevel` represents a `len(view.name)`: `standard_2018=6, standard_201801=8, standard_20180102=10, standard_2018010203=12`
- `level` represents the date quantum level: `Y=1, YM=2, YMD=3, YMDH=4` or `Y=1, YD=2` or `M=1, MD=2`, etc. 
- `skipBelow` really means "skip all levels above this value"

I would suggest renaming `lastLevel` to `lastViewNameSize` or something like that, and renaming `skipBelow` to `skipAbove` or `skipLevelsAbove` or something a little more clear. I realize if you envision a tree of time quantums with Y at the top, then levels are below, but the logic here has levels increasing numerically.

field.go
```diff
@@ -708,29 +712,78 @@ func (f *Field) ClearBit(rowID, colID uint64, t *time.Time) (changed bool, err e
...
+			i++
+		}
+	}
+	me = me[:i]
+	year := strings.Index(me[0].name, "_") + 4
+	month := year + 2
+	day := month + 2
+	sort.Slice(me, func(i, j int) (lt bool) {
+		var eq bool
+		//	ensure all catA are grouped together:
```
travisturner: what is `catA`?

internal/public.pb.go
```diff
@@ -27,7 +27,7 @@ import proto "github.com/golang/protobuf/proto"
 import fmt "fmt"
 import math "math"
 
-import encoding_binary "encoding/binary"
+import binary "encoding/binary"
```
travisturner: why did this file change?

view.go
```diff
@@ -57,9 +57,9 @@ type View struct {
 	// prevent sending multiple `CreateSliceMessage` messages
 	maxSlice uint64
 
-	broadcaster Broadcaster
-	stats       StatsClient
-
+	broadcaster  Broadcaster
+	stats        StatsClient
+	viewType     string
```
travisturner: this doesn't appear to be used

field.go
```diff
@@ -599,7 +602,7 @@ func (f *Field) createViewIfNotExistsBase(name string) (*View, bool, error) {
 	return view, true, nil
 }
 
-func (f *Field) newView(path, name string) *View {
+func (f *Field) newView(path string, name string) *View {
```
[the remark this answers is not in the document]
tgruben: an artifact of removal of my "viewkey" object.  I originally was using a thing to represent the type of a view, but in the end i just performed some logic on the name to take the easy way out.

field.go
```diff
@@ -708,29 +712,78 @@ func (f *Field) ClearBit(rowID, colID uint64, t *time.Time) (changed bool, err e
 	} else if v {
 		changed = v
 	}
-
-	// Exit early if no timestamp is specified.
-	if t == nil {
+	if len(f.views) == 1 { // assuming no time views
```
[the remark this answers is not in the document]
tgruben: I think a better ticket would be to add some method of identifying view type.  That was my struggle with this pr.  On version of this code stored the type on view creation, but that would be a breaking change so i backed out of that route.  I feel this current method is pretty much a hack, but i 'm not sure if finding a better solution warrants the effort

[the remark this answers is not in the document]
travisturner: let's leave this as is for now. we can decide later whether to handle the `standard` view for time fields differently.

PR #1433, opened 2018-06-29
consolidate http errors into a shared response type

### LINKED ISSUE #1280: Proposal: make HTTP responses more clear and consistent
### Description
Some of our responses could stand to be updated.

Successfully creating an index or a frame returns `{}`, which is strange, and when the entity already exists, the response is a non-json string. One alternative is to change this to match the response from SetBit/ClearBit, summarized in this table:

| request | result | current response | proposed response |
| ------------- | ------------- |------------- |------------- |
| create index | create | `{}` | `{"results":[true]}` |
| create index | already exists | `creating index: index already exists` | `{"results":[false]}` |
| create frame | create | `{}` | `{"results":[true]}` |
| create frame | already exists | `creating frame: frame already exists` | `{"results":[false]}` |

More sensibly: 

| request | result | current response | proposed response |
| ------------- | ------------- |------------- |------------- |
| create index | create | `{}` | `{"success":true}` |
| create index | already exists | `creating index: index already exists` | `{"success":false, "error":"creating index: index already exists"}` |
| create frame | create | `{}` | `{"success":true}` |
| create frame | already exists | `creating frame: frame already exists` | `{"success":false, "error":"creating frame: frame already exists"}` |
| delete index | delete | `{}` | `{"success":true}` |
| delete index | not found | `{}` | `{"success":false, "error": "index not found}` |
| get index schema | not found | `index not found` | `{"success":false, "error": "index not found"}` |
| delete frame | delete | `{}` | `{"success":true}` |
| delete frame | not found | `{}` | `{"success":false, "error": "frame not found}` |
| create field | create | `{}` | `{"success":true}` |
| create field | already exists | `creating field: field already exists` | `{"success":false, "error": "creating field: field already exists"}` |
| id | get | `40a891fa-243b-4d71-ae24-4f5c78a0f4b1` | `{"id": `40a891fa-243b-4d71-ae24-4f5c78a0f4b1"}` |

Of course, some of these responses might be more convenient as plain text for whatever reason. If that's the case, we can make exceptions.

Perhaps a simpler option is to replace all `{}` responses with empty responses, and ensure that the HTTP response codes are appropriate. For example, POST to `/index/indexname` returns `200 OK`, but might return `201 CREATED` instead.

HTTP responses are as clear and consistent as possible.

travisturner:
## Overview

This PR adds a custom `Error` to the `http` package which is used to return success true/false json responses.

TODO:
- [x] Modify API to return error when deleting a non-existent index or field
- [x] Address the other http endpoints (like `/id`).

Fixes #1280

### DIFF
--- a/holder.go
+++ b/holder.go
@@ -304,7 +304,7 @@ func (h *Holder) CreateIndex(name string, opt IndexOptions) (*Index, error) {
 
 	// Ensure index doesn't already exist.
 	if h.indexes[name] != nil {
-		return nil, ConflictError{ErrIndexExists}
+		return nil, NewConflictError(ErrIndexExists)
 	}
 	return h.createIndex(name, opt)
 }
@@ -374,7 +374,7 @@ func (h *Holder) DeleteIndex(name string) error {
 	// Confirm index exists.
 	index := h.index(name)
 	if index == nil {
-		return NotFoundError{ErrIndexNotFound}
+		return NewNotFoundError(ErrIndexNotFound)
 	}
 
 	// Close index.
--- a/holder.go
+++ b/holder.go
@@ -304,7 +304,7 @@ func (h *Holder) CreateIndex(name string, opt IndexOptions) (*Index, error) {
 
 	// Ensure index doesn't already exist.
 	if h.indexes[name] != nil {
-		return nil, ErrIndexExists
+		return nil, ConflictError{ErrIndexExists}
 	}
 	return h.createIndex(name, opt)
 }
--- a/holder.go
+++ b/holder.go
@@ -371,10 +371,10 @@ func (h *Holder) DeleteIndex(name string) error {
 	h.mu.Lock()
 	defer h.mu.Unlock()
 
-	// Ignore if index doesn't exist.
+	// Confirm index exists.
 	index := h.index(name)
 	if index == nil {
-		return nil
+		return NotFoundError{ErrIndexNotFound}
 	}
 
 	// Close index.
--- a/http/handler.go
+++ b/http/handler.go
@@ -578,19 +578,14 @@ func (h *Handler) handlePostIndex(w http.ResponseWriter, r *http.Request) {
 
 	resp := successResponse{}
 
-	err := func() error {
-		// Decode request.
-		var req postIndexRequest
-		err := json.NewDecoder(r.Body).Decode(&req)
-		if err == io.EOF {
-			// If no data was provided (EOF), we still create the index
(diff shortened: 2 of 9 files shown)


travisturner:
@jaffee @alanbernstein I'm taking this out of WIP. I believe I've addressed everything in the proposal that hasn't been deprecated.

http/handler.go
```diff
@@ -530,30 +576,23 @@ func (h *Handler) handlePostIndex(w http.ResponseWriter, r *http.Request) {
...
-	if errors.Cause(err) == pilosa.ErrIndexExists {
-		http.Error(w, err.Error(), http.StatusConflict)
-		return
-	} else if err != nil {
-		http.Error(w, err.Error(), http.StatusInternalServerError)
-		return
-	}
+	resp := successResponse{}
+
+	err := func() error {
```
jaffee: I don't think this closure is improving readability or brevity... consider removing.

http/handler.go
```diff
@@ -606,60 +645,51 @@ func (h *Handler) handlePostField(w http.ResponseWriter, r *http.Request) {
...
+		// Decode request.
+		var req postFieldRequest
+		dec := json.NewDecoder(r.Body)
+		dec.DisallowUnknownFields()
+		err := dec.Decode(&req)
+		if err == io.EOF {
+			// If no data was provided (EOF), we still create the field
+			// with default values.
+		} else if err != nil {
```
jaffee: combine the if and the else as `if err != nil && err != io.EOF`

http/handler.go
```diff
@@ -606,60 +645,51 @@ func (h *Handler) handlePostField(w http.ResponseWriter, r *http.Request) {
...
+		if err != nil {
+			return err
 		}
-		return
-	}
-	// Encode response.
-	if err := json.NewEncoder(w).Encode(postFieldResponse{}); err != nil {
-		h.Logger.Printf("response encoding error: %s", err)
-	}
+		return nil
```
jaffee: skip the if, and just return err

holder.go
```diff
@@ -304,7 +304,7 @@ func (h *Holder) CreateIndex(name string, opt IndexOptions) (*Index, error) {
 
 	// Ensure index doesn't already exist.
 	if h.indexes[name] != nil {
-		return nil, ErrIndexExists
+		return nil, ConflictError{ErrIndexExists}
```
jaffee: seems like there are still a few of these (unkeyed fields)

Issue #1371, opened 2018-06-11
http endpoints cleanup


jaffee:
### Description

We have some silly endpoints like `/slices/max` and `/info`. What does info return? Well that's kind of my point - you have no idea. What other endpoints are under `/slices`? None. Just `/slices/max`. Turns out the only info under `/info` is slicewidth. Let's roll all this information under the `/status` endpoint. This has the added benefit of putting all the information that a client library wants to know in one place!

We have a number of endpoints that are for internal use only - let's make this explicit and nest them under `/internal` then maybe people won't get angry if we change them post-1.0. I think the list is:
- `/cluster/message`
- `/fragment/*`
- `/*/attr/diff`

Imports are frame specific - why are the `/import` and `/import-value` endpoints top-level? Let's put them under the `/index/{index}/field/{field}`. Bonus - then we won't need a separate import-value because it will be clear from the field type.

This also seems like a good idea: #1283

### Success criteria (What criteria will consider this ticket closeable?)

We do all that stuff ☝️

travisturner:
Regarding `/info`, when we discussed this the idea was to include compile-time information at a single endpoint (sliceWidth, version). See #1236.
I'm not opposed to munging everything together into a single endpoint as long as we're thoughtful about it. We just got out of a situation where `/status` returned many different things, each selectively used by different functions, so I'd like to avoid getting in that place again.

alanbernstein:
Related: https://github.com/pilosa/pilosa/issues/1280

PR #1441, opened 2018-07-02
move internal http endpoints under /internal

travisturner:
## Overview

This PR addresses the `internal` part of #1371.
The portion pertaining to import consolidation is a bit more involved, so I'm treating that separately.

### DIFF
--- a/docs/administration.md
+++ b/docs/administration.md
@@ -232,7 +232,7 @@ Note: This will only work when the replication factor is >= 2
   - List of all indexes on your cluster
   - List of all frames in your indexes
   - Max slice per index, listed in the `/slices/max` endpoint
-- With this information you can query the `/fragment/nodes` endpoint and iterate over each slice
+- With this information you can query the `/internal/fragment/nodes` endpoint and iterate over each slice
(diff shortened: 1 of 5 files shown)


docs/administration.md
```diff
@@ -232,7 +232,7 @@ Note: This will only work when the replication factor is >= 2
   - List of all indexes on your cluster
   - List of all frames in your indexes
   - Max slice per index, listed in the `/slices/max` endpoint
```
yuce: Max *shard* per index, listed in the `/internal/shards/max` endpoint

docs/administration.md
```diff
@@ -232,7 +232,7 @@ Note: This will only work when the replication factor is >= 2
...
-- With this information you can query the `/fragment/nodes` endpoint and iterate over each slice
+- With this information you can query the `/internal/fragment/nodes` endpoint and iterate over each slice
```
yuce: iterate over each *shard*

travisturner: @yuce I didn't make all the necessary changes to `docs`. We're going to address that all at once.

PR #1459, opened 2018-07-05
ensure /schema excludes views and includes all field options (like "keys")

### LINKED ISSUE #1349: remove views from API responses
### Description

Several API endpoints have view names in their responses. (like /schema, /slices/max). Since we're removing views from exposure in the API, it is probably confusing to refer to them in API responses, so I think we should modify the responses to not include views.

API responses don't reference views either explicitly or implicitly.

travisturner:
## Overview

This PR ensures that `views` are excluded from external calls to `/schema`.
It also adds `API.Indexes()` for use by the `gossip` package.
Finally, it adds the `keys` field option in schema output.

Fixes #1349

### DIFF
--- a/api.go
+++ b/api.go
@@ -477,8 +477,14 @@ func (api *API) ClusterMessage(ctx context.Context, reqBody io.Reader) error {
 }
 
 // Schema returns information about each index in Pilosa including which fields
+// they contain.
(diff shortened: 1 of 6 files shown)


travisturner:
closing in favor of #1466

PR #1473, opened 2018-07-06
Fix test

travisturner:
## Overview

This test was failing with a different error depending on the order of the `call.Args` map. I removed the error being raised in `translateCall()` and assume that it will be handled appropriately downstream.

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -1579,7 +1579,7 @@ func (e *executor) translateCall(index string, idx *Index, c *pql.Call) error {
 		if field == nil {
 			// Instead of returning ErrFieldNotFound here,
 			// we just return, and don't attempt the translation.
-			// The assumption is that the non-existant field
+			// The assumption is that the non-existent field
 			// will raise an error downstream when it's used.
 			return nil
 		}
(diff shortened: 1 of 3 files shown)


executor.go
```diff
@@ -1575,7 +1577,11 @@ func (e *executor) translateCall(index string, idx *Index, c *pql.Call) error {
 	if fieldName != "" {
 		field := idx.Field(fieldName)
 		if field == nil {
-			return ErrFieldNotFound
+			// Instead of returning ErrFieldNotFound here,
+			// we just return, and don't attempt the translation.
+			// The assumption is that the non-existant field
```
alanbernstein: "non-existent"

alanbernstein:
Incidentally, if I run this with `-count=1000`, I get `too many open files in system` (my `ulimit -n` is 7168, so not surprising).

If I run with `-count=500`, it passes. `-count=600`, I get this:

```
    --- FAIL: TestExecutor_Execute_SetValue/#00 (0.00s)
    	pilosa.go:226: run cluster: starting cluster: starting server 0: setting up server: getting listener: net.Listen: listen tcp: lookup localhost: no such host
```

None of this is a problem for this change, but I wonder if there is any cleanup that should be happening and isn't. If so it might be worth looking into for the sake of CI jobs.

jaffee:
That could be gossip sice its TCP and UDP

PR #1540, opened 2018-07-19
Fix linter issues: vet

### LINKED ISSUE #1521: Metalinter: vet
```
gossip/gossip.go:308::error: github.com/pilosa/pilosa/vendor/github.com/hashicorp/memberlist.NodeEvent composite literal uses unkeyed fields (vet)
gossip/gossip.go:312::error: github.com/pilosa/pilosa/vendor/github.com/hashicorp/memberlist.NodeEvent composite literal uses unkeyed fields (vet)
gossip/gossip.go:316::error: github.com/pilosa/pilosa/vendor/github.com/hashicorp/memberlist.NodeEvent composite literal uses unkeyed fields (vet)
http/error.go:20::error: struct field code has json tag but is not exported (vet)
ctl/import.go:62::error: struct field client has json tag but is not exported (vet)
server/server_test.go:280::error: unreachable code (vet)
index.go:409::error: struct field options has json tag but is not exported (vet)
```

codysoyland:
Fixes #1521

### DIFF
--- a/index.go
+++ b/index.go
@@ -405,9 +405,8 @@ func (p indexSlice) Less(i, j int) bool { return p[i].Name() < p[j].Name() }
 
 // IndexInfo represents schema information for an index.
 type IndexInfo struct {
-	Name    string       `json:"name"`
-	options IndexOptions `json:"options"`
-	Fields  []*FieldInfo `json:"fields"`
+	Name   string       `json:"name"`
+	Fields []*FieldInfo `json:"fields"`
 }
 
 type indexInfoSlice []*IndexInfo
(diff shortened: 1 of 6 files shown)


index.go
```diff
@@ -405,9 +405,8 @@ func (p indexSlice) Less(i, j int) bool { return p[i].Name() < p[j].Name() }
 
 // IndexInfo represents schema information for an index.
 type IndexInfo struct {
-	Name    string       `json:"name"`
-	options IndexOptions `json:"options"`
```
jaffee: this is kind of interesting... perhaps this was unexported by mistake? seems like we might want to return options with index info

codysoyland: It's actually not referred to by anything, hence the removal.

jaffee: right, but it would be useful in an HTTP response serialized to json. I think this is the struct that gets returned if you make /index/name GET request.

codysoyland: I made a separate PR to add that feature.

PR #1542, opened 2018-07-19
Add IndexOptions to IndexInfo json response

codysoyland:
This adds IndexOptions to responses that serialize IndexInfo, including /schema, /index, and /index/{index}.

This came up during discussion on #1540.

### DIFF
--- a/http/handler_internal_test.go
+++ b/http/handler_internal_test.go
@@ -31,6 +31,7 @@ func TestPostIndexRequestUnmarshalJSON(t *testing.T) {
 		err      string
 	}{
 		{json: `{"options": {}}`, expected: postIndexRequest{Options: pilosa.IndexOptions{}}},
+		{json: `{"options": {"keys": true}}`, expected: postIndexRequest{Options: pilosa.IndexOptions{true}}},
 		{json: `{"options": 4}`, err: "options is not map[string]interface{}"},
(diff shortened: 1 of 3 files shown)


index.go
```diff
@@ -95,7 +95,7 @@ func (i *Index) Options() IndexOptions {
 }
 
 func (i *Index) options() IndexOptions {
-	return IndexOptions{}
+	return IndexOptions{i.keys}
```
jaffee: should be `IndexOptions{Keys: i.keys}` no?

http/handler_internal_test.go
```diff
@@ -31,6 +31,7 @@ func TestPostIndexRequestUnmarshalJSON(t *testing.T) {
 		err      string
 	}{
 		{json: `{"options": {}}`, expected: postIndexRequest{Options: pilosa.IndexOptions{}}},
+		{json: `{"options": {"keys": true}}`, expected: postIndexRequest{Options: pilosa.IndexOptions{true}}},
```
jaffee: also here, should use `Keys:`

index.go
codysoyland: Sure, no big reason not to use the field labels. I think the linter only complains when they are cross-package, but I agree they are good to use. Pushed another commit that adds them.

jaffee: ah, I thought it complained all the time - anywho

PR #1501, opened 2018-07-13
Support keys on import CLI.

benbjohnson:
## Overview

This pull request adds support for specifying string keys with `pilosa import`. There was an existing partial implementation but it did not work and required both the index and field to use string keys.

Keys are translated server-side so performance will be limited by the translation file speed. Previous tests show the speed to be quite high though.

## Usage

Given an index `i`, a field `f`, and the following `data.csv` file: 

```
FOO,BAR
FOO,BAZ
```

The import can be run using the normal import command:

```sh
$ pilosa import -i i -f f data.csv
```

---

Fixes #1488

## Pull request checklist

- [x] I have read the [contributing guide](https://github.com/pilosa/pilosa/blob/master/CONTRIBUTING.md).
- [x] I have agreed to the [Contributor License Agreement](https://cla-assistant.io/pilosa/pilosa).
- [x] I have updated the [documentation](https://github.com/pilosa/pilosa/tree/master/docs).
- [x] I have resolved any merge conflicts.
- [x] I have included tests that cover my changes.
- [x] All new and existing tests pass.

## Code review checklist
This is the checklist that the reviewer will follow while reviewing your pull request. You do not need to do anything with this checklist, but be aware of what the reviewer will be looking for.

- [ ] Ensure that any changes to external docs have been included in this pull request.
- [ ] If the changes require that minor/major versions need to be updated, tag the PR appropriately.
- [ ] Ensure the new code is [properly commented](https://github.com/golang/go/wiki/CodeReviewComments#doc-comments) and follows [Idiomatic Go](https://dmitri.shuralyov.com/idiomatic-go).
- [ ] Check that tests have been written and that they cover the new functionality.
- [ ] Run tests and ensure they pass.
- [ ] Build and run the code, performing any applicable integration testing.

PR #1557, opened 2018-07-31
Support keys on import CLI -> master

travisturner:
## Overview

I just realized that #1501 was merged into the `develop` branch. I cherry-picked that commit and fixed the merge conflicts. This PR is the merge to master.

## Pull request checklist

## Code review checklist

### DIFF
--- a/api.go
+++ b/api.go
@@ -609,9 +609,34 @@ func (api *API) Import(_ context.Context, req *ImportRequest) error {
 		return errors.Wrap(err, "validating api method")
 	}
 
+	index := api.holder.Index(req.Index)
+	if index == nil {
+		return newNotFoundError(ErrIndexNotFound)
+	}
+
 	field, err := api.indexField(req.Index, req.Field, req.Shard)
 	if err != nil {
 		return errors.Wrap(err, "getting field")
+	}
+
+	// Translate row keys.
+	if field.keys() {
+		if len(req.RowIDs) != 0 {
+			return errors.New("row ids cannot be used because field uses string keys")
+		}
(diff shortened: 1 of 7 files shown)


travisturner:
There's still an issue with this, so I'm going to hold of on merging this.
The logic here works as long as shard 0 is on the same node as the primary translate store (because all key imports—column key imports, at least—are being sent to shard 0. If, however, shard 0 is located on a secondary translate node, then the `translate store read only` error is returned. We need to address this issue specifically, and also address how best to handle writes to non-primary nodes generally.

benbjohnson:
@travisturner I think we’ll need to have the primary be the coordinator for imports. We could have replicas RPC to the primary for unknown keys too. That might be slow when importing a lot of new keys though. What do you think?

travisturner:
@benbjohnson when you say "be the coordinator" are you referring specifically to `cluster.Coordinator`? I was just looking into the viability of that. It would be nice if we could have the Coordinator node be the primary translate store, and all non-Coordinator nodes stream replica data from the Coordinator. In that case we could get rid of the `translation.primary-url` configuration option.

We would still need to modify the Import logic in the client, but instead of sending columnKey imports to shard 0, we could always send them to the Coordinator.

We would also need to address what happens when the coordinator is changed (i.e. we'd need to modify each node's primary translate store reference).

And finally, there is the case where an import contains columnIDs and rowKeys. I think there, too, we would send those through the Coordinator (for translation).

travisturner:
Yeah, that makes sense that having a single, shared primary doesn't work for replication scaling. What do you think about having each node treat the node before it in the cluster as its primary (with the exception of the primary/coordinator, of course)?

travisturner:
I created issue #1578 to discuss/address this.

PR #1569, opened 2018-08-01
Backport #1568

codysoyland:
Backport #1568 from master to v1.0


codysoyland:
These sleeps don't seem to affect the build. It appears that at least one of these tests is flawed in a way that we don't understand. Closing this for now.

PR #1600, opened 2018-08-15
Maintain available shards set

benbjohnson:
## Overview

This pull request removes the previous `MaxShard` tracking and replaces it with an `Available Shards` set tracking. This allows sparse shard tracking without implicitly tracking all shards in between.

## Pull request checklist

## Code review checklist


travisturner:
This is great, and also addresses an issue from #591:
```
curl -X POST http://localhost:10101/index/i
curl -X POST http://localhost:10101/index/i/field/f
curl -X POST http://localhost:10101/index/i/query -d 'Set(6018148517796380732, f=0)'

curl -X POST http://localhost:10101/index/i/query -d 'Row(f=0)'
{"results":[{"attrs":{},"columns":[6018148517796380732]}]}
```

encoding/proto/proto.go
```diff
@@ -768,12 +801,35 @@ func decodeNodeEventMessage(pb *internal.NodeEventMessage, m *pilosa.NodeEvent)
...
-	decodeNode(pb.Node, m.Node)
-	m.MaxShards = pb.MaxShards.Standard
+	decodeIndexStatuses(pb.Indexes, m.Indexes)
 	m.Schema = &pilosa.Schema{}
 	decodeSchema(pb.Schema, m.Schema)
 }
 
+func decodeIndexStatuses(a []*internal.IndexStatus, m []*pilosa.IndexStatus) {
+	for i := range a {
+		m[i] = &pilosa.IndexStatus{}
```
travisturner: There's an `index out of range` issue here when `a` contains indexes (and `m` does not).
It seems like `m` is always an empty slice, so maybe we just need to set `m` to contain `len(a)` of its zero-value before the for loop. Or do an append in the for loop instead.

You can see the error by starting up a two node cluster, add an index, then restart the cluster; the error will occur after the restart on the second node where `len(a) = 1` and `len(m) = 0`.

benbjohnson: Good catch. I was hitting a gnarly unit test bug that was caused by this but covered up by a `recover()`. Fixed in f4c9c0f.

server.go
```diff
@@ -604,19 +604,18 @@ func (s *Server) mergeRemoteStatus(ns *NodeStatus) error {
...
-		}
-		if newMax > oldmaxshards[index] {
-			oldmaxshards[index] = newMax
-			localIndex.setRemoteMaxShard(newMax)
+	// Sync available shards.
+	for _, is := range ns.Indexes {
+		for _, fs := range is.Fields {
+			f := s.holder.Field(is.Name, fs.Name)
+
+			// if we don't know about an field locally, log an error because
```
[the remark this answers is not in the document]
benbjohnson: Fixed in f4c9c0f.

benbjohnson:
@travisturner Fixed up this PR & broken integration tests in f4c9c0f. Ready for review again.

PR #1619, opened 2018-09-04
removing bounds check

tgruben:
## Overview

Picked up some new goodness at gophercon 2018. so I added  some hints in order to remove bounds check (BCE)

`BenchmarkBitmap_Union-8                             3748          2890          -22.89%
BenchmarkBitmap_Intersect-8                         3752          2859          -23.80%
BenchmarkBitmap_Difference-8                        3853          2964          -23.07%`

### DIFF
--- a/roaring/roaring.go
+++ b/roaring/roaring.go
@@ -2136,15 +2136,26 @@ func intersectArrayBitmap(a, b *Container) *Container {
 }
 
 func intersectBitmapBitmap(a, b *Container) *Container {
-	output := &Container{bitmap: make([]uint64, bitmapN), containerType: containerBitmap}
+	var (
 
-	for i := range a.bitmap {
-		v := a.bitmap[i] & b.bitmap[i]
-		output.bitmap[i] = v
-		output.n += int(popcount(v))
+	ab = a.bitmap[:bitmapN]
+	bb = b.bitmap[:bitmapN]
+	buf = make([]uint64, bitmapN)
+	ob =buf[:bitmapN]
 
+	n int
+)
+	for i := 0; i < bitmapN; i++ {
+		v := ab[i] & bb[i]
+		ob[i] = v
+		n += int(popcount(v))
 	}
-	output.optimize()
+
+	output := &Container{
+		bitmap: ob,
+		n:      n,
+		containerType: containerBitmap,
+	}
 	return output
 }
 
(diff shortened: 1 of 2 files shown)


alanbernstein:
It might be nice to add some brief comments to explain why the hints are there, or link to a post explaining it, etc.

roaring/roaring.go
```diff
@@ -3333,9 +3375,16 @@ func popcount(x uint64) uint64 {
 }
 
 func popcountAndSlice(s, m []uint64) uint64 {
```
jaffee: let's add a comment that this *needs* to be called with slices which are exactly `bitmapN` long. I also generally agree with @alanbernstein that we need to call out the bounds checking code as such - it's likely that we'll be able to remove some or all of it as the compiler improves, so we want to be clear on why it's there in the first place.

roaring/roaring.go
```diff
@@ -2871,16 +2903,26 @@ func xorArrayBitmap(a, b *Container) *Container {
...
 func xorBitmapBitmap(a, b *Container) *Container {
-	output := &Container{
-		bitmap:        make([]uint64, bitmapN),
-		containerType: containerBitmap,
-	}
+	var (
+		ab = a.bitmap[:bitmapN]
+		bb = b.bitmap[:bitmapN]
+		buf = make([]uint64, bitmapN)
+		ob =buf[:bitmapN]
```
jaffee: gofmt

roaring/roaring.go
```diff
@@ -3335,9 +3382,16 @@ func popcount(x uint64) uint64 {
 }
 
 func popcountAndSlice(s, m []uint64) uint64 {
+	var (
+		a = s[:bitmapN]
+		b = m[:bitmapN]
+	)
+	_ = a[bitmapN-1]
+	_ = b[bitmapN-1]
```
jaffee: in my testing, this code is consistently faster without these two `_ =...` lines. See jaffee/bounds-check - I ran it on my macbook and on an Azure VM

roaring/roaring.go
```diff
@@ -2138,15 +2138,25 @@ func intersectArrayBitmap(a, b *Container) *Container {
...
+		ob[i] = ab[i] & bb[i]
+		n += int(popcount(ob[i]))
+	}
 
+	output := &Container{
+		bitmap:        ob,
+		n:             n,
+		containerType: containerBitmap,
 	}
-	output.optimize()
```
jaffee: removing this optimize seems like a pretty big change that could affect performance differently for different workloads. Did you have any thoughts on this?

tgruben: My rationale was the overhead of the optimize call was making benchmark comparison of bounds checks inconclusive.  The optimize call was overshadowing the bce by a margin large enough to make bce almost a rounding error

PR #1621, opened 2018-09-05
Adds --field-keys and --index-keys options to pilosa import

### LINKED ISSUE #1570: pilosa import --create doesn't use --string-keys

### Expected behavior

`./pilosa import --create --string-keys --index i1 --field f1 CSV_FILE` should create `i1` and `f1` with `keys=true`.

### Actual behavior

Running the following:
```
curl localhost:10101/index/i1/query -d 'Row(f1="foo")'
```

Results in:
```
{"error":"executing: string 'row' value not allowed unless field 'keys' option enabled"}
```
```
./pilosa import --create --index i1 --field f1 --string-keys 1.csv
```
With the following CSV:
```
foo,bar
```
Then:
```
curl localhost:10101/index/i1/query -d 'Row(f1="foo")'
```

yuce:
## Overview

* I had to add `client.CreateFieldWithOptions` and `client.EnsureFieldWithOptions` functions in order to be able to pass create a field with options.
* `client.CreateFieldWithOptions` takes a `pilosa.FieldOptions` argument, but `http.postFieldRequest` requires the options as `http.fieldOptions`. Had to  convert from the former to the latter. I had to set different fields depending on the type of the field, since e.g., setting `Min` or `Max` is invalid for `set` fields (I'm happy that the tests cover that)
* I had to use `--index-keys` and `--field-keys` since `--index.keys` and `--field.keys` makes a few tests fail
* I didn't remove `--string-keys`. Should I ?

Fixes #1570

## Pull request checklist

## Code review checklist

### DIFF
--- a/client.go
+++ b/client.go
@@ -41,10 +41,12 @@ type InternalClient interface {
 	ImportK(ctx context.Context, index, field string, bits []Bit) error
 	EnsureIndex(ctx context.Context, name string, options IndexOptions) error
 	EnsureField(ctx context.Context, indexName string, fieldName string) error
+	EnsureFieldWithOptions(ctx context.Context, index, field string, opt FieldOptions) error
 	ImportValue(ctx context.Context, index, field string, shard uint64, vals []FieldValue) error
 	ImportValueK(ctx context.Context, index, field string, vals []FieldValue) error
 	ExportCSV(ctx context.Context, index, field string, shard uint64, w io.Writer) error
 	CreateField(ctx context.Context, index, field string) error
+	CreateFieldWithOptions(ctx context.Context, index, field string, opt FieldOptions) error
 	FragmentBlocks(ctx context.Context, uri *URI, index, field, view string, shard uint64) ([]FragmentBlock, error)
 	BlockData(ctx context.Context, uri *URI, index, field, view string, shard uint64, block int) ([]uint64, []uint64, error)
 	ColumnAttrDiff(ctx context.Context, uri *URI, index string, blks []AttrBlock) (map[uint64]map[string]interface{}, error)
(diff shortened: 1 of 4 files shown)


jaffee:
was `--string-keys` broken before 1.0? If so, I think we can safely remove it. If it was functional at 1.0, we need to preserve that functionality though

jaffee:
what exactly was the problem with index.keys and field.keys? My thought was that we'd export the `FieldOptions` and `IndexOptions` attributes to the command struct then create a flag for each option which populates the option struct. e.g.
```go
flags.BoolVar(&Importer.FieldOptions.Keys, "field.keys", false, "use keys=true when creating a field")
flags.IntVar(&Importer.FieldOptions.Min, "field.min", 0, "use a minimum of 0 when creating an int field")
```

yuce:
Honestly I didn't try making `--index.keys` before, since `--index-keys` worked fine, but below are the errors:
```
--- FAIL: TestImportConfig (0.00s)
	root_test.go:121: Problem with test 0, err: 'strconv.ParseBool: parsing "": invalid syntax'
--- FAIL: TestInspectNoPath (0.00s)
	inspect_test.go:33: Command 'inspect' without args should error but: err: 'invalid option in configuration file: index', output: ''
--- FAIL: TestInspectMultiPath (0.00s)
	inspect_test.go:40: Command 'inspect' without args should error but: err: 'invalid option in configuration file: index', output: ''
```
When I have this:
```
flags.BoolVar(&Importer.IndexOptions.Keys, "index.keys", false, "use keys=true when creating an index")
```
It seems like it doesn't like `index` in the option name. When I change it to `indx` there are no errors.

yuce:
I think `--string-keys` was initially used to tell to use interpret the CSV as string rows and columns. It didn't work when creating the schema (hence https://github.com/pilosa/pilosa/issues/1570). There's a comment in `ctl/import.go` about line 53: `TODO: remove this in a future release` so I think we can just remove it.

jaffee:
seems safe to me

yuce:
OK, removed `--string-keys`.

jaffee:
ah, I see the issue - because index is already a flag on its own we can't nest anything under it - because of how it behaves if you use a config file

PR #1622, opened 2018-09-05
Roaring import

travisturner:
## Overview

This is my attempt at pulling the "roaring import" code out of PR #1496.
I wasn't able to remove this code from #1496 without introducing conflicts in the commit replay, but I thought at least we could review this code independently.

@tgruben You should probably confirm that this PR contains all the necessary code.


fragment.go
```diff
@@ -1325,71 +1325,65 @@ func (f *fragment) mergeBlock(id int, data []pairSet) (sets, clears []pairSet, e
...
-	defer f.mu.Unlock()
 	// Verify that there are an equal number of row ids and column ids.
 	if len(rowIDs) != len(columnIDs) {
 		return fmt.Errorf("mismatch of row/column len: %d != %d", len(rowIDs), len(columnIDs))
 	}
 
 	// Disconnect op writer so we don't append updates.
-	f.storage.OpWriter = nil
-
+	localBitmap := roaring.NewBitmap()
```
jaffee: why all these changes to the `bulkImport` method? is this related to the roaring import work?

roaring/roaring_internal_test.go
```diff
@@ -3225,3 +3227,34 @@ func runContainerFunc(f interface{}, c ...*Container) *Container {
 	}
 	return nil
 }
+
+func TestUnmarshalStdRoaring(t *testing.T) {
```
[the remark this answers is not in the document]
tgruben: so which dependency should we  add in order to produce, the java or golang standard roaring?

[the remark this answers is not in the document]
jaffee: golang seems like the easiest - as long as it's only a dependency in the test I think that'd be worthwhile. we can defer the additional testing work to another ticket, but with this much finicky code, I feel strongly that we should use our existing test infrastructure to fully exercise it.

client.go
```diff
@@ -51,6 +51,7 @@ type InternalClient interface {
 	RowAttrDiff(ctx context.Context, uri *URI, index, field string, blks []AttrBlock) (map[uint64]map[string]interface{}, error)
 	SendMessage(ctx context.Context, uri *URI, msg []byte) error
 	RetrieveShardFromURI(ctx context.Context, index, field string, shard uint64, uri URI) (io.ReadCloser, error)
+	ImportRoaringBytes(ctx context.Context, node *Node, index, field string, shard uint64, roaringBytes []byte, forward bool) error
```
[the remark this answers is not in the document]
tgruben: what is the use case for the offset?

fragment.go
tgruben: yes, there was a fundamental change to how the snapshot method works, basically a local bitmap is built, then merged into the existing shard bitmap as opposed to setting bits on the larger file.

api.go
```diff
@@ -294,6 +295,42 @@ func (api *API) Field(_ context.Context, indexName, fieldName string) (*Field, e
 	return field, nil
 }
 
+// ImportRoaringBytes fast loading of standard roaring format
+func (api *API) ImportRoaringBytes(ctx context.Context, roaringBytes []byte, indexName, fieldName string, shard uint64, forward bool) (err error) {
```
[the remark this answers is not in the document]
tgruben: i think that was an artifact of an integration test, but i can't really recall.  I can't seem to think of a reason why it should remain, so i'll remove

client.go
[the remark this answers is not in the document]
jaffee: being able to import columns greater than 2^32

roaring/roaring_internal_test.go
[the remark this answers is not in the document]
tgruben: i will create a ticket

api.go
```diff
@@ -294,6 +295,42 @@ func (api *API) Field(_ context.Context, indexName, fieldName string) (*Field, e
 	return field, nil
 }
 
+// ImportRoaringBytes fast loading of standard roaring format
+func (api *API) ImportRoaringBytes(ctx context.Context, roaringBytes []byte, indexName, fieldName string, shard uint64 ) (err error) {
```
jaffee: `shard uint64 )` ...gofmt?

fragment.go
```diff
@@ -1425,6 +1419,50 @@ func (f *fragment) importValue(columnIDs, values []uint64, bitDepth uint) error
...
+
+	if f.storage.Count() > 0 {
+		bm = f.storage.Union(bm)
+	}
+
+	for _,rowID := range set {
+		n := bm.CountRange(rowID*ShardWidth, (rowID+1)*ShardWidth)
+		f.cache.BulkAdd(rowID, n)
+	}
+
```
jaffee: @tgruben do we need to invalidate the cache here?

tgruben: most definately

jaffee:
I made a bunch of changes/fixes to this today, and added some tests. It still needs external docs (I added some doc strings). can you take a look @tgruben @travisturner

travisturner:
thanks @jaffee, I'll take a look

fragment.go
```diff
@@ -1323,73 +1323,67 @@ func (f *fragment) mergeBlock(id int, data []pairSet) (sets, clears []pairSet, e
...
-	defer f.mu.Unlock()
 	// Verify that there are an equal number of row ids and column ids.
 	if len(rowIDs) != len(columnIDs) {
 		return fmt.Errorf("mismatch of row/column len: %d != %d", len(rowIDs), len(columnIDs))
 	}
 
 	// Disconnect op writer so we don't append updates.
-	f.storage.OpWriter = nil
-
+	localBitmap := roaring.NewBitmap()
```
travisturner: Continuing the previously resolved converstion, I think it would be helpful to add a comment above the declaration of `localBitmap` that describes what is happening. Then the "disconnect op writer..." comment should be moved down to be above the "... = nil" line.

fragment.go
```diff
@@ -1323,73 +1323,67 @@ func (f *fragment) mergeBlock(id int, data []pairSet) (sets, clears []pairSet, e
...
+	localBitmap.OpWriter = nil
 	// Process every bit.
 	// If an error occurs then reopen the storage.
 	lastID := uint64(0)
-	if err := func() error {
-		set := make(map[uint64]struct{})
-		for i := range rowIDs {
-			rowID, columnID := rowIDs[i], columnIDs[i]
+	set := make(map[uint64]struct{})
```
travisturner: consider renaming this to something like `rowSet`, or something not `set`

fragment.go
```diff
@@ -1323,73 +1323,67 @@ func (f *fragment) mergeBlock(id int, data []pairSet) (sets, clears []pairSet, e
...
+		delete(f.checksums, int(rowID/HashBlockSize))
 	}
 
-	// Write the storage to disk and reload.
-	if err := f.snapshot(); err != nil {
-		return errors.Wrap(err, "snapshotting")
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	//f.storage.Unmmap()
```
travisturner: is this comment indicating anything?

fragment.go
```diff
@@ -1323,73 +1323,67 @@ func (f *fragment) mergeBlock(id int, data []pairSet) (sets, clears []pairSet, e
...
+	// Update cache counts for all rows.
```
travisturner: It seems like this comment should be something like "Merge localBitmap into existing storage."

fragment.go
```diff
@@ -1323,73 +1323,67 @@ func (f *fragment) mergeBlock(id int, data []pairSet) (sets, clears []pairSet, e
...
+	var results *roaring.Bitmap
+	if f.storage.Count() > 0 {
+		results = f.storage.Union(localBitmap)
+	} else {
+		results = localBitmap
 	}
 
-	return nil
+	for rowID := range set {
```
travisturner: I think the "update counts" comment is meant to go before this line.

fragment.go
```diff
@@ -1452,6 +1491,13 @@ func track(start time.Time, message string, stats StatsClient, logger Logger) {
 }
 
 func (f *fragment) snapshot() error {
+	return snapshot(f, f.storage)
+}
+
+// snapshot writes the fragment f with bm as the data. It is unprotected, and
+// f.mu must be locked when calling it.
+func snapshot(f *fragment, bm *roaring.Bitmap) error { // nolint: interfacer
```
travisturner: calling this function `snapshot` is confusing, not only because of it's similarity to `fragment.snapshot`, but because I don't think it's actually doing a snapshot, but rather an unprotected write.

http/handler.go
```diff
@@ -177,6 +177,7 @@ func (h *Handler) populateValidators() {
 	h.validators["GetFragmentData"] = queryValidationSpecRequired("index", "field", "shard")
 	h.validators["PostFragmentData"] = queryValidationSpecRequired("index", "field", "shard")
 	h.validators["GetFragmentBlocks"] = queryValidationSpecRequired("index", "field", "view", "shard")
+	h.validators["ImportRoaringBitmap"] = queryValidationSpecRequired("index", "field", "shard")
```
travisturner: This validator is not set up correct. In fact, I thing there are a few here that are not correct.

Also, in the client the method is `ImportRoaringBytes`.

http/handler.go
```diff
@@ -217,6 +218,7 @@ func newRouter(handler *Handler) *mux.Router {
 	router.HandleFunc("/index/{index}/field/{field}", handler.handlePostField).Methods("POST")
 	router.HandleFunc("/index/{index}/field/{field}", handler.handleDeleteField).Methods("DELETE")
 	router.HandleFunc("/index/{index}/field/{field}/import", handler.handlePostImport).Methods("POST")
+	router.HandleFunc("/index/{index}/field/{field}/importroaring/{shard}", handler.handlePostRoaringImport).Methods("POST")
```
travisturner: I think this should be `import-roaring` and the handler should be `handlePostImportRoaring`

Maybe elsewhere we drop the `Bytes` from `ImportRoaringBytes` for consistency.

http/handler.go
```diff
@@ -1427,3 +1429,48 @@ func GetHTTPClient(t *tls.Config) *http.Client {
...
 	return &http.Client{Transport: transport}
 }
+
+// handlPostRoaringImport
+func (h *Handler) handlePostRoaringImport(w http.ResponseWriter, r *http.Request) {
+	// Verify that request is only communicating over protobufs.
+	if r.Header.Get("Content-Type") != "application/x-binary" {
+		http.Error(w, "Unsupported media type", http.StatusUnsupportedMediaType)
+		return
+	} /*else if validHeaderAcceptJSON(r.Header) {
```
travisturner: do we need this json comment?

roaring/roaring_internal_test.go
```diff
@@ -3225,3 +3227,37 @@ func runContainerFunc(f interface{}, c ...*Container) *Container {
...
```
travisturner: let's expand the word `Standard` here (or maybe even refer to it as "Official"?)

travisturner:
@jaffee @tgruben since I created this PR I can only leave comments, I can't approve or decline it. Anyway, I've left several comments.

fragment.go
jaffee: @tgruben any thoughts on this?

http/handler.go
tgruben: No need for that comment

http/client.go
```diff
@@ -531,6 +525,42 @@ func (c *InternalClient) marshalImportValuePayload(index, field string, shard ui
...
+	if err != nil {
+		return errors.Wrap(err, "executing request")
+	}
+	defer resp.Body.Close()
+
+	// Validate status code.
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("invalid status: %d", resp.StatusCode)
+	}
+	io.Copy(ioutil.Discard, resp.Body)
```
jaffee: err... this should probably reading the body and returning an error `if resp.Err != ""`, right?

tgruben: yep, my first guess was just only http errors where sent back, but i was wrong

travisturner: yes, good catch

jaffee:
I found some lingering issues in the client and API method around how queries were getting forwarded. We needed the remote bool to signal when a query should be processed only on the receiving node, or forwarded around.

I also fixed a data race where the roaring data was getting modified by the local node concurrently to being forwarded to remote nodes.

I *think* this is about ready now, but I'm going to run the race detector and do one last pass.

jaffee:
I think this is actually ready now. Would like one more approval since I made so many changes.

fragment.go
```diff
@@ -1323,73 +1323,70 @@ func (f *fragment) mergeBlock(id int, data []pairSet) (sets, clears []pairSet, e
...
+		// Write to storage.
+		_, err = localBitmap.Add(pos)
+		if err != nil {
+			return err
+		}
+		// Reduce the StatsD rate for high volume stats
+		f.stats.Count("ImportBit", 1, 0.0001)
+		// import optimization to avoid linear foreach calls
+		// slight risk of concurrent cache counter being off but
+		// no real danger
```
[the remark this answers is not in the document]
jaffee: I know this is merged now, but does anyone have any insight on this? @tgruben @travisturner ?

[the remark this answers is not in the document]
travisturner: sorry, I reviewed this yesterday and couldn't come to an understanding of this comment. was hoping @tgruben would chime in.

PR #1624, opened 2018-09-05
Add DeleteAvailableShard()

benbjohnson:
## Overview

This commit adds the ability to remove an 'available shard' from the shard cache. This does not affect shards known to be available because of local data.

Fixes #

## Pull request checklist

## Code review checklist


api.go
```diff
@@ -327,6 +327,38 @@ func (api *API) DeleteField(_ context.Context, indexName string, fieldName strin
...
+
+	// Send the delete field message to all nodes.
+	err := api.server.SendSync(
+		&DeleteAvailableShardMessage{
+			Index:   indexName,
+			Field:   fieldName,
+			ShardID: shardID,
+		})
+	if err != nil {
+		api.server.logger.Printf("problem sending DeleteField message: %s", err)
```
travisturner: this log line refers to Field instead of Shard.

http/handler.go
```diff
@@ -217,6 +217,7 @@ func newRouter(handler *Handler) *mux.Router {
 	router.HandleFunc("/index/{index}/field/{field}", handler.handlePostField).Methods("POST")
 	router.HandleFunc("/index/{index}/field/{field}", handler.handleDeleteField).Methods("DELETE")
 	router.HandleFunc("/index/{index}/field/{field}/import", handler.handlePostImport).Methods("POST")
+	router.HandleFunc("/index/{index}/field/{field}/remote-available-shards/{shardID}", handler.handleDeleteRemoteAvailableShard).Methods("DELETE")
```
travisturner: We should move this to the `/internal` handler list. I think this would typically get called by a `deleteShard` function executed on a Pilosa node. I suppose there may be cases where we would want to make the shard unavailable without actually removing the data, but I feel like that would also be as a result of an internal function running (a resize event, for example).

field.go
```diff
@@ -214,13 +214,29 @@ func (f *Field) AvailableShards() *roaring.Bitmap {
...
+func (f *Field) AddRemoteAvailableShards(b *roaring.Bitmap) {
 	f.mu.Lock()
 	defer f.mu.Unlock()
 	f.remoteAvailableShards = f.remoteAvailableShards.Union(b)
 }
 
+// RemoveAvailableShard removes a shard from the bitmap cache.
+//
+// NOTE: This can be overridden on the next sync so all nodes should be updated.
+func (f *Field) RemoveAvailableShard(v uint64) error {
```
travisturner: we now have a `saveAvailableShards` method which needs to be called along with this method to persist the `remoteAvailableShards` to disk.

benbjohnson:
@travisturner Pull request fixed up in 3c59801. Ready for re-review.

field.go
travisturner: @benbjohnson did this make sense? I think this still needs to be implemented.

benbjohnson: @travisturner It was pulled in through a rebase: https://github.com/pilosa/pilosa/blob/2dfe881e9d7fad88b0832e8861db3d0b97caa672/field.go#L233-L240

Is that what you were thinking?

travisturner: In the case where `RemoveAvailableShard()` is called (in `api.DeleteAvailableShard()` and `server.ReceiveMessage()`), I don't think that `saveAvailableShards()` is called. So the representation of `remoteAvailableShards` on disk still includes the shard that was removed. Even if it went through `AddRemoteAvailableShards()`, that method is doing a `Union`, so it wouldn't remove anything anyway.

benbjohnson:
@travisturner I rebased this PR which invalidated your approval. Can you re-approve this PR?

PR #1646, opened 2018-09-17
Adds DirectAdd function to roaring.Bitmap

yuce:
## Pull request checklist

## Code review checklist


roaring/roaring.go
```diff
@@ -167,6 +167,13 @@ func (b *Bitmap) Add(a ...uint64) (changed bool, err error) {
 	return changed, nil
 }
 
+// DirectAdd adds values to the bitmap by bypassing the op log.
+func (b *Bitmap) DirectAdd(values []uint64) {
```
jaffee: I don't like that this method takes a slice while `Add` takes a variadic argument. I think that requiring a slice adds unnecessary complication and overhead in any case. 

I'd advocate for simply renaming `Bitmap.add` to `Bitmap.DirectAdd`. This removes an extra method call, and will prevent some users of the api from having to wrap their input in a slice. Other users will have to loop over DirectAdd instead of just calling it, but I think this is a worthwhile tradeoff.

yuce: That makes sense. I've changed the signature of `DirectAdd` so it adds a single value. I didn't rename `add` since there are a bunch of places that uses `add`. I think we should rename those in another PR.

jaffee:
I think this PR is the place to do it honestly... no point in adding and then removing a new method. Do you have `gorename` working? I can do it in one command and push if that's easier.

Issue #1500, opened 2018-07-12
Explore GroupBy


tgruben:
Add the ability provided a list of fields, return group count results

`GroupBy(fields=[cab_type,year], limit=100)`
### Expected behavior
Output:
{ fields:[yellow, 2010], total:193484}
{ fields:[green, 2010], total:3723}
{ fields:[yellow, 2011], total:4892845}
{ fields:[green, 2011], total:32003}

travisturner:
See #1496 for exploration work.

Issue #1482, opened 2018-07-10
Explore column/row iterations and "SELECT field..." query


jaffee:
### Description

Play around with the ability to expose row and column iteration internally, and try to get a sense for what the APIs should look like, both at the internal API level, and potentially the pql.Call representation.

### Success criteria (What criteria will consider this ticket closeable?)
We should be able to make more focused tickets that break up further work.

PR #1647, opened 2018-09-17
copy non roaring-import code from Todds's row-iterate PR

jaffee:
## Overview

- Adds a PQL function Rows which will return the rowid's for a given field all the arguments are optional except for the field

`Row(field=aField,column=10,offset=100,limit=1000)`

The overhead in generating this list was lower than I expected.

- Adds PQL function GroupBy which returns product of all fieldValues with Count
`GroupBy(fields=[general,sub])`

`GroupBy(fields=[general,sub:10:1,detail:100:])`

Fixes: #1500, #1482 

## Pull request checklist

## Code review checklist


executor.go
```diff
@@ -194,6 +196,12 @@ func (e *executor) executeCall(ctx context.Context, index string, c *pql.Call, s
 	case "TopN":
 		e.Holder.Stats.CountWithCustomTags(c.Name, 1, 1.0, []string{indexTag})
 		return e.executeTopN(ctx, index, c, shards, opt)
+	case "Rows":
```
jaffee: Copied comments from previous PR:

	
  @travisturner
travisturner 27 days ago  Member
What do you think about this being called Column instead of Rows? In that case it would be similar to the Row query. Additionally, instead of returning RowIDs it could return a *Column (or Column) potentially containing IDs and attributes. I haven't thought this through very much, so I'm really just throwing out for discussion.

  @tgruben
tgruben 26 days ago  Member
I like the concept of Column returning all the details about a specific column (rows and attributes). I think the label breaks down when we are trying to interact with all/some rows contained in a field. I feel we need an additional term for dealing with the field based issues. Perhaps introduce some ...Info based terms ie FieldInfo, ColumnInfo,RowInfo. Still feels hackish, hoping this will trigger some synapse action elsewhere.

executor.go
```diff
@@ -714,14 +722,369 @@ func (e *executor) executeDifferenceShard(ctx context.Context, index string, c *
...
+// getGroupByFilterFunction returns a rowFilter based on the
+// field directive provided.
+func getGroupByFilterFunction(fieldDirective string) (rowFilter, error) {
+	parts := strings.Split(fieldDirective, ":")
+	hasLimit := false
+	hasOffset := false
+	limit := uint64(0)
+	offset := uint64(0)
+	var err error
+	// fieldDirective can have one of the following forms:
```
jaffee: this comment should be part of the doc string on the function IMO - even though it isn't exported

executor.go
```diff
@@ -714,14 +722,369 @@ func (e *executor) executeDifferenceShard(ctx context.Context, index string, c *
...
+	row  *Row
+	gCnt GroupCount
+}
+
+// product generates the cartesian product of the input
+// using tail recursion.
+func product(input [][]gbi) []ppi {
+	if len(input) == 0 { // base return empty list
+		return []ppi{
+			{gCnt: GroupCount{Group: make([]FieldRow, 0)}},
```
jaffee: why does this return one empty record instead of no records? @tgruben

jaffee: ping @tgruben

fragment.go
```diff
@@ -1776,39 +1794,21 @@ func (f *fragment) rows() []uint64 {
...
+			if !addRow {
+				break
+			}
 		}
-
-		if c.Contains(colVal) {
-			rows = append(rows, vRow)
+		if addRow {
+			lastRow = vRow
+			rows = append(rows, key>>shardVsContainerExponent)
```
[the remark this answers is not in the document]
jaffee: good call... not sure how that snuck in

executor.go
```diff
@@ -714,14 +722,327 @@ func (e *executor) executeDifferenceShard(ctx context.Context, index string, c *
...
+		i++
+	}
+	for j < len(other) && len(result) < limit {
+		result = append(result, other[j])
+		j++
+	}
+	return result
+}
+
+func (e *executor) executeGroupBy(ctx context.Context, index string, c *pql.Call, shards []uint64, opt *execOptions) ([]GroupCount, error) {
```
tgruben: Logically this makes sense given the limit/offset constraints, but the number of cluster round trips are large.   
Total Round Trips = # Rows calls + GroupBy 
I would like to package that into one trip and avoid the overhead of sending rows across the network

jaffee: yeah, we could do them async (so all at once), which would make it roughly 1 + GroupBy, but it still isn't ideal. The problem is that `Rows` calls which have a column argument should only be executed on the shard where that column lives, and so we *must* do those first before passing the resultant set of row ids to the GroupBy call on each shard. I think ideally, we'd do just the calls with `column` first and then pass the next ones on to each shard. In fact... I think we might need to do this for the code to be logically correct. I haven't yet written a test to verify this, but consider the following situation:

```
| field-row | shard 0 | shard 1 |
|-----------+---------+---------|
| a-0       | yes     | no      |
| a-1       | no      | yes     |
| b-0       | no      | yes     |
| b-1       | yes     | no      |
```

What does `GroupBy(Rows(field=a), Rows(field=b), limit=1)` return with the current logic? Remember that the lower limit of the `Rows` and `GroupBy` call is passed to each `Rows` query.

tgruben: Calling async doesn't limit the total amount of network traffic in my universe :).  I don't understand your table, what are the columns that are being intersected?

jaffee: Calling async doesn't limit the total amount of network traffic
true, but it affects the total latency which is arguably more important.

    I don't understand your table, what are the columns that are being intersected?

the leftmost is showing rows in various fields - the right two columns are showing whether or not that shard has data for that row.

The correct output for the query I gave would be something like `{Group: a.0-b.1, count: N}`, but I don't think it would return any results with the way the code is currently written. The rows passed to executeGroupByShard would just be [[0], [0]] and since no shard has row 0 in both a and b, no data would be returned.

jaffee:
@travisturner @tgruben This is now ready (at long last!).  It's pretty involved - happy to walk through it in person, over the phone, etc.

executor.go
```diff
@@ -2096,3 +2507,144 @@ func isString(v interface{}) bool {
...
+				}
+			}
+		}
+	}
+
+	for i := 1; i < len(gbi.rows)-1; i++ {
+		gbi.rows[i].row = gbi.rows[i].row.Intersect(gbi.rows[i-1].row)
+	}
+
+	return gbi
```
travisturner:
```suggestion
	return gbi, nil
```

executor.go
```diff
@@ -777,14 +783,304 @@ func (e *executor) executeDifferenceShard(ctx context.Context, index string, c *
...
+			return 1
+		}
+	}
+	return 0
+}
+
+func (e *executor) executeGroupByShard(_ context.Context, index string, c *pql.Call, shard uint64, childRows []RowIDs) ([]GroupCount, error) {
+	iter, err := newGroupByIterator(childRows, c.Children, index, shard, e.Holder)
+	if err != nil {
+		return errors.Wrapf(err, "getting group by iterator for shard %d", shard)
```
travisturner:
```suggestion
		return nil, errors.Wrapf(err, "getting group by iterator for shard %d", shard)
```

zhyhang:
Rows call lack column key to id translation?

travisturner:
@zhyhang I implemented that somewhere. Maybe it didn't get pulled into this PR. I'll take a look and make sure that work gets included here. Thanks for pointing this out.

travisturner:
Thanks @zhyhang 
I added a commit to a PR into this PR: https://github.com/jaffee/pilosa/pull/5

executor.go
```diff
@@ -777,14 +783,301 @@ func (e *executor) executeDifferenceShard(ctx context.Context, index string, c *
...
+	other, err := e.mapReduce(ctx, index, shards, c, opt, mapFn, reduceFn)
+	if err != nil {
+		return nil, err
+	}
+	results, _ := other.(RowIDs)
+	return results, nil
+}
+
+func (e *executor) executeRowsShard(_ context.Context, index string, c *pql.Call, shard uint64) (RowIDs, error) {
+	// Fetch index.
```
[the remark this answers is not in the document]
jaffee: agree - could cause confusion otherwise. Might be worth rethinking at some point though. I wonder if we could work the index not found error into the field not found error.

executor.go
```diff
@@ -2096,3 +2504,135 @@ func isString(v interface{}) bool {
...
+		gbi.rows[i].row = nextRow
+		gbi.rows[i].id = rowID
+		if hasPrev && rowID != prev {
+			ignorePrev = true
+		}
+		if wrapped {
+			for j := i - 1; j >= 0; j-- {
+				nextRow, rowID, wrapped := gbi.rowIters[j].Next()
+				if nextRow == nil {
+					gbi.done = true
```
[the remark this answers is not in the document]
jaffee: I added a test which exercises this.

travisturner:
@jaffee this all looks good to me.
it's failing the metalinter, and needs to be updated with master.

PR #1651, opened 2018-09-18
treat import timestamps as UTC

### LINKED ISSUE #1650: Importing timestamps causes unexpected results with local time zone.

### Expected behavior
Importing the date `2018-01-01T00:00` with `timeQuantum = "YMD"` will result in data in views:
- 2018
- 201801
- 20180101

### Actual behavior
Because of my local time zone setting (-0600 CST), importing the date `2018-01-01T00:00` with `timeQuantum = "YMD"` results in data in views:
- 2017
- 201712
- 20171231

```
curl -X POST http://localhost:10101/index/i -d '{"options": {}}'
curl -X POST http://localhost:10101/index/i/field/f -d '{"options": {"type": "time", "timeQuantum": "YMD"}}'

pilosa import -i i -f f ~data.csv
```
where `data.csv` contains:
```
1,101,2018-01-01T00:00
```

### Possible Solution
I've determined that the timezone offset is applied in `api.Import` with this line:
```
t := time.Unix(0, ts)
```
In that case, `ts` is the number of nanoseconds past epoch.

It seems like we should modify this so that the timezone of the local machine is not considered in the conversion of nanoseconds to a `time.Time` object. Any objections to my changing this to ignore timezone (i.e. use UTC)?

MacOSX, Time Zone (-0600 CST)

travisturner:
## Overview

This addresses what I consider a bug in timestamp imports. See #1650 for a description.

Fixes #1650

## Pull request checklist

## Code review checklist

### DIFF
--- a/api.go
+++ b/api.go
@@ -732,7 +732,7 @@ func (api *API) Import(_ context.Context, req *ImportRequest) error {
 		if ts == 0 {
 			continue
 		}
-		t := time.Unix(0, ts)
+		t := time.Unix(0, ts).UTC()
 		timestamps[i] = &t
 	}
 
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"math/rand"
 	"reflect"
 	"sort"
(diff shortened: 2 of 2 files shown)


travisturner:
I'm not sure how to test this. I basically need the pilosa server to be running on a VM with local time set to something non-UTC. Or I need something like a [time.SetLocal()](https://github.com/golang/go/issues/10701) function, which doesn't exist.

travisturner:
I supposed I could write the test such that it would apply when run locally (on my non-UTC laptop for example). But there wouldn't be a guarantee that the test would be exercised correctly in a CI test suite.

Anyone have any other ideas?

jaffee:
I agree that if the local time is set to UTC, this will be hard to test reliably, but I think that between dev environments and CI, the following would probably catch any regressions before they made it to a release:

1. Import some data into a field that is at the very beginning of a year (as in your example), and at the very end of the same year. 
2. Verify that no views are created for either the previous year, or the next year.

travisturner:
@jaffee I added a test to cover this. I confirmed that the test fails on my local machine if I remove the `.UTC()` conversion.

PR #1659, opened 2018-09-21
WIP: fixes pass-by-value issue in proto decode

travisturner:
## Overview

Fixes a pass-by-value issue in proto decoding of `NodeStatus`.


server.go
```diff
@@ -606,6 +608,21 @@ func (s *Server) node() Node {
...
+	if !ok {
+		s.logger.Printf("invalid message type: %T", pb)
+	}
+
+	switch instruction.Type {
+	case "ShareNodeStatus":
+		if err := s.cluster.shareNodeStatus(); err != nil {
+			s.logger.Printf("error sharing node status: %s", err)
+		}
+	}
```
jaffee: no default in this switch?

cluster.go
```diff
@@ -1044,8 +1044,56 @@ func (c *cluster) unprotectedSetStateAndBroadcast(state string) error {
...
+			is.Fields = append(is.Fields, &FieldStatus{
+				Name:            f.Name,
+				AvailableShards: availableShards,
+			})
+		}
+		m.Indexes = append(m.Indexes, is)
+	}
+
+	go func() {
+		c.broadcaster.SendSync(m)
```
jaffee: why call this in a goroutine instead of calling SendAsync? Why not call it synchronously?

In any case, is there any reason not to check the error and log or return it?

server.go
```diff
@@ -553,6 +553,8 @@ func (s *Server) receiveMessage(m Message) error {
 		s.holder.recalculateCaches()
 	case *NodeEvent:
 		s.cluster.ReceiveEvent(obj)
+	case *Instruction:
+		s.handleInstruction(obj)
```
jaffee: I'm not sure I understand the need for this generic instruction type. To me, it feels like it will hide internal communication complexity instead of making it obvious in our proto definition. I'm not necessarily opposed to it... just want to understand why you chose to go this route.

api.go
```diff
@@ -801,6 +801,8 @@ func importExistenceColumns(index *Index, columnIDs []uint64) error {
 }
 
 // MaxShards returns the maximum shard number for each index in a map.
+// TODO (2.0): This method has been deprecated. Instead, use
+// AvailableShardsByIndex.
```
jaffee: is it possible now to remove the `/internal/shards/max` endpoint?

server.go
travisturner: I don't really have a good argument for using `Instruction` over having a specific message type for each instruction (and there's currently only one instruction type anyway). I can change it to be specific to the instruction type.

api.go
travisturner: That's a good question. I'll check to see if any of the clients are still using that.

server.go
travisturner: I guess the default would be to log an error. With that said, if we remove the `Instruction` wrapper, this switch statement will go away anyway.

PR #1666, opened 2018-09-25
Store() function

### LINKED ISSUE #637: Store bitmap result directly
### Description
There is currently no way to compute a bitmap query and store it directly as a new row. 

Example: `Store(<bitmap_query>, frame=foo, rowID=10)` -> `{"results":[{"rowID": 10}]}`
(maybe rowID input is optional)

Any query that returns a bitmap result can be immediately written to a new frame+row, without returning the full bitmap result to the client.

travisturner:
## Overview

This PR implements `Store(<row_query>, <fieldname>=10)`. Currently, it's only supported on `set` fields. It's not clear to me yet how we would support this on `mutex` and `bool` fields. And I don't think it makes sense to store a `row` result into `int` or `time` fields (although I suppose if `Store()` took an optional timestamp, it could apply the row to all time quantums).

Also, the PQL function is `Store()`, but internally everything is `SetRow()` because it seemed like a complement to `ClearRow()`.

Fixes #637 (at least the first phase of it)

TODO:
- [x] Docs
- [x] Decide if we want to support `time` fields.

## Pull request checklist

## Code review checklist

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -1226,11 +1226,7 @@ func (e *executor) executeSetRow(ctx context.Context, index string, c *pql.Call,
 	if field == nil {
 		return false, ErrFieldNotFound
 	}
-
-	switch field.Type() {
-	case FieldTypeSet:
-		// These field types support SetRow().
-	default:
+	if field.Type() != FieldTypeSet {
 		return false, fmt.Errorf("SetRow() is not supported on %s field types", field.Type())
 	}
 
@@ -1298,11 +1294,11 @@ func (e *executor) executeSetRowShard(ctx context.Context, index string, c *pql.
 			return false, errors.Wrapf(err, "creating fragment: %d", shard)
 		}
 	}
-	cleared, err := fragment.setRow(src, rowID)
+	set, err := fragment.setRow(src, rowID)
 	if err != nil {
 		return false, errors.Wrapf(err, "setting row %d on view %s shard %d", rowID, viewStandard, shard)
 	}
-	changed = changed || cleared
+	changed = changed || set
 
 	return changed, nil
 }
--- a/executor.go
+++ b/executor.go
@@ -184,6 +184,8 @@ func (e *executor) executeCall(ctx context.Context, index string, c *pql.Call, s
 		return e.executeClearBit(ctx, index, c, opt)
 	case "ClearRow":
 		return e.executeClearRow(ctx, index, c, shards, opt)
+	case "Store":
+		return e.executeSetRow(ctx, index, c, shards, opt)
 	case "Count":
 		e.Holder.Stats.CountWithCustomTags(c.Name, 1, 1.0, []string{indexTag})
 		return e.executeCount(ctx, index, c, shards, opt)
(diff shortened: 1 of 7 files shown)


executor.go
```diff
@@ -1213,6 +1215,98 @@ func (e *executor) executeClearRowShard(_ context.Context, index string, c *pql.
...
+	fieldName, err := c.FieldArg()
+	if err != nil {
+		return false, errors.New("SetRow() argument required: field")
+	}
+	field := e.Holder.Field(index, fieldName)
+	if field == nil {
+		return false, ErrFieldNotFound
+	}
+
+	switch field.Type() {
```
tgruben: I would prefer an if for a one state check
 ` if field.Type() != FieldTypeSet{
  return false, fmt.Errorf("SetRow() is not supported on %s field types", field.Type())
}`

fragment.go
```diff
@@ -491,6 +491,56 @@ func (f *fragment) unprotectedClearBit(rowID, columnID uint64) (changed bool, er
...
+	// First container of the row in storage.
+	headContainerKey := rowID << shardVsContainerExponent
+
+	// Remove every existing container in the row.
+	for i := uint64(0); i < (1 << shardVsContainerExponent); i++ {
+		f.storage.Containers.Remove(headContainerKey + i)
+	}
+
+	// From the given row, get the rowSegment for this shard.
+	seg := row.segment(f.shard)
```
tgruben: i think this check needs to be before the remove

travisturner: In a multi-shard scenario, if you don't remove the existing row, you could potentially have updated data on one shard, and old data on another.

executor.go
travisturner: Yeah, this makes sense. I left it this way in the case that I was going to modify this PR to include the `time` field type.

jaffee:
I vote "no" on time fields - keep things simple for now.

tgruben:
I vote only set fields

travisturner:
Since we're only supporting `set` fields as the destination field, I'm taking this PR out of WIP.

alanbernstein:
This supports an arbitrary bitmap call, right? I would suggest adding a second docs example explicitly showing something simple like `Store(Intersect(x, y), z)`, to demonstrate that you can sort of perform computations this way. Maybe a test that does something like that as well.

I agree on only supporting `set` fields for now, but some thoughts on future directions... Storing to `mutex` seems reasonable - doesn't it make sense (conceptually at least) to write some bits to one row of a `mutex` field, then clear any conflicting bits as necessary? 

I can imagine performing arithmetic operations on `int` fields and storing the entire result into another `int` field. This is obviously far-future kind of work though, and its not clear if it would even make sense to use Pilosa for this.

travisturner:
@alanbernstein I added another example to the query docs.

I think that `mutex` support is do-able, but it will likely require that we handle the individual columns, checking for existing values on each.

I like the idea of storing arithmetic operations performed on `int` fields. That made me wonder about doing arithmetic operations on an `int` field in place. For example, add 7 to all columns in `int` field `f`. I think this would require us to do multiple passes and re-writing of rows, and I'm not even sure that's useful, but something we could likely do fairly easily.

PR #1668, opened 2018-09-28
Proxy Support for import

### LINKED ISSUE #1589: /import endpoint should forward to the correct node (which owns the shard)
It should probably handle the replication too - right now the clients issue separate requests for replicas.

This will make it easier to support things like imports behind a load balancer as in #1129

Todd's import-roaring work does this, and so may be a good reference. We've also talked about functionality like this for imports with string keys in #1557

tgruben:
## Overview

Provides the capability to forward import request to the appropriate node behind a proxy.  The go-pilosa client would have to be updated to just send to the proxy but still provide the shard.  This is been integration tested but no automated tests have been provided.  Feedback regarding testing is welcome.
Fixes #1589

### DIFF
--- a/api.go
+++ b/api.go
@@ -751,7 +751,7 @@ func (api *API) Import(ctx context.Context, req *ImportRequest) error {
 		return errors.Wrap(err, "importing")
 	}
 	//if here need to forward to the correct node
-	bits := make([]Bit, len(req.ColumnIDs), len(req.ColumnIDs))
+	bits := make([]Bit, len(req.ColumnIDs))
 	for i := range req.ColumnIDs {
 		bit := Bit{}
 		if len(req.RowIDs) > 0 {
--- a/api.go
+++ b/api.go
@@ -750,16 +750,26 @@ func (api *API) Import(ctx context.Context, req *ImportRequest) error {
 		}
 		return errors.Wrap(err, "importing")
 	}
-
+	//if here need to forward to the correct node
 	bits := make([]Bit, len(req.ColumnIDs), len(req.ColumnIDs))
 	for i := range req.ColumnIDs {
-		bits[i] = Bit{
-			RowID:     req.RowIDs[i],
-			ColumnID:  req.ColumnIDs[i],
-			RowKey:    req.RowKeys[i],
-			ColumnKey: req.ColumnKeys[i],
-			Timestamp: req.Timestamps[i],
+		bit := Bit{}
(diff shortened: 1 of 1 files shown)


api.go
```diff
@@ -691,63 +691,90 @@ func (api *API) FieldAttrDiff(_ context.Context, indexName string, fieldName str
...
+		return errors.Wrap(err, "importing")
+	}
+	//if here need to forward to the correct node
+	bits := make([]Bit, len(req.ColumnIDs))
+	for i := range req.ColumnIDs {
+		bit := Bit{}
+		if len(req.RowIDs) > 0 {
+			bit.RowID = req.RowIDs[i]
+		}
+		if len(req.ColumnIDs) > 0 {
```
travisturner: this particular `if` statement is not necessary.

api.go
```diff
@@ -691,63 +691,90 @@ func (api *API) FieldAttrDiff(_ context.Context, indexName string, fieldName str
...
+		}
+		bits[i] = bit
 	}
-
-	// Import into fragment.
-	err = field.Import(req.RowIDs, req.ColumnIDs, timestamps)
-	if err != nil {
-		api.server.logger.Printf("import error: index=%s, field=%s, shard=%d, columns=%d, err=%s", req.Index, req.Field, req.Shard, len(req.ColumnIDs), err)
+	if err := api.server.defaultClient.Import(ctx, req.Index, req.Field, req.Shard, bits); err != nil {
+		return errors.Wrap(err, "importing forward")
```
travisturner: change error message to be "forwarding import"

api.go
```diff
@@ -691,63 +691,90 @@ func (api *API) FieldAttrDiff(_ context.Context, indexName string, fieldName str
...
+		if err != nil {
+			api.server.logger.Printf("import error: index=%s, field=%s, shard=%d, columns=%d, err=%s", req.Index, req.Field, req.Shard, len(req.ColumnIDs), err)
+		}
+		return errors.Wrap(err, "importing")
+	}
+	//if here need to forward to the correct node
+	bits := make([]Bit, len(req.ColumnIDs))
+	for i := range req.ColumnIDs {
+		bit := Bit{}
+		if len(req.RowIDs) > 0 {
```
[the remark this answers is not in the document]
tgruben: That is my assumption, either you send keys or ids, but the number should match up

tgruben:
closing this request to refactor

PR #1672, opened 2018-10-01
Index trackExistence is true by default Fixes #1637

### LINKED ISSUE #1637: [proposal] enable trackExistence by default
### Description

I think that we should seriously consider enabling trackExistence by default for new indexes before we cut a new release. This won't strictly be a breaking change, but will reduce performance somewhat for imports. My concern is that new users will create an index with the default option, and then get frustrated when the `Not` query, and any other functionality which depends on existence is not available to them. I also have an inkling more functionality will depend on existence in the future, and existing users won't be able to take advantage of it unless they've specifically enabled it in the past whether they needed it at the time or not.

We've discussed enabling trackExistence by default, made a decision, and implemented it.

yuce:
## Overview

Fixes #1637 

## Pull request checklist

## Code review checklist

### DIFF
--- a/http/handler.go
+++ b/http/handler.go
@@ -521,7 +521,12 @@ func (p *postIndexRequest) UnmarshalJSON(b []byte) error {
 		return err
 	}
 	// Unmarshal expected values.
-	var _p _postIndexRequest
+	_p := _postIndexRequest{
+		Options: pilosa.IndexOptions{
+			Keys:           false,
+			TrackExistence: true,
+		},
+	}
 	if err := json.Unmarshal(b, &_p); err != nil {
 		return errors.Wrap(err, "unmarshalling expected values")
 	}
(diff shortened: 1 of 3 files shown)


travisturner:
I would expect the default logic to be defined in the `pilosa` package (I get that it may also need to be in the `http` package, but having it confined there is not really enforcing a default). 

I think at the very least the instance of `Index` that is returned by `NewIndex()` should have:
```trackExistence: true,```

travisturner:
I don't love the potential double negative that comes with renaming this option to `DisableTrackExsistence`, but in this case maybe that's the right thing to do since having the default be `false` is cleaner.

Anyone have any opinions on that?

jaffee:
I think we should prioritize making the option read more intuitively at the cost of some implementation complexity for us. That is to say, I prefer trackExistence over disableExistenceTracking. 

Most of our existing boolean options already follow this pattern `verbose`, `coordinator`, `diagnotics`, `keys`. The only one I see that doesn't is `tls-skip-verify`.

yuce:
@travisturner I've pushed a commit that sets `trackExistence `in `NewIndex`.

PR #1674, opened 2018-10-01
converted to pilosa.logger

### LINKED ISSUE #1644: expunge all direct calls to log
### Description

As an example, some code in TranslateFile (translate.go) calls log.Printf directly rather than going through a pilosa.Logger. This results in a lot of output when (e.g.) running `go test`, that can't easily be controlled by testing utils. Anything in Pilosa directly calling log should be rewritten to use the `pilosa.Logger` interface with an implementation injected at the time of creation (e.g. when the TranslateFile struct is created).

No more direct `log` calls. No more spurious output when running tests.

tgruben:
## Overview

replace occurrences of raw-logger, gossip's raw still remains for it requires a raw logger

Fixes #1644


holder.go
```diff
@@ -52,7 +52,7 @@ type Holder struct {
 
 	// Key/ID translation
 	translateFile            *TranslateFile
-	NewPrimaryTranslateStore func(interface{}) TranslateStore
+	NewPrimaryTranslateStore func(interface{}) (TranslateStore, error)
```
jaffee: We can't actually make this change since it's an exported field. Luckily, I don't believe we actually need to. See my comment in the implementation below.

http/translator.go
```diff
@@ -43,9 +41,9 @@ func NewTranslateStore(node interface{}) pilosa.TranslateStore {
 	case *pilosa.Node:
 		n = v
 	default:
-		log.Printf("WARNING: a *pilosa.Node is the only type supported by NewTranslateStore().")
```
jaffee: I don't think that this particular log is too problematic - it isn't polluting the output of Pilosa, and will only happen if someone's code is actually wrong. If I'm understanding this correctly it would not actually be unreasonable to panic here, although perhaps we should leave as a log to play it safe.

translate.go
```diff
@@ -90,6 +89,12 @@ func OptTranslateFileMapSize(mapSize int) TranslateFileOption {
 		return nil
 	}
 }
+func OptTranslateFileLogger(l Logger) TranslateFileOption {
```
jaffee: This doesn't seem to actually be used?

http/translator.go
tgruben: Panic would  be more appropriate in reguards to maintaing the commitment to the released api, which philosophically i question in this scenerio, if we can bypass a panic situation with an alteration of an api, i feel its preferable to make the modification.  

Printing a log message returning a nil node in the case of an errant url would result in program failure as well, so no real difference in net result ie messge and program immediate termination.

translate.go
jaffee: right, but presumably we would want to pass the logger in that's used by the rest of the Pilosa components.

server.go
```diff
@@ -236,7 +237,9 @@ func OptServerClusterHasher(h Hasher) ServerOption {
 
 func OptServerTranslateFileMapSize(mapSize int) ServerOption {
 	return func(s *Server) error {
-		s.holder.translateFile = NewTranslateFile(OptTranslateFileMapSize(mapSize))
+		s.holder.translateFile = NewTranslateFile(
+			OptTranslateFileMapSize(mapSize),
+			OptTranslateFileLogger(s.logger))
```
jaffee: this works, but it creates a dependency on the order in which server options are passed to NewServer. It would probably be better to save the map size here and call NewTranslateFile in NewServer after the options have been applied to the server.

tgruben: I guess the success criteria of this ticket is faulty :)

server.go
```diff
@@ -236,7 +237,7 @@ func OptServerClusterHasher(h Hasher) ServerOption {
 
 func OptServerTranslateFileMapSize(mapSize int) ServerOption {
 	return func(s *Server) error {
-		s.holder.translateFile = NewTranslateFile(OptTranslateFileMapSize(mapSize))
+		s.holder.translateFile = NewTranslateFile( OptTranslateFileMapSize(mapSize))
```
jaffee: gofmt

PR #1677, opened 2018-10-03
Fixes #1632

### LINKED ISSUE #1632: columnAttrs logic should be in executor, not API
### Description

currently, `API.Query` passes all pql Calls in a request to the executor, which returns the results of those calls, then (if columnAttrs are requested) API.Query iterates through the results them to build up a slice of all unique column ids in all the results. Then it creates the `[]*ColumnAttrSets` which contains all the column ids and their attributes.

This is problematic for a few reasons - after #700, a client can request columnAttrs per pql Call rather than per request, but we still get all the column attrs for all the Calls in the request. Second, it's conceivable that in the future we wouldn't want to keep the attrs for all columns on every node in the cluster as we do now. We might want to only keep attrs for columns in shards which a given node is responsible for. If we made this change, we would need the executor's internal queries to the other nodes in the cluster to return their column attrs. Third, its possible that some efficiency can be gained by building up the `[]*ColumnAttrSets` incrementally as Calls are processed, rather than all at once at the end.

`executor.Execute` should return a `QueryResponse`, and API.Query shouldn't have to do much more than call `executor.Execute`.

yuce:
## Overview

Fixes #1632 

## Pull request checklist

## Code review checklist

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -118,17 +118,6 @@ func (e *executor) Execute(ctx context.Context, index string, q *pql.Query, shar
 
 	resp.Results = results
 
-	// Translate response objects from ids to keys, if necessary.
-	// No need to translate a remote call.
-	if !opt.Remote {
-		for i := range results {
-			results[i], err = e.translateResult(index, idx, q.Calls[i], results[i])
-			if err != nil {
-				return resp, err
-			}
-		}
-	}
-
 	// Fill column attributes if requested.
 	if opt.ColumnAttrs {
 		// Consolidate all column ids across all calls.
@@ -148,8 +137,8 @@ func (e *executor) Execute(ctx context.Context, index string, q *pql.Query, shar
 		}
 
 		// Translate column attributes, if necessary.
-		if e.Holder.translateFile != nil {
-			for _, col := range resp.ColumnAttrSets {
+		if idx.Keys() {
+			for _, col := range columnAttrSets {
 				v, err := e.Holder.translateFile.TranslateColumnToString(index, col.ID)
 				if err != nil {
 					return resp, err
@@ -159,6 +148,17 @@ func (e *executor) Execute(ctx context.Context, index string, q *pql.Query, shar
 		}
 
 		resp.ColumnAttrSets = columnAttrSets
+	}
+
+	// Translate response objects from ids to keys, if necessary.
+	// No need to translate a remote call.
+	if !opt.Remote {
+		for i := range results {
+			results[i], err = e.translateResult(index, idx, q.Calls[i], results[i])
+			if err != nil {
+				return resp, err
+			}
+		}
 	}
 
 	return resp, nil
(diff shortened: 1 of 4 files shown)


executor.go
```diff
@@ -105,27 +106,85 @@ func (e *executor) Execute(ctx context.Context, index string, q *pql.Query, shar
...
+
+		// Retrieve column attributes across all calls.
+		columnAttrSets, err := e.readColumnAttrSets(e.Holder.Index(index), columnIDs)
+		if err != nil {
+			return resp, errors.Wrap(err, "reading column attrs")
+		}
+
+		// Translate column attributes, if necessary.
+		if e.Holder.translateFile != nil {
+			for _, col := range resp.ColumnAttrSets {
```
travisturner: `resp.ColumnAttrSets` needs to be `columnAttrSets`

executor.go
```diff
@@ -105,27 +106,85 @@ func (e *executor) Execute(ctx context.Context, index string, q *pql.Query, shar
...
+		}
+
+		// Retrieve column attributes across all calls.
+		columnAttrSets, err := e.readColumnAttrSets(e.Holder.Index(index), columnIDs)
+		if err != nil {
+			return resp, errors.Wrap(err, "reading column attrs")
+		}
+
+		// Translate column attributes, if necessary.
+		if e.Holder.translateFile != nil {
```
travisturner: this line should be
```
if idx.Keys() {
```

executor.go
```diff
@@ -105,27 +106,85 @@ func (e *executor) Execute(ctx context.Context, index string, q *pql.Query, shar
...
 			if err != nil {
-				return nil, err
+				return resp, err
 			}
 		}
 	}
-	return results, nil
+
+	// Fill column attributes if requested.
+	if opt.ColumnAttrs {
```
travisturner: This entire code block needs to be moved before the previous block that is calling `translateResult()`.
This is because in the case where column keys are used, `bm.Columns()` is empty if the translation has already happened.

executor.go
yuce: Wouldn't that break compatibility with Pilosa 1.0 ?

travisturner: I mean that this `range` loop is using the wrong argument.
`resp.ColumnAttrSets` is empty, but `columnAttrSets` was populated in the code block above.
The `range` needs to loop over `columnAttrSets`.

yuce:
Updated

executor.go
```diff
@@ -1706,9 +1765,16 @@ func (e *executor) mapperLocal(ctx context.Context, shards []uint64, mapFn mapFu
...
+var translateCallCol = map[string]struct{}{
+	"Set":            {},
+	"Clear":          {},
+	"Row":            {},
+	"SetColumnAttrs": {},
+}
+
 func (e *executor) translateCall(index string, idx *Index, c *pql.Call) error {
 	var colKey, rowKey, fieldName string
-	if c.Name == "Set" || c.Name == "Clear" || c.Name == "Row" {
```
travisturner: I definitely like the idea of cleaning this up, but let's make it a switch statement instead of a map lookup:
```
switch c.Name {
case: "Set", "Clear", "Row", "SetColumnAttrs":
    ...
case: "SetRowAttrs":
    ...
default:
    ...
}
```

travisturner: @yuce I just merged #1679 which actually implements this `switch`, so you can just rebase master for the update, and add in `"SetColumnAttrs"`.

yuce: Updated

pilosa.go
```diff
@@ -117,7 +117,7 @@ var nameRegexp = regexp.MustCompile(`^[a-z][a-z0-9_-]{0,63}$`)
 // ColumnAttrSet represents a set of attributes for a vertical column in an index.
 // Can have a set of attributes attached to it.
 type ColumnAttrSet struct {
-	ID    uint64                 `json:"id"`
```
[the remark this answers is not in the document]
yuce: Why do we try to remove `ID` from the result in the first place ?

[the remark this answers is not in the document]
travisturner: Well, we omit the key when it doesn't apply, so it seems to make sense to omit the ID when it doesn't apply. And when there is a key, the ID is always `0`, so it's not really useful anyway. I don't have a strong opinion on this, however. I'll make a ticket suggesting we revisit that output and it can be discussed there.

PR #1684, opened 2018-10-09
Add distributed tracing.

benbjohnson:
## Overview

This pull request adds distributed tracing throughout the request paths in `API`, `Executor`, `Holder`, `Fragment`, and the `InternalClient`. It also implements HTTP header injection to pass tracing identifiers between nodes.

## Usage

The tracing uses the production settings suggested by Jaeger by default. Two additional flags have been added to allow control over the sampler:

```
--tracing.sampler-type
--tracing.sampler-param
```

For testing/dev, you can have all spans reported by using these settings:

```sh
pilosa server --tracing.sampler-type const --tracing.sampler-param 1
```

## Pull request checklist

## Code review checklist

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -3,141 +3,192 @@
 
 [[projects]]
   branch = "master"
+  digest = "1:7deffcad61694d0cf57178c7d0c41df6b1357612e215914e35cf8b6353abf65c"
   name = "github.com/CAFxX/gcnotifier"
   packages = ["."]
+  pruneopts = ""
   revision = "39b0596a2da3c92787b3319c6b5425a474b4e0da"
 
 [[projects]]
   branch = "master"
+  digest = "1:7519fc1e7fa9cde38634cfaf8a2ba842ea20aab2c14d81d6122ab8d6b1658d1c"
   name = "github.com/DataDog/datadog-go"
   packages = ["statsd"]
+  pruneopts = ""
   revision = "ef3a9daf849df2d7ee3bbf13808dfb481069a773"
 
 [[projects]]
+  digest = "1:f82b8ac36058904227087141017bb82f4b0fc58272990a4cdae3e2d6d222644e"
   name = "github.com/StackExchange/wmi"
   packages = ["."]
+  pruneopts = ""
   revision = "5d049714c4a64225c3c79a7cf7d02f7fb5b96338"
   version = "1.0.0"
 
 [[projects]]
   branch = "master"
+  digest = "1:354e62d5acb9af138e13ec842f78a846d214a8d4a9f80e578698f1f1565e2ef8"
   name = "github.com/armon/go-metrics"
   packages = ["."]
+  pruneopts = ""
   revision = "3c58d8115a78a6879e5df75ae900846768d36895"
 
 [[projects]]
+  digest = "1:ed112122ed4a920d944cc99b9d00b0441c11685939c28462c719488d36fe29aa"
   name = "github.com/boltdb/bolt"
   packages = ["."]
+  pruneopts = ""
   revision = "2f1ce7a837dcb8da3ec595b1dac9d0632f0f99e8"
   version = "v1.3.1"
 
 [[projects]]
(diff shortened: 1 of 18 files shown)


codysoyland:
Very nice start. I did some testing and was able to get traces on my local Jaeger instance, however none of the distributed queries on a local cluster seem to have traces from "remote" nodes. For example, `InternalClient.QueryNode` seems to be the lowest level I see in my traces, but I don't see anything from the remote node. I see a lot of orphaned traces that I think are from cases where `context.Background()` is passed in. Could this be something to do with the sample rate? Only a few of the requests I sent in get traces created, and I am not sure how to configure that.

Also it seems that the Jaeger environment config doesn't work (JAEGER_AGENT_HOST, etc). Is your approach to just add Pilosa options to cover these options on an as-needed basis? We will probably at least want to expose the agent/collector service addresses as config options.

Also, would be good to get some data logged out of the queries, for example, in HandlePostQuery (from my POC):

```
span.LogFields(
               log.String("query", req.Query),
               log.String("index", req.Index),
       )
```

(Not sure the difference between span.LogKV and span.LogFields as it pertains to Jaeger.)

FYI, my environment is just `pilosa server` (no options) running alongside Jaeger using Docker: ```docker run -d --name jaeger   -e COLLECTOR_ZIPKIN_HTTP_PORT=9411   -p 5775:5775/udp   -p 6831:6831/udp   -p 6832:6832/udp   -p 5778:5778   -p 16686:16686   -p 14268:14268   -p 9411:9411   jaegertracing/all-in-one:1.7```

benbjohnson:
> _none of the distributed queries on a local cluster seem to have traces from "remote" nodes_

Sorry about that. I think I fixed the issue but I'm having issues with getting a cluster running that doesn't fully replicate. I have `cluster.replicas = 1` so I'm not sure what I'm doing wrong. Can you give the newest commit a try on your cluster? (6f6dbef)

> _Could this be something to do with the sample rate? Only a few of the requests I sent in get traces created, and I am not sure how to configure that._

You can use a `const` tracer with a tracing rate of `1` to see all calls traced:

```
--tracing.sampler-type const --tracing.sampler-param 1
```

> _Is your approach to just add Pilosa options to cover these options on an as-needed basis?_

Yes, it seemed messy to have users mix `JAEGER` environment variables with Pilosa configuration—especially since we're abstracting the tracing a bit now. I can add the agent address to the config.

> _Also, would be good to get some data logged out of the queries_

One thing I was worried about was logging it if users sent a large `SetBit()` calls in one query. Should I restrict the size of the query string being logged?

codysoyland:
Hey Ben, thanks for the reply. I'm still having no luck getting the remote trace to join the local one, but I made a little progress debugging. I'm mostly testing with Set() queries so the replication ensures that there will always be a remote trace.

I agree that having our own config flags for the various Jaeger options makes sense, and we will definitely need to have one to specify the agent address.

I'm not really sure about how query logging will affect performance, so we can leave it out for now, but it did make me start thinking about how we can cross-reference queries to traces. Maybe we can get the jaeger trace ID into the pilosa response headers? Something to think about.

There were a couple of problems I found with the injecting/extracting logic:

1. The `InternalClient.do` method wasn't called from anywhere, as the individual methods call out to ` c.httpClient.Do(`.
2. I think the reference to `opentracing.GlobalTracer()` is wrong because Jaeger has its own tracer implementation.

These are both addressed in https://github.com/codysoyland/pilosa/commit/c472b293303a4c4668659d579efc7b8bc72766dd

After applying that patch, I now see `Uber-Trace-Id` headers in the inter-node http requests (via ngrep), but Jaeger is still not displaying them together. Here is a screenshot of what I'm seeing:
2018-10-23 at 1 41 pm

Should I expect to see the traces from the remote nodes nested under the QueryNode calls in this trace? (I do see traces from the remote nodes in two separate traces.)

benbjohnson:
@codysoyland OK, it took way too long to figure out why the distributed tracing wasn't propagating but I finally got it working. I also added a config parameter for setting the agent host/port.

benbjohnson:
Thanks @codysoyland. I hard reset to your branch and pushed up. Can you approve again when you get a chance and I'll merge it once CI finishes.

PR #1695, opened 2018-10-17
Available shard persistance bug

### LINKED ISSUE #1694: Problem with availableShard handling
### Expected behavior
Restart pilosa node  produces running pilosa node
### Actual behavior
Node stops restart fails with issues with the handling of the available shard file

Load pdk usecase taxi
stop and start pilosa

tgruben:
## Overview

The .available.shard file was getting corrupted anytime the new availableShards bitmap was smaller than the previous state
Fixes #1694


field_internal_test.go
```diff
@@ -361,3 +361,42 @@ func TestField_PersistAvailableShards(t *testing.T) {
 	}
 
 }
+
+func TestField_PersistAvailableShardsTaxiBug(t *testing.T) {
```
travisturner:
```suggestion
// Ensure that persisting available shards having a smaller footprint (for example,
// when going from a bitmap to a smaller, RLE representation) succeeds.
func TestField_PersistAvailableShardsFootprint(t *testing.T) {
```

PR #1699, opened 2018-10-22
add `clear` functional option for imports

### LINKED ISSUE #1657: Implement Inverse Imports
### Description

We currently only support importing data into Pilosa to mass `Set` bits. It should be easy to support mass `Clear`ing bits using the same import mechanisms.

* Add `inverse` parameter to `API.Import` and `API.ImportRoaring` functions. When set to `true`, `Clear` bits instead of `Set`ting them.
* Add `inverse` query argument to `/import` and `/import-roaring` HTTP endpoints.
* Update client libraries to support inverse imports.

travisturner:
## Overview

This PR adds functional options to all exposed Import-related api methods. Providing `OptImportOptionsClear(true)` to an `Import()` method will treat the payload as clear-bits.

Fixes #1657 

TODO:
- [x] We don't currently support any type of `clearValue` for `int` fields. The `setValue` import requires a `column`, `value`, but for `clearValue` it doesn't really make sense to require the value for clears. In that case you just need the `column`. Is that a reasonable way to handle it?
- [x] documentation
- [x] update `ImportRoaring` to support clear

## Pull request checklist

## Code review checklist

### DIFF
--- a/api.go
+++ b/api.go
@@ -757,9 +757,18 @@ func (api *API) Import(_ context.Context, req *ImportRequest, opts ...ImportOpti
 }
 
 // ImportValue bulk imports values into a particular field.
-func (api *API) ImportValue(_ context.Context, req *ImportValueRequest) error {
+func (api *API) ImportValue(_ context.Context, req *ImportValueRequest, opts ...ImportOption) error {
 	if err := api.validate(apiImportValue); err != nil {
 		return errors.Wrap(err, "validating api method")
+	}
+
+	// Set up import options.
+	options := &ImportOptions{}
+	for _, opt := range opts {
+		err := opt(options)
+		if err != nil {
+			return errors.Wrap(err, "applying option")
+		}
 	}
 
 	index := api.holder.Index(req.Index)
@@ -783,13 +792,15 @@ func (api *API) ImportValue(_ context.Context, req *ImportValueRequest) error {
 	}
 
 	// Import columnIDs into existence field.
-	if err := importExistenceColumns(index, req.ColumnIDs); err != nil {
-		api.server.logger.Printf("import existence error: index=%s, field=%s, shard=%d, columns=%d, err=%s", req.Index, req.Field, req.Shard, len(req.ColumnIDs), err)
-		return errors.Wrap(err, "importing existence columns")
+	if !options.Clear {
+		if err := importExistenceColumns(index, req.ColumnIDs); err != nil {
+			api.server.logger.Printf("import existence error: index=%s, field=%s, shard=%d, columns=%d, err=%s", req.Index, req.Field, req.Shard, len(req.ColumnIDs), err)
+			return errors.Wrap(err, "importing existence columns")
+		}
 	}
 
 	// Import into fragment.
-	err = field.importValue(req.ColumnIDs, req.Values)
+	err = field.importValue(req.ColumnIDs, req.Values, options)
 	if err != nil {
 		api.server.logger.Printf("import error: index=%s, field=%s, shard=%d, columns=%d, err=%s", req.Index, req.Field, req.Shard, len(req.ColumnIDs), err)
 	}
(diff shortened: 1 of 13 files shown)


http/client_test.go
```diff
@@ -638,6 +654,37 @@ func TestClient_ImportKeys(t *testing.T) {
...
+		// Verify Sum.
+		sum, cnt, err = field.Sum(nil, fldName)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if sum != 30 || cnt != 2 {
+			t.Fatalf("unexpected values: got sum=%v, count=%v; expected sum=30, cnt=2", sum, cnt)
+		}
+
+		// Verify Range
```
yuce: Dot at the end of the comment is missing.

http/client.go
```diff
@@ -537,7 +579,19 @@ func (c *InternalClient) ImportRoaring(ctx context.Context, uri *pilosa.URI, ind
...
+	for _, opt := range opts {
+		err := opt(options)
+		if err != nil {
+			return errors.Wrap(err, "applying option")
+		}
+	}
+
 	url := fmt.Sprintf("%s/index/%s/field/%s/import-roaring/%d?remote=%v", uri, index, field, shard, remote)
+	if options.Clear {
+		url += "&clear=true"
```
yuce: I think it's better to use `url.Values` to create the query part of the URL instead of string concatenation. Something like:
```go
url := fmt.Sprintf("%s/index/%s/field/%s/import-roaring/%d", uri, index, field, shard)
vs := url.Values{}
vs.Set("remote", remote.String())
if options.Clear{
    vs.Set("clear", "true")
}
url = fmt.Sprintf("%s?%s", url, vs.Encode())
```

fragment.go
```diff
@@ -688,12 +712,11 @@ func (f *fragment) importSetValue(columnID uint64, bitDepth uint, value uint64)
 // A bitmap can be passed in to optionally filter the computed columns.
 func (f *fragment) sum(filter *Row, bitDepth uint) (sum, count uint64, err error) {
 	// Compute count based on the existence row.
-	row := f.row(uint64(bitDepth))
+	consider := f.row(uint64(bitDepth))
 	if filter != nil {
-		count = row.intersectionCount(filter)
-	} else {
-		count = row.Count()
+		consider = consider.Intersect(filter)
```
alanbernstein: Isn't `intersectionCount` faster than separate `Intersect` and `Count`?

travisturner: Yes, but this change is necessary because of a bug with this `sum()` method. It wasn't taking the not-null bitmap (`row(bitDepth)`) into consideration when calculating sum. So if not-null was set to `0`, but the BSI rows still contained a non-zero value, that value would be included in the sum. Basically, we need `consider` later in the method. 

With all that said, there may be a way to leverage `instersectCount()` that I'm missing. Can you suggest a different approach?

api.go
```diff
@@ -256,10 +256,20 @@ func (api *API) Field(_ context.Context, indexName, fieldName string) (*Field, e
...
 // of the rows in this shard of this field concatenated together in one long
 // bitmap.
-func (api *API) ImportRoaring(ctx context.Context, indexName, fieldName string, shard uint64, remote bool, data []byte) (err error) {
+func (api *API) ImportRoaring(ctx context.Context, indexName, fieldName string, shard uint64, remote bool, data []byte, opts ...ImportOption) (err error) {
 	if err = api.validate(apiField); err != nil {
 		return errors.Wrap(err, "validating api method")
 	}
+
+	// Set up import options.
+	options := &ImportOptions{}
```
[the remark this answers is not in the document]
travisturner: I added a function to consolidate these. I can't share the function with `http/client.go` without exporting it (because it's a separate package). I don't think it's worth exporting for that one instance.

travisturner:
I implemented it at that API. Looks like it needs to be added to the http handler.

fragment.go
alanbernstein: No, that makes sense.

PR #1701, opened 2018-10-23
Proxy import

### LINKED ISSUE #1589: /import endpoint should forward to the correct node (which owns the shard)

tgruben:
## Overview

Allows an import request to be performed behind a proxy server.  This is achieved by adding additional http query arguments (either "shard" for id data or "keyImport" for string key based imports) needed to route the request to the appropriate destination node without requiring parsing the POSTed data.  

Fixes #1589


travisturner:
@tgruben gofmt

http/client.go
```diff
@@ -427,6 +427,44 @@ func (c *InternalClient) marshalImportPayload(index, field string, shard uint64,
 	return buf, nil
 }
 
+// Forward the request to the new host in request
+func (c *InternalClient) Forward(ctx context.Context, w http.ResponseWriter, req *http.Request, scheme, host string) error {
+
+	body, err := ioutil.ReadAll(req.Body) // it appears I must consume the
```
jaffee: "it appears I must consume the"...?

http/handler.go
```diff
@@ -1013,6 +1013,11 @@ func (h *Handler) handlePostImport(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	//if not the execution destination forward on to correct for execution
```
jaffee: I think you accidentally a word.

api.go
```diff
@@ -983,6 +984,60 @@ func (api *API) Info() serverInfo {
 	}
 }
 
+// EnsureOperatingNode determine if the current node is the destination for execution if not forward on to correct node
+func (api *API) EnsureOperatingNode(indexName string, w http.ResponseWriter, r *http.Request) bool {
```
jaffee: the comment on this method needs to explain what the return values mean. "true" meaning that the query did not belong to this node was unintuitive for me. I don't care which value means what though, I just want it documented.

api.go
```diff
@@ -983,6 +984,60 @@ func (api *API) Info() serverInfo {
...
+			api.forwardTo(w, r, forwardNode)
+			return true
+		}
+
+	}
+	return false
+}
+
+func (api *API) ownsShard(indexName string, shard uint64) (node *Node, owner bool) {
+	nodes := api.cluster.shardNodes(indexName, shard)
```
jaffee: cluster has an `ownsShard` method

api.go
```diff
@@ -983,6 +984,60 @@ func (api *API) Info() serverInfo {
...
+		node = nodes[i]
+	}
+	return
+}
+
+func (api *API) forwardTo(w http.ResponseWriter, r *http.Request, node *Node) {
+	api.server.defaultClient.Forward(r.Context(), w, r, node.URI.Scheme, node.URI.HostPort())
+}
+
+func (api *API) getCoordinator() *Node {
```
jaffee: cluster has a `coordinatorNode` method.

api.go
```diff
@@ -983,6 +984,60 @@ func (api *API) Info() serverInfo {
...
+		// if not coordinator..forward on
+		if !api.Node().IsCoordinator {
+			api.forwardTo(w, r, api.getCoordinator())
+			return true
+		}
+		return false
+	}
+
+	if shardS, hasShard := values["shard"]; hasShard {
+		shard, err := strconv.ParseUint(shardS[0], 10, 64)
```
jaffee: I take it shardS is a []string that's guaranteed to always be of length 1?

jaffee: is it possible that it might exist, but have length 0?

http/client.go
```diff
@@ -427,6 +427,44 @@ func (c *InternalClient) marshalImportPayload(index, field string, shard uint64,
...
+	forReq, err := http.NewRequest(req.Method, url, bytes.NewReader(body))
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return nil
+	}
+
+	forReq.Header = make(http.Header)
+	for h, val := range req.Header {
+		forReq.Header[h] = val
+	}
```
jaffee: I think it's possible to forward the original request without creating a copy (maybe without reading from body?)

In the PDK proxy I used the same object, but had to do a few things:
```
	orig.URL = reqURL // set to new URL
	orig.Host = ""    // can/should be empty for client requests
	orig.RequestURI = "" // should not be set in a client request
	orig.ContentLength = int64(len(origbody)) // can probably leave this alone if body is untouched.
```
I had to read the body in the PDK because I was modifying it, but it'd be cool if we could avoid it.

http/client.go
```diff
@@ -427,6 +427,44 @@ func (c *InternalClient) marshalImportPayload(index, field string, shard uint64,
...
+
+	resp, err := c.httpClient.Do(forReq.WithContext(ctx))
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadGateway)
+		return nil
+	}
+	if resp.StatusCode != http.StatusOK {
+		http.Error(w, "Forward Failed", resp.StatusCode)
```
jaffee: I think we should maintain the same body from the original response instead of replacing with "Forward Failed". Maybe we could prepend "forward failed: " in the spirit of Go error messages adding context to the error.

jaffee:
I'm not opposed to the strategy taken here (a generic forwarding method), but I'm wondering why you decided to depart from the way that we did things in ImportRoaring (calling client.ImportRoaring).

api.go
tgruben: just an artifact of using the  go stdlib URL.Query api.  The only correct way to call this method is with shard=SHARDNUM and anything other than that would be an error.  So just error handling needs to be tightened up

jaffee:
@tgruben 	ping before this gets too out of date.

travisturner:
@jaffee to answer your last question here, I believe the difference between this implementation and `ImportRoaring` is that this one attempts to forward requests without having to unmarshal the payload in order to determine the destination. (I could be wrong; I'm still going through it).

travisturner:
After reviewing this PR, I don't think it's going to work as is because it's not taking replicas into account. Currently, we are relying on the clients to send import data to each replica; the clients ask the cluster—or in the case of the `pilosa import` command, the host specified by the `--host` config argument—for the list of nodes responsible for a particular fragment, and then sends the import data to each node in that list.

This PR forwards the import request to a single replica (the last node in the list of replicas), so as it is the other replicas would be missing data and would rely on anti-entropy to clean them up (which may actually delete the data if replicas is more than 2).

I think we're going to have to modify this so that it forwards the request to all replicas. I still need to think through how that logic works with the clients, because we'll need to know how to distinguish:
- not owned by current node, forward to all replicas
- owned by current node, write local
- owned by current node, write local AND forward to remaining replicas

travisturner:
I've done a lot of work refactoring this at the api level (so that api can forward import requests when necessary). That work is in a WIP commit on this branch: https://github.com/travisturner/pilosa/commits/proxy-import
I feel ok about that piece of it, and it seems to be an improvement over the existing api logic.

I started to look at the import forwarding at the handler (http) level (implemented in this PR) and realized that it introduces http into the pilosa api. To date, the http package is just a `Handler` implementation on `Server`, but this changes that. So I'm trying to think through how the http forwarding can interact with api without them being so dependent on each other.

travisturner:
I pushed another commit to that branch which removes the `net/http` dependency from api. I'm going to start working on tests for the http forwarding, then I'll push everything to this PR.

travisturner:
@tgruben @jaffee I just pushed a commit to this PR that makes some pretty significant changes. It also modifies the `InternalClient` api, so it would need to be part of a 2.0 release if we went this route. At this point I'm just looking for feedback and discussion around the approach, and from there we can decide if it's worth going forward with it.

travisturner:
@jaffee this PR does not include `ImportRoaring`, although I don't see why it couldn't apply the same logic to that method. The primary goal of this PR is to introduce request forwarding prior to the unmarshaling of payload, so as I understand it, it makes no difference if the import is csv or roaring; the goal is to get the request to the right node(s).

With that said, this PR breaks the API, and it's rather complicated given the two levels of (possible) destination check. So I'm not married to this solution, rather it was one approach looking for feedback.

jaffee:
since this is breaking (2.0 probably required), out of date, and not a top priority, I'm going to close this PR. The ticket is iceboxed, but we should resurrect it when going to 2.0

PR #1707, opened 2018-10-27
Add base system, curl and jq for debug and checks.

### LINKED ISSUE #1706: Docker image lacks of basic tools
Small image is very cool but not when it lacks minimal debug/check tooling. For e.g. in Kubernetes it would avoid having  excessive sidecar and init containers.
Also we need curl and jq to poll for status and other API objects pilosa exposes.

dene14:
## Overview

Small image is very cool but not when it lacks minimal debug/check tooling. For e.g. in Kubernetes it would avoid having  excessive sidecar and init containers.

Fixes #1706

## Pull request checklist

- [o] I have updated the [documentation](https://github.com/pilosa/pilosa/tree/master/docs).
- [o] I have included tests that cover my changes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,9 +5,11 @@ COPY . /go/src/github.com/pilosa/pilosa/
 RUN cd /go/src/github.com/pilosa/pilosa \
     && CGO_ENABLED=0 make install-dep install FLAGS="-a"
 
-FROM scratch
+FROM alpine:3.8
 
 LABEL maintainer "dev@pilosa.com"
+
+RUN apk add --no-cache curl jq
 
 COPY --from=builder /go/bin/pilosa /pilosa
 


jaffee:
@codysoyland 	any thoughts on this? I don't have a problem with it, but maybe it'd be better to publish under a separate name - pilosa-extra, pilosa-debug or something like that?

dene14:
> I would prefer we not use a separate name; it will be simpler to continue only releasing one build for each version release.

Quite agree. There is no reason to create a separate image. We should be able to attach into container and do some plain stuff in there anyway.

yuce:
It's quite trivial to create and use custom Pilosa docker images, and everyone's needs can be different. I think we shouldn't change the current docker image. If we must, I agree with @jaffee that it's better to name it differently (and keep the minimal image). Also I think it would be better to base the extra image on Ubuntu (which I think is 40-50 MB).

jaffee:
hey @yuce I didn't see your comment before I merged this - I'm playing around with docker-compose right now for cluster testing stuff and may need to make some changes anyway, so we can keep discussing this, but based on my discussions with @codysoyland I think that the few extra megs we've added here is probably worth it.

codysoyland:
@yuce I agree in principle that this won't solve everybody's use case, but I think this is a good compromise between minimalism and practicality. The image grows only about 25% bigger, but includes "batteries" such as a package manager that lets people install anything. I was very tempted to reject this on minimalistic purism, but I convinced myself that is not pragmatic. 

I believe Alpine is the most popular base image and is <5MB. We still have time to change this before the next release, so I would welcome your argument in favor of using Ubuntu as a base image, but, my preference for Ubuntu aside, I believe Alpine is the right call.

Also, considering the suggestion to create separately named images with different base images, or advising our users to create their own custom images, consider the difficulty in changing out the image used in a production environment, as you would have to shut down pilosa and remount the data volume on a different container with a different base image. Also not all users have time to create and manage a custom registry. I think it will be much simpler if there is only one official image.

yuce:
@codysoyland @jaffee I've suggested Ubuntu since Ubuntu is more mainstream, and their images are not that bigger from Alpine's. I assume they would have better security since they have much more resources and their business depends on that.

I agree that changing docker image names on production wouldn't make much sense, but I would assume people who would like to use extra tools in the docker image would use `pilosa-extra` exclusively.

codysoyland:
Ubuntu is quite a bit bigger:
```
ubuntu:latest: 85.8MB
alpine:latest: 4.41MB
```

yuce:
I think that's the size of the image after expanded. The download size was between 30-40MB for me.

PR #1712, opened 2018-11-01
Ensure btree comparison doesn't fail for smallish N

seebs:
## Overview

 enterprise/b: Ensure 64-bit range for btree keys.

The cmp() function used to compare btree keys was using the
trick of comparing unsigned values by coercing the result
of subtraction to a signed type. This is fine as long as
the range of differences never actually exceeds the limits
of the signed type. For instance, with int64, as long as
the magnitude of the difference is under 2^63 or so, it
works reasonably well.

Plain int, however, can be a 32-bit type, at which point
the magnitude of difference needed to break it is only
2^31 or so.

Subtraction and type conversion is enough cheaper than
branches that this is worth preserving, but it's worth
preserving by switching to an explicit int64 for the
operation and return type. This will not actually affect
performance except for people using the btree code on
32-bit machines, so it probably won't ever matter.
Which may also be true of the potential wrong answers,
but "might be slower" is a better risk than "might
crash".

## Pull request checklist

Haven't written a test for this. Possibly I should; this is the kind of thing that'd be worth having test cases for, perhaps.

Not sure whether this change is substantive enough to need to be in the docs.

## Code review checklist

### DIFF
--- a/enterprise/b/btree.go
+++ b/enterprise/b/btree.go
@@ -75,7 +75,7 @@ type (
 	//	  0 if a == b
 	//	> 0 if a >  b
 	//
-	Cmp func(a, b uint64) int
+	Cmp func(a, b uint64) int64
 
 	d struct { // data page
 		c int
--- a/enterprise/b/containers_btree.go
+++ b/enterprise/b/containers_btree.go
@@ -23,8 +23,8 @@ import (
 	"github.com/pilosa/pilosa/roaring"
 )
 
-func cmp(a, b uint64) int {
-	return int(a - b)
+func cmp(a, b uint64) int64 {
+	return int64(a - b)
 }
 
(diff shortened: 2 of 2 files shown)


travisturner:
@seebs can you make this two separate PRs? I can review the b+tree portion, but I'd like @codysoyland to look at the handler changes.

codysoyland:
FYI, the handler changes are duplicated from #1711, so after that is merged, this PR will only contain the btree changes. I would just suggest creating a more descriptive title and description.

seebs:
I... totally thought I had successfully used the github UI today. I guess not. Will fix that up.

seebs:
did i seriously just paste in the header from the WRONG commit? why, yes, yes i did. this is me before breakfast and caffeine.

PR #1713, opened 2018-11-01
Allow backslash, carriage return in PQL strings

yuce:
## Overview

Fixes #

## Pull request checklist

## Code review checklist


travisturner:
This change works for me as I expect. @jaffee or @alanbernstein do either of you have any reservations with applying this change to master?

travisturner:
Actually, I just noticed a problem trying this:
```
curl -X POST http://localhost:10101/index/i/query -d 'SetColumnAttrs(1, name="left \"inner\" right")'
```
I'm not sure how to escape double quotes within a double-quoted string.

When I revert to the existing peg file, the above query results in this:
```
{"results":[{"attrs":{},"columns":[1]}],"columnAttrs":[{"id":1,"attrs":{"name":"left \\\"inner\\\" right"}}]}
```
which seems to me like it's over-escaping things.

yuce:
@travisturner I've pushed a fix. All tests pass, even the one that failed earlier.

yuce:
That curl command works with the latest fix:
```
curl -X POST http://localhost:10101/index/i/query -d 'SetColumnAttrs(1, name="left \"inner\" right")'
{"results":[null]}
```

travisturner:
I think what I'm struggling with is determining how/if the parser handles escaping, and I don't think it really does. The changes made so far allow it to parse the PQL without error, but I don't think the results that we get are expected.

For example (payloads provided with `curl -X POST http://localhost:10101/index/i/query -d`:
```
1) 'SetColumnAttrs(1, name="one \"foo\" three")' ---> "attrs":{"name":"one \\\"foo\\\" three"}
2) "SetColumnAttrs(1, name='one "foo" three')"   ---> "attrs":{"name":"one foo three"}
3) "SetColumnAttrs(1, name='one \"foo\" three')" ---> "attrs":{"name":"one \"foo\" three"}
4) 'SetColumnAttrs(1, name="one \'foo\' three")' ---> [isn't a complete bash command]
5) 'SetColumnAttrs(1, name="one 'foo' three")'   ---> "attrs":{"name":"one foo three"}
```

I think some of my confusion here has to do with the combination of bash and PQL parsing.
1) seems to result in the escaped, literal characters of the `\"`
2) this may be an issue with bash parsing, but the double quotes get lost completely here.
3) in a way this seems correct, but it wasn't intuitive
4) this is the inverse (single/double quote) of (3) and it isn't a valid command.
5) similar to (2)

yuce:
The following test for this fix passes just fine, so I don't think the problem is with the parser:
```go
q, err := pql.ParseString(`MyCall( key= value, foo='bar', age = 12 , bool0=true, bool1=false, x=null, escape="\" \\escape\n"  )`)
...
reflect.DeepEqual(q.Calls[0],
			&pql.Call{
				Name: "MyCall",
				Args: map[string]interface{}{
					"key":    "value",
					"foo":    "bar",
					"age":    int64(12),
					"bool0":  true,
					"bool1":  false,
					"x":      nil,
					"escape": `\" \\escape\n`,
				}
```
Checking where else the problem can be.

yuce:
Btw, I've ran the following query with the Go client, and the problem happens with that too:
```go
	attrs := map[string]interface{}{
		"name": "left \"inner\" right",
	}
	_, err = client.Query(i.BatchQuery(
		f.Set(5, 10),
		i.SetColumnAttrs(10, attrs),
	))
```

yuce:
I've added the following code to the Go client test, and I think there's no problem at all:
```go
	resp, err := client.Query(f.Row(5), pilosa.OptQueryColumnAttrs(true))
	if err != nil {
		panic(err)
	}
	fmt.Println(resp.Column())
```
Prints: `{10  map[name:left \"inner\" right]}`

yuce:
I think everything's alright. I've decoded the response coming from curl in python, and this is what I get:
```python
>>> s = '''{"results":[{"attrs":{},"columns":[10]}],"columnAttrs":[{"id":10,"attrs":{"name":"left \\\"inner\\\" right"}}]}'''
>>> json.loads(s)
{u'columnAttrs': [{u'id': 10, u'attrs': {u'name': u'left "inner" right'}}], u'results': [{u'attrs': {}, u'columns': [10]}]}
```

yuce:
Everything is not alright. There's an extra slash being appended. Here's some client code:
```go
	client := pilosa.DefaultClient()
	schema, err := client.Schema()
	if err != nil {
		panic(err)
	}
	i := schema.Index("i")
	f := i.Field("f")
	attrs := map[string]interface{}{
		"name": `left "inner" right`,
	}
	_, err = client.Query(i.BatchQuery(
		f.Set(6, 60),
		i.SetColumnAttrs(60, attrs),
	))
	if err != nil {
		panic(err)
	}
	resp, err := client.Query(f.Row(6), pilosa.OptQueryColumnAttrs(true))
	if err != nil {
		panic(err)
	}
	ci := resp.Column()
	name := ci.Attributes["name"]
	fmt.Println(name, name == `left "inner" right`)
}
```
Checking to see where the problem is...

yuce:
I think I've found the problem. The Go client auto-escapes quotes, so
```
`left "inner" right` becomes `left \"inner\" right`
```
But the server interprets the value of an attribute as a raw string, so sends back
```
left \"inner\" right
```
Although it should have unquoted that string and sent:
```
left "inner" right
```

yuce:
OK, found the fix. It's a bit rough now, will clean it up and push.

yuce:
Pushed the fix. Escape characters were not supported in attribute values before this fix.

yuce:
Output of curl with the latest commit:
```
$ curl -X POST http://localhost:10101/index/i/query?columnAttrs=true -d 'Row(f=6)'
{"results":[{"attrs":{},"columns":[60]}],"columnAttrs":[{"id":60,"attrs":{"name":"left \"inner\" right"}}]}
```

yuce:
I've disabled unescaping for single quoted strings, since `strconv.Unquote` doesn't support that.

Issue #1692, opened 2018-10-12
Cluster can't recover from "STARTING" state after outage of a node during data import 


dmibor:
This happened during some test data load into the 3 node cluster. When cluster is idle haven't seen this issue, nodes join and cluster properly recovers.

### Expected behavior
Node should be able to rejoin the cluster and cluster should get back to work in NORMAL state
### Actual behavior
Cluster stays in STARTING state until failed node explicitly restarted 
### Steps to reproduce the behavior
Dont see easy way to catch that - I would give logs of what happened, maybe they would be useful

By the looks of it story goes like this - node is unresponsive for few seconds while importing some data, it's detected by everyone, node is kicked out, node recovers, everyone receives nodeJoin at 12:22:44 but cluster never gets back to NORMAL state.

Coordinator:
```
2018/10/11 12:22:30 [DEBUG] memberlist: Initiating push/pull sync with: node2:14000
2018/10/11 12:22:32 fragment: snapshot complete test_index/test_field_1/standard/2 took 3.45779668s
2018/10/11 12:22:32 fragment: snapshotting test_index/test_field_1/standard_2018/2
2018/10/11 12:22:34 fragment: snapshot complete test_index/test_field_1/standard_201807/29 took 4.47381738s
2018/10/11 12:22:34 [DEBUG] memberlist: Failed ping: 65ccdd1a-a865-4fd2-90f7-659e0e726b08 (timeout reached)
2018/10/11 12:22:35 fragment: snapshot complete test_index/test_field_1/standard_2018/2 took 2.42606313s
2018/10/11 12:22:35 [ERR] memberlist: Failed fallback ping: read tcp node_coordinator:57226->node2:14000: i/o timeout
2018/10/11 12:22:35 [INFO] memberlist: Suspect 65ccdd1a-a865-4fd2-90f7-659e0e726b08 has failed, no acks received
2018/10/11 12:22:35 fragment: snapshotting test_index/test_field_1/standard_201807/2
2018/10/11 12:22:35 [DEBUG] memberlist: Failed ping: 65ccdd1a-a865-4fd2-90f7-659e0e726b08 (timeout reached)
2018/10/11 12:22:36 fragment: snapshot complete test_index/test_field_1/standard_201807/2 took 831.405793ms
2018/10/11 12:22:37 [ERR] memberlist: Failed fallback ping: read tcp node_coordinator:57228->node2:14000: i/o timeout
[...]

PR #1717, opened 2018-11-02
Cluster tests

jaffee:
## Overview

fixes: #1692 (need to confirm with @dmibor)

Creates a framework for running cluster tests in docker.

We add a docker-compose file which runs a 3-node pilosa cluster and then another node which gets the test code. The test node contains [pumba](https://github.com/alexei-led/pumba) and has the docker socket mounted so that it can control docker from inside its container.

TODO: 
- [x] figure out a way to skip the cluster tests when running the normal test suite.
- [x] add a Makefile and/or instructions for running and re-running the cluster tests. Currently docker-compose up will run the tests, but if one makes changes to the code it isn't always clear what needs to be done to rebuild with those changes. Sometimes it's possible to just rebuild the test container but leave the Pilosa nodes, sometimes everything needs to be rebuilt. `docker-compose down`, `docker-compose up --build`, and `docker-compose build <container>` all have their place.
- [x] figure out how to stop docker-compose once the test is done running. currently the test container stops, but pilosa keeps running.
- we need go-pilosa master (for `Status().State`), but currently there are dependency issues which prevent that. Figure these out.
- [x] fix the bug that the current test is exposing :)

Fixes #

## Pull request checklist

## Code review checklist


jaffee:
This is ready to go now mod the go-pilosa dependency issues which aren't a major problem and we'll probably need to cut a release of something to fix.

Dockerfile-withgo
```diff
@@ -0,0 +1,26 @@
+# This Dockerfile is used for cluster testing - it produces a much larger image
```
codysoyland: Is it possible to move this file into `internal/clustertests`?
It would also be nice to include the main Dockerfile dynamically and add to it, but it seems we can't easily do that without something like [this](https://github.com/moby/moby/issues/4027).

Gopkg.lock
```diff
@@ -189,6 +189,15 @@
   revision = "c01d1270ff3e442a8a57cddc1c92dc1138598194"
   version = "v1.2.0"
 
+[[projects]]
```
codysoyland: Have we determined if this will cause a dependency cycle since go-pilosa depends on pilosa?

codysoyland:
In addition to the comments above, we should talk about adding this to our CircleCI config. I can help with that. We need to configure Circle to allow us to run docker commands: https://circleci.com/docs/2.0/building-docker-images/

Dockerfile-withgo
jaffee: We need to be able to include the current Pilosa code tree when building (for easy testing of changes), hence the COPY command. `COPY ../..` is not allowed because of:

```
Step 3/13 : COPY ../.. /go/src/github.com/pilosa/pilosa/
COPY failed: Forbidden path outside the build context: ../.. ()
```

I agree it would be very nice to be able to base the build off another Dockerfile. 

I don't see a way to include the current Pilosa codebase without having an extra Dockerfile at the top level though since our current image uses a builder and doesn't include the codebase.

Makefile
```diff
@@ -68,6 +68,22 @@ release: check-clean
 	$(MAKE) release-build GOOS=linux GOARCH=386
 	$(MAKE) release-build GOOS=linux GOARCH=386 ENTERPRISE=1
 
+# Run cluster integration tests using docker. Requires docker daemon to be
+# running. This will catch changes to internal/clustertests/*.go, but if you
+# make changes to Pilosa, you'll want to run clustertests-build to rebuild the
+# pilosa image.
+clustertests:
+	cd internal/clustertests;\
```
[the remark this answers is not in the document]
jaffee: good call on the first two comments. Running the tests from outside docker isn't an option right now due to networking shenanigans. It's also not desirable as that puts the client outside of pumba's control and we might want to test network partitions and such against the client.

It's my hope that we can add a timeout to CI to guard against the (hopefully infrequent) deadlock scenario?

internal/clustertests/cluster_test.go
```diff
@@ -0,0 +1,152 @@
...
+		r, err = cli.Query(idx.Count(f.Row(0)))
+		if err != nil {
+			t.Fatalf("count querying: %v", err)
+		} else if r.Result().Count() != 1000 {
+			t.Fatalf("count after import is %d", r.Result().Count())
+		}
+
+	})
+
+	down := exec.Command("/pumba", "stop", "clustertests_pilosa3_1", "clustertests_pilosa2_1", "clustertests_pilosa1_1")
```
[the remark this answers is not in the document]
jaffee: using the exit code option obviates this entirely

Makefile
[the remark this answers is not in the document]
codysoyland: Gotcha, and yes, CircleCI has a configurable timeout that defaults to 10 minutes.

Dockerfile-withgo
```diff
@@ -0,0 +1,26 @@
...
+
+LABEL maintainer "dev@pilosa.com"
+
+COPY . /go/src/github.com/pilosa/pilosa/
+
+RUN cd /go/src/github.com/pilosa/pilosa \
+    && CGO_ENABLED=0 make install-dep install FLAGS="-a"
+
+# download pumba for fault injection
+RUN wget https://github.com/alexei-led/pumba/releases/download/0.6.0/pumba_linux_amd64 -O /pumba
```
codysoyland: Use ADD: https://docs.docker.com/engine/reference/builder/#add

.circleci/config.yml
```diff
@@ -22,6 +22,7 @@ jobs:
       - persist_to_workspace:
           root: .
           paths: "*"
+      - setup_remote_docker
```
codysoyland: I think this needs to be a step inside the cluster-tests section, just below `- *fast-checkout`.

codysoyland:
You need to add it to the workflow, near the bottom. This can go anywhere below `- build`:
```
       - cluster-tests:
           requires:
             - build
```

PR #1720, opened 2018-11-03
added bulk support for column attributes -WIP

tgruben:
## Overview
Initial pass at bulk importing of column attributes for discussion.  In order to utilize this feature, you will have to update my patched version of the go client.

### DIFF
--- a/http/handler.go
+++ b/http/handler.go
@@ -269,12 +269,6 @@ func newRouter(handler *Handler) *mux.Router {
 	router.HandleFunc("/internal/translate/column-ids", handler.handlePostTranslateColumnIDs).Methods("POST").Name("PostTranslateColumnIDs")
 	router.HandleFunc("/internal/column-attrs", handler.handlePostColumnAttrs).Methods("POST").Name("PostColumnAttrs")
 
-	// TODO: Apply MethodNotAllowed statuses to all endpoints.
-	// Ideally this would be automatic, as described in this (wontfix) ticket:
-	// https://github.com/gorilla/mux/issues/6
-	// For now we just do it for the most commonly used handler, /query
-	router.HandleFunc("/index/{index}/query", handler.methodNotAllowedHandler).Methods("GET")
-
 	router.Use(handler.queryArgValidator)
 	return router
 }


http/handler.go
```diff
@@ -239,6 +240,9 @@ func newRouter(handler *Handler) *mux.Router {
 	router.HandleFunc("/index/{index}", handler.handlePostIndex).Methods("POST").Name("PostIndex")
 	router.HandleFunc("/index/{index}", handler.handleDeleteIndex).Methods("DELETE").Name("DeleteIndex")
 	//router.HandleFunc("/index/{index}/field", handler.handleGetFields).Methods("GET") // Not implemented.
+	router.HandleFunc("/index/{index}/bulk-column-attrs", handler.handlePostBulkColumnAttrs).Methods("POST").Name("BulkColumnAttrs")
```
travisturner: I think this should be `import-column-attrs` and `handleImportColumnAttrs`

attr.go
```diff
@@ -28,6 +28,7 @@ const (
 	attrTypeInt    = 2
 	attrTypeBool   = 3
 	attrTypeFloat  = 4
+	thisisa        = 5
```
travisturner: is this necessary?

apimethod_string.go
```diff
@@ -4,9 +4,9 @@ package pilosa
 
 import "strconv"
 
-const _apiMethod_name = "apiClusterMessageapiCreateFieldapiCreateIndexapiDeleteFieldapiDeleteIndexapiDeleteViewapiExportCSVapiFragmentBlockDataapiFragmentBlocksapiFieldapiFieldAttrDiffapiImportapiImportValueapiIndexapiIndexAttrDiffapiQueryapiRecalculateCachesapiRemoveNodeapiResizeAbortapiSetCoordinatorapiShardNodesapiViews"
+const _apiMethod_name = "apiClusterMessageapiCreateFieldapiCreateIndexapiDeleteFieldapiDeleteAvailableShardapiDeleteIndexapiDeleteViewapiExportCSVapiFragmentBlockDataapiFragmentBlocksapiFieldapiFieldAttrDiffapiImportapiImportValueapiIndexapiIndexAttrDiffapiQueryapiRecalculateCachesapiRemoveNodeapiResizeAbortapiSetCoordinatorapiShardNodesapiViews"
```
travisturner: let's commit an updated api stringer in its own PR. you can just remove this file from this PR for now.

internal/public.proto
```diff
@@ -40,6 +40,10 @@ message Bit {
 	int64 Timestamp = 3;
 }
 
+message BulkColumnAttrRequest {
```
travisturner: I don't think regenerating should have changed `private.pb.go`. It looks like is a golang or protoc version difference. Let's check with @codysoyland to be sure we're all using the same version to generate code. And if the version does need to change, let's first commit the version-specific change in its own PR.

http/handler.go
```diff
@@ -1657,3 +1661,37 @@ func (h *Handler) handlePostImportRoaring(w http.ResponseWriter, r *http.Request
...
+	}
+	indexName := mux.Vars(r)["index"]
+	body, err := ioutil.ReadAll(r.Body)
+	start := time.Now()
+
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+		// Decode request.
```
travisturner: this is strange indentation. do you need to run `gofmt`?

internal/public.proto
```diff
@@ -40,6 +40,10 @@ message Bit {
...
+        repeated ColumnAttrSet ColAttrSets= 1;
```
travisturner: I believe the indentation here is off.

handler.go
```diff
@@ -76,6 +76,10 @@ func (n nopHandler) Close() error {
 
 var NopHandler Handler = nopHandler{}
 
+type BulkColumnAttrRequest struct{
+ColumnAttrSets	[]*ColumnAttrSet
```
travisturner: gofmt

api.go
```diff
@@ -1014,6 +1014,34 @@ func (api *API) Info() serverInfo {
 	}
 }
 
+func (api *API) BulkImportColumnAttrs(indexName string, req *BulkColumnAttrRequest) error {
+	if len(req.ColumnAttrSets)==0{
+		return nil //nothing to do
+	}
+	//assume if keys all keys else ids
```
travisturner: add some clarity to this comment.

encoding/proto/proto.go
```diff
@@ -668,6 +678,12 @@ func encodeRecalculateCaches(*pilosa.RecalculateCaches) *internal.RecalculateCac
 	return &internal.RecalculateCaches{}
 }
 
+func decodeBulkColumnAttrRequest(msg *internal.BulkColumnAttrRequest, m *pilosa.BulkColumnAttrRequest) {
+        m.ColumnAttrSets = make([]*pilosa.ColumnAttrSet, len(msg.ColAttrSets))
+	fmt.Println("decodeBulkColumnAttrRequest")
```
travisturner: remove debug print

PR #1733, opened 2018-11-08
Adds NoStandardView field option. Fixes #1710

### LINKED ISSUE #1710: Add a flag to suppress the standard view on time fields.
### Description
`time` fields automatically create/populate a `standard` view which represents all setbits with no time value taken into consideration (like a `set` field). It's possible that a user wouldn't need that data. In that case, we could give the user the option to suppress that view (i.e. improve write performance and reduce storage requirements).

I guess this would mean modifying `OptFieldTypeTime()` to include this additional option. Note that that would be an API change. We could also provide a separate option, but that may be confusing.

yuce:
## Overview

Fixes #1710 

* Adds `NoStandardView` field option
* Adds `OptFieldTypeTimeOptions` function which is used instead of `OptFieldTypeTime` for suppressing standard view for time options. I've added another opt function instead of modifying `OptFieldTypeTime` in order to keep compatibility and also I guess not everyone would like to enable suppressing standard view.

## Pull request checklist

## Code review checklist


yuce:
The problem turned out to be my mistake. All tests pass with the updated code.

field.go
```diff
@@ -132,6 +132,10 @@ func OptFieldTypeInt(min, max int64) FieldOption {
 }
 
 func OptFieldTypeTime(timeQuantum TimeQuantum) FieldOption {
+	return OptFieldTypeTimeOptions(timeQuantum, false)
+}
+
+func OptFieldTypeTimeOptions(timeQuantum TimeQuantum, noStandardView bool) FieldOption {
```
travisturner: what do you think about changing the function signature of `OptFieldTypeTime` to be:
```
func OptFieldTypeTime(timeQuantum TimeQuantum, args ...bool) FieldOption
```
so that we don't have to introduce another Opt function? I believe technically that wouldn't break an existing implementation, and we could check for `len(args)` and use `args[0]` to represent `noStandardView`.

travisturner: Regardless, I think we should add a `TODO (2.0)` comment to clean this up. At that point, I feel like we should suppress the standard view by default, and possibly not support it at all; I'm not convinced that it's a useful view, and it can easily be supported by maintaining a separate `set` field.

travisturner: I guess my thinking is that we won't likely be adding anything else to this function until we get to 2.0. I realize that that may not be the case, but I would be ok taking that chance.

And I agree, it's definitely not a clean solution, but if feels a bit more contained than adding a new function that is almost the same thing. Again, I see this as a somewhat temporary solution until 2.0.

yuce: updated

yuce:
updated

Issue #735, opened 2017-07-18
Create Go compiler benchmark


jaffee:
Met @TocarIP at Gophercon who suggested we submit a Pilosa benchmark to the Go compiler team.

From him:
Basic requirements are pretty simple:
1)  Bench output must be in the same format as “go test –bench”. Using go test is nice, but  as long as bench outputs same
`“BenchmarkFOO       200000          10457 ns/op     97.92 MB/s”` lines other build options are ok
2) It should be runnable on 1  machine
3) Benchmark should be Cpu  bound
4) It should be relatively stable with preferably <3% swings in consecutive runs
5) Runtime should be in the order of 10 seconds
6) Hotspots should be in Go code and not in some cgo library

TocarIP:
Quick grep shows some benchmarks in roaring/, is this something worth looking at?

jaffee:
Hey @TocarIP - sorry we've let this hang for so long. The benchmarks in `roaring/roaring_test.go` are definitely worth looking into. I think the ones currently in there are all pretty relevant to Pilosa's performance, but I'll let @tgruben chime in, particularly on the Container* ones as I think he wrote a lot of them.

We could (and should) definitely expand this suite as well. There are several fundamental bitmap operations, and we've really only benchmarked IntersectionCount in there. There are also more combinations of bitmap formats that we could benchmark on. I think ideally we would have:
```
for op in [Intersect, Union, Xor, Difference, IntersectionCount]
    for pairing in [BitmapBitmap, ArrayArray, RunRun, BitmapRun, BitmapArray, ArrayRun]
        Benchmark op(pairing)
```
plus difference is not commutative, so we'd need difference(run, bitmap), difference(array, bitmap), and difference(run, array).

TocarIP:
I've spent some time looking into current benchmarks.
Looking at Bitmap_IntersectionCount_ArrayRun/Bitmap_IntersectionCount_ArrayBitmap shows some possibilities for optimiaztion. I've sent PR with those optimization implemented by hand, so if you want you receive benefits, without waiting for go1.11/1.12/etc
ContainerLinear/ContainerReverse are more of a gc/memory allocator benchmark and are not that interesting.
ContainerColumn/ContainerOutsideIn/ContainerInsideOut  spend 99% in memmove, so they don't provide any new info relative to simple memmove benchmark.
SliceAscending/SliceDescending/Bitmap_IntersectionCount_BitmapRun look interesting, but I haven't looked into possible optimizations.
[...]

PR #1741, opened 2018-11-13
Seebs/bench

### LINKED ISSUE #854: review intersectBitmapRun() logic, specifically the return container type.
### Description
in `roaring/roaring.go`, the comment for `intersectBitmapRun()` says:

```
// intersectBitmapRun returns an array container if the run container's
// cardinality is < ArrayMaxSize. Otherwise it returns a bitmap container.
```
Based on the code, however, it appears that it may return an array (not a bitmap) in the second case with this code:
```
if output.n < ArrayMaxSize {
    output.bitmapToArray()
}
```
I tried changing this to:
```
output.Optimize()
```
but this caused one of the tests to fail. Which makes me think that the test is expecting this function to return an array, when it should, optimized, return a different type.

Tests pass and we can correctly optimize `intersectBitmapRun()`.

seebs:
## Overview

Some benchmark updates, minor bug fixes, and an easy low-hanging performance fix.

(Note: No actual documentation changes, but I have updated all the documentation related to what I'm changing.)

Fixes #854 
Updates #735 

## Pull request checklist

## Code review checklist

### DIFF
--- a/roaring/roaring.go
+++ b/roaring/roaring.go
@@ -1903,16 +1903,26 @@ func intersectionCount(a, b *Container) int32 {
 
 func intersectionCountArrayArray(a, b *Container) (n int32) {
 	statsHit("intersectionCount/ArrayArray")
-	na, nb := len(a.array), len(b.array)
-	for i, j := 0, 0; i < na && j < nb; {
-		va, vb := a.array[i], b.array[j]
-		if va < vb {
-			i++
-		} else if va > vb {
-			j++
-		} else {
+	s1, s2 := a.array, b.array
+	if len(s1) == 0 || len(s2) == 0 {
+		return 0
+	}
+	if len(s1) > len(s2) {
+		s1, s2 = s2, s1
+	}
+	l2 := len(s2)
+	i2 := 0
+	v2 := s2[0]
+	for _, v1 := range s1 {
+		for v2 < v1 {
+			i2++
+			if i2 >= l2 {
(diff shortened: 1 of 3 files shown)


roaring/roaring.go
```diff
@@ -1868,16 +1868,26 @@ func intersectionCount(a, b *Container) int32 {
...
 func intersectionCountArrayArray(a, b *Container) (n int32) {
-	na, nb := len(a.array), len(b.array)
-	for i, j := 0, 0; i < na && j < nb; {
-		va, vb := a.array[i], b.array[j]
-		if va < vb {
-			i++
-		} else if va > vb {
-			j++
-		} else {
+	s1, s2 := a.array, b.array
```
alanbernstein: There's a de facto convention for roaring variable names, and there's a new convention being introduced here, because of the swap. Whatever that is, can we settle on something and stick with it, in any future swap logic?

I might have used `n1` instead of `l1`, and maybe `c1, c2` - `c` for container.

seebs: good catch, i'll fix that.

seebs: on consideration, it's better to just use the existing convention, and just let the swap be a swap of a and b. we already have these implicitly in a number of places; an intersection count between "a" and "b" can be implemented by passing them to a concrete intersection count function in either order, so it's not breaking anything to just say "without loss of generality, let a be the shorter array".

Issue #1765, opened 2018-11-28
Proposal: Improve performance of Union function when unioning many bitmaps


richardartoul:
### Description
The existing implementation of the `Union` function is very expensive (terms of CPU cycles and allocations) if many different bitmaps need to be unioned together.

### Success criteria (What criteria will consider this ticket closeable?)
This ticket would be considered closeable if we were able to make significant improvements in terms of reducing CPU cycles and allocations when unioning together many bitmaps.

### Additional Context
We're currently using [Pilosa's Roaring bitmap implementation](https://github.com/pilosa/pilosa/tree/master/roaring) as well as the "roaring" [Roaring Bitmap implementation](https://github.com/RoaringBitmap/roaring) in our [open source time-series database M3DB](https://github.com/m3db/m3). We want to move away from using two different roaring bitmap implementations because converting between them at query time is expensive and allocation heavy, however, when we tried to switch to using only the Pilosa implementation, we found that the `Union` operation was far too CPU and allocation heavy when its used to union together many different bitmaps at once which is required by our use-case.

We're proposing that we take on the work of improving the `Union` functions performance (either by modifying the existing function or creating a new one) so that we can achieve performance that is at least comparable to the roaring implementation which has a [FastOr](https://github.com/RoaringBitmap/roaring/blob/f158f6388279048d9a3ca095f9aeac58cd2aaed2/fastaggregation.go#L159) implementation that is (relatively) efficient at unioning together many different bitmaps.

If the Pilosa team is open to us contributing this work, we have a few questions:

1.) Does a function/method signature along the lines of

```golang
func(b *bitmap) Union(bitmaps ...*bitmap) *bitmap
```

seem reasonable? We may be able to achieve that by modifying the existing function and not adding a new one which would be nice since I believe this would be a non-breaking interface change.

2. There may be some potential for additional performance improvements if we know the bitmaps are immutable (as they are in our production database case). Is the Pilosa team open to us adding some kind of distinction to the roaring bitmap implementation to distinguish between immutable and mutable bitmaps? Maybe something along the lines of adding a `Seal` method? I'm not 100% sure we will need this until I start actually getting my hands dirty, but thought it might be good to discuss upfront. A potential example is when unioning two bitmap B into bitmap A and you find that bitmap B has a container that bitmap A does not. In that case, it would be much faster to just copy a reference to bitmap Bs container into bitmap A, but that is only safe if bitmap B is immutable and that container in bitmap A is marked as immutable. 

[...]

PR #1766, opened 2018-11-29
Implement fast roaring bitmap union (Part 1)

richardartoul:
## Overview

This pull partially addresses #1765 by providing a much faster and less allocation heavy implementation of the union operation.

## Pull request checklist

## Code review checklist

### DIFF
--- a/roaring/roaring.go
+++ b/roaring/roaring.go
@@ -406,9 +406,10 @@ func (b *Bitmap) Union(others ...*Bitmap) *Bitmap {
 // UnionInPlace returns the bitwise union of b and other, modifying
 // b in place.
 func (b *Bitmap) UnionInPlace(others ...*Bitmap) {
-	for _, other := range others {
-		b.unionIntoTarget(other, b)
-	}
+	b.unionIntoTarget(b, others...)
+	// for _, other := range others {
+	// 	b.unionIntoTarget(other, b)
+	// }
 }
 
 type wrapperIter struct {
@@ -421,17 +422,135 @@ type wrapperIter struct {
 // be left unchanged, but target will be modified in place. Used to share
 // the union logic between the copy-on-write and in-place functions.
 func (b *Bitmap) unionIntoTarget(target *Bitmap, others ...*Bitmap) {
-	otherIters := make([]wrapperIter, 0, len(others))
-	for _, other := range others {
-		otherIter, _ := other.Containers.Iterator(0)
+	fmt.Println("huh: ", len(others))
+	otherIters := make([]wrapperIter, 0, len(others)+1)
+	bIter, _ := b.Containers.Iterator(0)
+	next := bIter.Next()
+	if next {
+		fmt.Println("Appending self")
 		otherIters = append(otherIters, wrapperIter{
-			iter: otherIter,
+			iter:    bIter,
+			hasNext: true,
+			handled: false,
 		})
 	}
 
+	for _, other := range others {
+		otherIter, _ := other.Containers.Iterator(0)
+		next := otherIter.Next()
+		if next {
+			otherIters = append(otherIters, wrapperIter{
+				iter:    otherIter,
+				hasNext: true,
+				handled: false,
+			})
+		}
+	}
+
 	// Loop until we've exhausted every iter.
-	for {
-		hasNext := false
+	hasNext := true
+	fmt.Println(len(otherIters))
+	for hasNext {
+		// Loop until every iters current value has been handled.
+		// for {
+		for i, iIter := range otherIters {
+			fmt.Println("copter")
+			if !iIter.hasNext || iIter.handled {
+				continue
+			}
+
+			fmt.Println("here?")
+			// Can store key-level statistics here
+			iKey, iContainer := iIter.iter.Value()
+			fmt.Println("iKey: ", iKey)
+			fmt.Println("iContainer: ", iContainer)
+			n := iContainer.n
+			needsUnion := false
+			hasMaxRange := iContainer.n == maxContainerVal+1
+			for _, jIter := range otherIters[i:] {
+				if hasMaxRange {
+					continue
+				}
+
+				// Calculate key-level statistics here
+				jKey, jContainer := jIter.iter.Value()
+				fmt.Println("jContainer.n: ", jContainer.n)
+
+				if iKey == jKey {
+					needsUnion = true
+					n += jContainer.n
+					if !hasMaxRange {
+						hasMaxRange = jContainer.n == maxContainerVal+1
+					}
+				}
+			}
+
+			if !needsUnion {
+				fmt.Println("Cloning")
+				// TODO: Don't clone if sealed
+				target.Containers.Put(iKey, iContainer.Clone())
+				otherIters[i].handled = true
+				continue
+			}
+
+			// Need to union
+			fmt.Println("ikey: ", iKey)
+			if hasMaxRange {
+				panic("maxRange")
+				fmt.Println("maxRange")
+				// Use the max range
+				container := &Container{
+					runs:          []interval16{{start: 0, last: maxContainerVal}},
+					containerType: containerRun,
+					n:             maxContainerVal + 1,
+				}
+				target.Containers.Put(iKey, container)
+			} else {
+				// TODO: Implement this
+				// if n < ArrayMaxSize {
+				// 	// Use an array
+				// 	// container := &Container{
+				// 	// 	containerType: containerArray,
+				// 	// 	array:         make([]uint16, 0, n),
+				// 	// }
+				// }
+				// else {
+				// Use a bitmap
+				fmt.Println("bitmap")
+				buf := make([]uint64, bitmapN)
+				ob := buf[:bitmapN]
+				container := &Container{
+					bitmap:        ob,
+					n:             0,
+					containerType: containerBitmap,
+				}
+
+				for _, jIter := range otherIters {
+					jKey, jContainer := jIter.iter.Value()
+
+					if iKey == jKey {
+						if jContainer.isArray() {
+							// panic("unionBitmapArrayInPlace")
+							fmt.Println("array into bitmap")
+							unionBitmapArrayInPlace(container, jContainer)
+							fmt.Println("After union array: ", container.n)
+						} else if jContainer.isRun() {
+							// panic("unionBitmapRunInPlace")
+							unionBitmapRunInPlace(container, jContainer)
+						} else {
+							fmt.Println("bitmap into bitmap")
+							// panic("unionBitmapBitmapInPlace")
+							unionBitmapBitmapInPlace(container, jContainer)
+							fmt.Println("After union bitmap: ", container.n)
+						}
+					}
+				}
+				target.Containers.Put(iKey, container)
+			}
+		}
+		// }
+
+		hasNext = false
 		for i, otherIter := range otherIters {
 			next := otherIter.iter.Next()
 			otherIters[i].hasNext = next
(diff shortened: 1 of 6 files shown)


alanbernstein:
Thank you for the thorough comments!

richardartoul:
@alanbernstein No problem. Any idea why the race test is failing? there is no output

roaring/roaring.go
```diff
@@ -395,32 +395,300 @@ func (b *Bitmap) Intersect(other *Bitmap) *Bitmap {
...
+		hasNext = bitmapIters.next()
+	}
+
+	// Performing the popcount() operation with every union is wasteful because
+	// its likely the value will be invalidated by the next union operation. As
+	// a result, when we're performing all of our in-place unions we allow the value of
+	// n (container cardinality) to fall out of sync, and then at the very end we perform
+	// a "Repair" to recalculate all the container values. That way we never popcount()
+	// an entire bitmap container more than once per bulk union operation.
+	target.Containers.(*sliceContainers).repairBitmaps()
```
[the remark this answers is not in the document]
richardartoul: Yeah I actually had it like that, but it felt a little leaky, especially since it seems like none of the functions in this package would work if containers were anything other than slice (all the functions assume a container is an array, bitmap, or RLE).

Could you give me some guidance here? My thought process is add Repair() to Container interface, then in the sliceContainer implementation it will only repair bitmaps and in the tree implementation its a no-op. Is that reasonable?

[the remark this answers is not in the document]
jaffee: the containers themselves are always bitmap/array/rle. The btree is for tracking which containers actually have data and where they are. This becomes a performance problem when you have potentially 10s of millions of containers in a single bitmap (this actually can happen in Pilosa due to the way we store our fragments).

now that I've looked at it a bit more, though, `repairBitmaps` doesn't actually need to be a method on `sliceContainers`, it could be a function that takes an implementation of `Containers`, and iterates through the containers calling repair on each one. e.g.

```func RepairContainers(c Containers) {
    citer := c.Iterator()
    // iterate through the containers calling bitmapRepair on each
}
```

Though if this is significantly slower, we may need separate direct implementations.

Does that make sense? I think our naming could be better, but the takeaway is that what's inside a container is always the same, but the data structure tracking the containers might be different.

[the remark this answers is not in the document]
tgruben: should look something like this @richardartoul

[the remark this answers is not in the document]
travisturner: @tgruben wouldn't this also require exposing `IsBitmap()` and `BitmapRepair()` on the `Container` type?

roaring/roaring.go
```diff
@@ -395,32 +395,300 @@ func (b *Bitmap) Intersect(other *Bitmap) *Bitmap {
 	return output
 }
 
-// Union returns the bitwise union of b and other.
-func (b *Bitmap) Union(other *Bitmap) *Bitmap {
+// Union returns the bitwise union of b and other as a new bitmap.
```
travisturner: should read "... of b and others as ..."

roaring/roaring.go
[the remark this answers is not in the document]
tgruben: right...i forgot that wasn't in the roaring package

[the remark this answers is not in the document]
richardartoul: it can probably just be something like:

```golang
// roaring package
type Containers interface {
    ...
    Repair()
}

func (c *Container) Repair() {
    if c.isBitmap() {
        c.bitmapRepair()
    }
}
```

```golang
// btree package
func (btc *btreeContainers) Repair() {
    e, _ := btc.tree.Seek(0)
    _, c, err := e.Next()
    for err != io.EOF {
        c.Repair()
    }
    _, c, err = e.Next()
}
```

Hows that look?

roaring/roaring.go
```diff
@@ -395,32 +395,300 @@ func (b *Bitmap) Intersect(other *Bitmap) *Bitmap {
...
+				// union we can generate an RLE container that represents the entire
+				// range.
+				container := &Container{
+					runs:          []interval16{{start: 0, last: maxContainerVal}},
+					containerType: containerRun,
+					n:             maxContainerVal + 1,
+				}
+				target.Containers.Put(iKey, container)
+				bitmapIters[i:].markItersWithCurrentKeyAsHandled(iKey)
+			} else {
```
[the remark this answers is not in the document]
richardartoul: yeah that would be better, will fix

roaring/roaring.go
```diff
@@ -395,32 +395,300 @@ func (b *Bitmap) Intersect(other *Bitmap) *Bitmap {
...
+						} else if jContainer.isRun() {
+							unionBitmapRunInPlace(container, jContainer)
+						} else {
+							unionBitmapBitmapInPlace(container, jContainer)
+						}
+						bitmapIters[j].handled = true
+					}
+				}
+
+				// Now that we've calculated a container is that a union of all the containers
```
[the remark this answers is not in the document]
richardartoul: fixed

roaring/roaring.go
```diff
@@ -395,32 +395,300 @@ func (b *Bitmap) Intersect(other *Bitmap) *Bitmap {
...
+	} else {
+		bitmapIters = make(handledIters, 0, requiredSliceSize)
+	}
+
+	// Only include b in the list of iters if its not the same as target to avoid
+	// a wasteful self union.
+	if b != target {
+		bIter, _ := b.Containers.Iterator(0)
+		next := bIter.Next()
+		if next {
```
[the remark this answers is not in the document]
richardartoul: sure

roaring/roaring.go
```diff
@@ -395,32 +395,300 @@ func (b *Bitmap) Intersect(other *Bitmap) *Bitmap {
...
+				hasNext: true,
+				handled: false,
+			})
+		}
+	}
+
+	for _, other := range others {
+		otherIter, _ := other.Containers.Iterator(0)
+		next := otherIter.Next()
+		if next {
```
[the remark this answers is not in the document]
richardartoul: sure

roaring/roaring.go
richardartoul: fixed

roaring/roaring.go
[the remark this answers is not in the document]
tgruben: i like it.   But you need to put`e.Next()`  inside the loop

roaring/roaring.go
```diff
@@ -3637,3 +3988,85 @@ func readWithRuns(b *Bitmap, data []byte, pos int, keyN uint32) {
...
+		}
+
+		// Calculate key-level statistics here
+		currKey, currContainer := iter.iter.Value()
+
+		if key == currKey {
+			summary.isOnlyContainerWithKey = false
+			summary.n += currContainer.n
+			if !summary.hasMaxRange {
+				summary.hasMaxRange = (currContainer.n == maxContainerVal+1)
```
seebs: So far as I can tell, if this conditional ends up true, we are actually done. We could omit all tests in this function against summary.hasMaxRange, and just `if currContainer.n == maxContainerVal+1 { summary.hasMaxRange = true; return summary }`.

roaring/roaring.go
```diff
@@ -395,32 +399,299 @@ func (b *Bitmap) Intersect(other *Bitmap) *Bitmap {
...
+//          ----------------------------      |      ----------------------------      |      ----------------------------
+// Bitmap 2 |_______X_______________X___|     |      |_______X_______________X___|     |      |_______X_______________X___|
+//                  _                         |                              ^         |                              ^
+//          ----------------------------      |      ----------------------------      |      ----------------------------
+// Bitmap 3 |_______X___________________|     |      |_______X___________________|     |      |_______X___________________|
+//                  _                         |                                        |
+//          ----------------------------      |      ----------------------------      |      ----------------------------
+// Bitmap 4 |___X_______________________|     |      |___X_______________________|     |      |___X_______________________|
+//              _
+func (b *Bitmap) unionIntoTarget(target *Bitmap, others ...*Bitmap) {
```
seebs: Thinking about this:

I wonder if it wouldn't make more sense to have this be

`func (target *Bitmap) unionIntoTarget(others ...*Bitmap)`

There's a lot of special cases in the code that come from the first parameter being the special parameter, rather than the receiver.

roaring/roaring.go
```diff
@@ -395,32 +399,299 @@ func (b *Bitmap) Intersect(other *Bitmap) *Bitmap {
...
+// The algorithm works by iterating through all of the containers in all of the bitmaps concurrently.
+// At every "tick" of the outermost loop, we increment our pointer into the bitmaps list of containers
+// by 1 (if we haven't reached the end of the containers for that bitmap.)
+//
+// We then loop through all of the "current" values(containers) for all of the bitmaps
+// and for each container with a specific key that we encounter, we scan forward to see if any of the
+// other bitmaps have a container for the same key. If so, we calculate some summary statistics and
+// then use that information to make a decision about how to union all of the containers with the same
+// key together, perform the union, mark the unioned containers as "handled" and then move on to the next
+// batch of containers that share the same key.
```
seebs: I was trying to think why "handled" exists, and the answer is, we don't have any guarantee or even strong expectation that every container-key exists in even one bitmap. But I think it might be possible to streamline this logic.

Consider a heap of bitmaps, with the ordering function being "next container key", lowest-first. You pick the lowest thing off the heap, grab its container, iterate it, and push it back into the heap. As long as the top of the heap has the same key as your current container, do the same thing to that. When you run out of things with the same key, you have a pool of containers with the same key, and the heap contains the next likely key...

But that'll be horrendously expensive in the probably-common case where most bitmaps have most containers.

It does seem, though, like somehow arranging to process by "lowest container key first" might reduce bookkeeping; you wouldn't have to track "handled" if you weren't possibly handling a key that's higher than some bitmap's "current" key. And intuitively, for at least some of the formats, I think you'll get better performance iterating through them strictly in order, rather than suddenly grabbing something from far later in a bitmap and then jumping back.

roaring/roaring.go
```diff
@@ -98,8 +98,12 @@ type Containers interface {
 
 	Count() uint64
 
-	//Reset will clear the containers collection to allow for recycling during snapshot
+	// Reset will clear the containers collection to allow for recycling during snapshot
```
seebs: Idiomatically, I think go comments prefer to drop the helper verbs. "Reset clears the ..."

richardartoul: sure, I was just fixing the whitespace anyways

roaring/roaring.go
richardartoul: Yeah I see what you're saying. If you think about it though the `handledIter` slice thing is basically my way of doing it. If I tucked the `handled` logic into the `handledIter` datastructure itself, I could make it expose a separate iterator that provided the semantics of "lowest container key" iteration. I actually tried doing that in one implementation and I didn't like how it made the code read because its kind of a leaky abstraction and things get messy quickly.

Ultimately though we have to deal with the data in the format that it is. Even small amounts of allocations or restructuring will degrade the performance and memory utilization significantly. The reason this implementation is so fast is because it doesn't restructure the data at all, and in the optimistic case, I don't even have to allocate the `handledIter` slice.

Its not my favorite part of this code either, but I don't see an easy way to improve this that won't introduce a bunch of allocations or significantly complicate things.

roaring/roaring.go
richardartoul: That worked, thanks for the idea

richardartoul:
@seebs How does the latest version look? Also, any idea about the failing C.I?

roaring/roaring.go
```diff
@@ -395,32 +399,282 @@ func (b *Bitmap) Intersect(other *Bitmap) *Bitmap {
...
+//        900 + 3500 > ArrayMaxSize so we will need to upgrade to a bitset container which
+//        we will have to allocate, and then we will have to perform two unions into the
+//        new bitmap container: one for the array container generated in the previous step,
+//        and one for the bitset container in bitmap C.
+//
+// Approach taken by this function:
+//
+//     1. Detect that bitmaps A, B, and C all have containers for key 0.
+//     2. Estimate the resulting cardinality of the union of all their containers to be
+//        400 + 500 + 3500 > ArrayMaxSize and decide upfront to use a bitset for the target
```
seebs: I don't think this math is necessarily right -- we don't know whether the values overlap, so we don't know how many values we'll end up with. Not sure it's worth worrying about, though.

roaring/roaring.go
```diff
@@ -395,32 +399,282 @@ func (b *Bitmap) Intersect(other *Bitmap) *Bitmap {
...
+//
+// In the naive approach, we had to allocate two containers, whereas in the optimized approach
+// we only had to allocate one container, and we also had to perform less union operations. This
+// example is simplistic, but the impact in terms of CPU cycles and memory allocations achieved
+// by using the optimized alogorithm when unioning many large bitmaps can be huge.
+//
+// An additional optimization that this function makes is that it recognizes that even when
+// CPU support is present, performing the popcount() operation isn't free. Imagine a scenario
+// where 10 bitset containers are being unioned together one after the next. If every
+// bitset<->bitset union operation needs to keep the containers cardinality up to date, then
```
seebs: I think `containers'`, because that's a possessive.

roaring/roaring.go
```diff
@@ -395,32 +399,282 @@ func (b *Bitmap) Intersect(other *Bitmap) *Bitmap {
...
+// Bitmap 3 |_______X___________________|     |      |_______X___________________|     |      |_______X___________________|
+//                  _                         |                                        |
+//          ----------------------------      |      ----------------------------      |      ----------------------------
+// Bitmap 4 |___X_______________________|     |      |___X_______________________|     |      |___X_______________________|
+//              _
+func (b *Bitmap) unionInPlace(others ...*Bitmap) {
+	var (
+		requiredSliceSize = len(others)
+		// To avoid having to allocate a slice everytime, if the number of bitmaps
+		// being unioned is small enough we can just use this stack-allocated array.
```
seebs: Does this actually work? My concern would be that escape analysis might not be able to prove it's safe, in which case the array may end up heap-allocated anyway.

roaring/roaring.go
```diff
@@ -2529,6 +2841,28 @@ func unionBitmapBitmap(a, b *Container) *Container {
...
+	var (
+		ab = a.bitmap[:bitmapN]
+		bb = b.bitmap[:bitmapN]
+	)
+
+	// Manually unroll loop to make it a little faster.
+	// TODO(rartoul): Can probably make this a few x faster using
+	// SIMD instructions.
+	for i := 0; i < bitmapN; i += 4 {
+		ab[i] = ab[i] | bb[i]
```
seebs: I think `|=` would be more expressive here.

I'm also not sure whether the loop unrolling is effective, but I don't suppose it hurts?

roaring/roaring.go
richardartoul: yeah its definetly just a rough approximation. It prevents us from having to use a bitmap everytime, but will definitely be biased towards being more aggressive and using a bitmap because we can't be sure.

roaring/roaring.go
richardartoul: Yeah in my benchmarks this reduced the number of allocations per call by 1. This was one of the nitty gritty optimization I added at the end though, its just one allocation and its fairly well amortized so I can remove it if you think its too messy

roaring/roaring.go
richardartoul: I got like 15%ish gains when I unrolled it manually, although thats just on my laptop

roaring/roaring.go
richardartoul: sure thing

roaring/roaring.go
richardartoul: I tried to convey that by using the word "estimate" but maybe I can make the comment more explicit. I just didn't want to use a bitmap if I was unioning two containers, each with a cardinality of 1

roaring/roaring.go
richardartoul: this is another one of those overly aggressive optimization that helps a little, but isn't world changing. Can leave it out of its too messy

travisturner:
I did my best to confirm that the CircleCI race detection that was failing is not directly related to this PR. There seems to be and issue with cluster logic on the node-leave event, and I filed an issue for that here: #1772.

roaring/roaring.go
richardartoul: Updated the comment!

roaring/roaring.go
richardartoul: fixed!

richardartoul:
@travisturner Awesome thanks, I addressed the last bits of feedback about my comments. I'll rebase master and push again

roaring/roaring.go
richardartoul: switched to `|=`

richardartoul:
@travisturner everything should merge cleanly now

travisturner:
Thanks @richardartoul. Nice work!

PR #1773, opened 2018-12-08
Cancel queries on Context.Done()

### LINKED ISSUE #1745: Query Cancellation
### Description

Currently we are propagating contexts throughout our stack, but once a query has been kicked off, there's no way to cancel it early. This becomes an issue if a client executes an extremely complex query (think Union of 200,000 rows), and then gives up before the query completes. The Pilosa cluster may remain loaded by the query for quite some time. Pilosa executors should periodically check the context and be able to exit early from queries, gracefully unlocking any locks, cleaning up goroutines, releasing memory, etc.

Queries can be cancelled.

benbjohnson:
## Overview

This commit periodicially checks if the context has been cancelled
or if a deadline has been reached. If so, it returns a query-related
error message depending on the cause.

Fixes #1745

## Pull request checklist

## Code review checklist

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -87,6 +87,11 @@ func (e *executor) Execute(ctx context.Context, index string, q *pql.Query, shar
 
 	resp := QueryResponse{}
 
+	// Check for query cancellation.
+	if err := validateQueryContext(ctx); err != nil {
+		return resp, err
+	}
+
 	// Verify that an index is set.
 	if index == "" {
 		return resp, ErrIndexRequired
@@ -112,11 +117,15 @@ func (e *executor) Execute(ctx context.Context, index string, q *pql.Query, shar
 	if !opt.Remote {
 		if err := e.translateCalls(ctx, index, idx, q.Calls); err != nil {
 			return resp, err
+		} else if err := validateQueryContext(ctx); err != nil {
+			return resp, err
 		}
 	}
 
 	results, err := e.execute(ctx, index, q, shards, opt)
 	if err != nil {
+		return resp, err
+	} else if err := validateQueryContext(ctx); err != nil {
 		return resp, err
 	}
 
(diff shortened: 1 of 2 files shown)


benbjohnson:
@jaffee The `validateQueryContext()` is executed inside `executor.executeCall()` so it checks on nested calls as well. I tested it against a `Union()` with a ton of child calls (not sure if it was 50K but it was a lot) and it successfully cancels.

Do you have an example query that's not canceling?

jaffee:
@benbjohnson 	AFAICT, executor.executeCall is only called on top level queries and queries inside of an `Options` call. I feel like I must be missing something obvious, but I can't see how a call like: `Union(Row(afield=1), Row(afield=2), Row(afield=3)..., Row(afield=9999))` could be interrupted once it gets into executor.executeUnionShard

My very unscientific test was to launch one of these ridiculous super-Union queries from our taxi-demo and then kill the demo server and observe whether the load on Pilosa dropped or continued - my understanding is that if the connection dies the http server will cancel the context.

benbjohnson:
@jaffee Ah, you're totally right. I was executing a large query then waiting a few seconds and canceling but the query was so large it was still in the parse phase so it caught the cancel before execution.

I went through and checked all `pql.Call.Children` usage in `executor` and it looks like just the `executeBitmapCallShard()` needed a check added in order to catch them all. I did some more testing and verified that it cancels once it's in the execution phase. `/facepalm`

jaffee:
so what you're saying is, if our darn parser wasn't so slow...   ;)

This totally works though! Very cool to see the load drop immediately after killing the client. Thanks!

jaffee:
I'm going to go ahead and merge this as long as all checks are passing and it's up to date with the base branch.

PR #1809, opened 2019-01-02
Add support for container pooling

### LINKED ISSUE #1765: Proposal: Improve performance of Union function when unioning many bitmaps

richardartoul:
## Overview

This pull request adds support for pooling containers as discussed in #1765

The idea is that the existing implementation is wasteful when bitmaps are short-lived because their is no way to re-use them (and their containers). This P.R adds the ability to pre-allocate containers on a given Bitmap, as well as Reset() a bitmap so that those contains can be reused. This allows the bitmaps to be pooled at the application level and can heavily reduce allocations and G.C pressure in situations where bitmaps are short lived.

Before
```
goos: darwin
goarch: amd64
pkg: github.com/pilosa/pilosa/roaring
BenchmarkUnionBulk-8   	  100000	     17075 ns/op	   33536 B/op	      21 allocs/op
PASS
ok  	github.com/pilosa/pilosa/roaring	1.902s
Success: Benchmarks passed.
``` 

After
```
goos: darwin
goarch: amd64
pkg: github.com/pilosa/pilosa/roaring
BenchmarkUnionBulkWithPooling-8   	  200000	     11738 ns/op	     172 B/op	       5 allocs/op
PASS
ok  	github.com/pilosa/pilosa/roaring	2.494s
Success: Benchmarks passed.
```

Fixes #1765

## Pull request checklist

## Code review checklist

### DIFF
--- a/roaring/containers.go
+++ b/roaring/containers.go
@@ -26,12 +26,12 @@ type sliceContainers struct {
 // ContainerPoolingConfiguration represents the configuration for
 // container pooling.
 type ContainerPoolingConfiguration struct {
-	// Whether an array should be allocated for each pooled container.
-	AllocateArray bool
+	// Maximum size of the allocated array that will be maintained in the pool.
+	MaxArraySize int
 	// Whether a bitmap should be allocated for each pooled container.
 	AllocateBitmap bool
-	// Whether a run should be allocated for each pooled container.
-	AllocateRuns bool
+	// Maximum size of the allocated runs that will be maintained in the pool.
+	MaxRunsSize int
 
 	// Maximum number of containers to pool.
 	MaxCapacity int
@@ -43,8 +43,8 @@ type ContainerPoolingConfiguration struct {
 // with default configuration.
 func NewDefaultContainerPoolingConfiguration(maxCapacity int) ContainerPoolingConfiguration {
 	return ContainerPoolingConfiguration{
-		AllocateArray:  true,
-		AllocateRuns:   true,
+		MaxArraySize:   ArrayMaxSize,
+		MaxRunsSize:    runMaxSize,
 		AllocateBitmap: true,
 
 		MaxCapacity: maxCapacity,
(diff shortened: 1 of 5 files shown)


jaffee:
we've diverged from this, and there are some coming changes to sync.Pool in Go 1.13... we will likely revisit something like this in the future.

PR #1817, opened 2019-01-09
reverse the order of items in the github issue template

travisturner:
## Overview

Personally, I would prefer the erroneous behavior experienced due to a bug be the first thing described in the issue. This PR reverses the order of items in the issue template to support that. I'm not sure about what the labels should be, but here's a suggestion.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,8 +1,8 @@
 For bugs, please provide the following:
 
-### Unxpected behavior
+### What's going wrong?
 
-### Expected behavior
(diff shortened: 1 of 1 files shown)


jaffee:
Definitely a good idea.

I'd support some friendlier labels like "What's going wrong?" "What was expected?"

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,8 +1,8 @@
 For bugs, please provide the following:
 
-### Expected behavior
+### Unxpected behavior
```
yuce: Typo here.

PR #1822, opened 2019-01-11
fixes a bug on upper end of bsi range queries

### LINKED ISSUE #1819: Row BETWEEN queries can be wrong on upper end of range.
### What's going wrong?
Row "between" queries seem to return the wrong results when a value is on the upper edge of the range. It seems like ranges are being performed as:
```
(1 < f < 10) => (1, 10]
(1 <= f < 10) => [1, 10]
(1 <= f <= 10) => [1, 10]
(1 < f <= 10) => (1, 10]
```

### What was expected?
I expect the following ranges to be performed as:
```
(1 < f < 10) => (1, 10)
(1 <= f < 10) => [1, 10)
(1 <= f <= 10) => [1, 10]
(1 < f <= 10) => (1, 10]
```

```
curl http://localhost:10101/index/i -d '{"options": {"keys": false, "trackExistence": false}}'
curl http://localhost:10101/index/i/field/f -d '{"options": {"type": "int", "min": 0, "max": 1023}}'

curl http://localhost:10101/index/i/query -d 'Set(1, f=5)'
```
The following are example query results:

query | got | expected
------------ | ------------ | ------------
Row(4 < f < 6) | [1] | [1]
Row(4 < f < 5) | [1] | []
Row(3 < f < 5) | [1] | []
Row(5 < f < 5) | [] | []
Row(5 < f < 6) | [] | []
Row(5 < f <= 5) | [] | []
Row(5 <= f <= 5) | [1] | [1]
Row(5 <= f < 5) | [1] | []

Notice that the queries of the form `(x </<= f < 5)` are incorrect.

travisturner:
## Overview

This PR addresses the issues described in #1819. I believe this was caused by the previous `BETWEEN` logic handling ranges as `[x, y)`, where the upper end was not inclusive. Now that the inclusivity is explicitly designated by `<=` and `>=` it's clear how to handle the range bounds.

Fixes #1819

## Pull request checklist

## Code review checklist

### DIFF
--- a/executor_test.go
+++ b/executor_test.go
@@ -1901,16 +1901,44 @@ func TestExecutor_Execute_Row_BSIGroup(t *testing.T) {
 	})
 
 	t.Run("BETWEEN", func(t *testing.T) {
-		if result, err := c[0].API.Query(context.Background(), &pilosa.QueryRequest{Index: "i", Query: `Row(0 < other < 1000)`}); err != nil {
-			t.Fatal(err)
-		} else if !reflect.DeepEqual([]uint64{0}, result.Results[0].(*pilosa.Row).Columns()) {
-			t.Fatalf("unexpected result: %s", spew.Sdump(result))
+		tests := []struct {
+			q   string
+			exp bool
+		}{
+			{q: `Row(0 < other < 1000)`, exp: false},
+			{q: `Row(0 <= other < 1000)`, exp: false},
+			{q: `Row(0 <= other <= 1000)`, exp: true},
+			{q: `Row(0 < other <= 1000)`, exp: true},
+
+			{q: `Row(1000 < other < 1000)`, exp: false},
+			{q: `Row(1000 <= other < 1000)`, exp: false},
+			{q: `Row(1000 <= other <= 1000)`, exp: true},
+			{q: `Row(1000 < other <= 1000)`, exp: false},
+
+			{q: `Row(1000 < other < 2000)`, exp: false},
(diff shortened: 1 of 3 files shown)


travisturner:
@alanbernstein I don't understand your question. Can you expand on that?

travisturner:
oh, do you mean supporting something like `x > fld > y` (as opposed to `x < fld < y`)?

travisturner:
@alanbernstein are you suggesting that the query language docs should have clarification around this
```
Range([<COMPARISON_VALUE> <COMPARISON_OPERATOR>] <FIELD> <COMPARISON_OPERATOR> <COMPARISON_VALUE>)
```
in that we should explain that `COMPARISON_OPERATOR` in this case can only be `<` or `<=` (and not other types like `>`, `==`, `!=`, etc)?

alanbernstein:
Sorry, yes that's what I meant. The other operators are still defined in the peg file; it's not obvious whether they're still in use for other syntax. If so, we should document the different situations. If they're not, we should simplify the docs page, and remove them from the peg file.

travisturner:
@alanbernstein they are definitely still in use for `int` fields.
I believe the two cases are:
```
conditional <- {p.startConditional()} condint condLT condfield condLT condint {p.endConditional()}
```
which covers `Range(x <= fld <= y)`, and 
```
arg <- field sp COND sp value
```
which covers things like `fld != 10`.

I think the only thing we need to do is specify in the docs which conditionals apply to each situation.

PR #1824, opened 2019-01-14
fixes #1823. Updates tests and docs for row range

### LINKED ISSUE #1823: Never returning response when leaving out the "to" parameter in a Row (range) query
### What's going wrong?

If the `to` parameter is left out from a `Row` query, but the `from` parameter is given, the response is never returned. Also, the CPU utilization and memory consumption increase very fast. Probably due to an infinite loop somewhere. [TODO: Will update with more information]

### What was expected?

`from` parameter for a `Row` (range) query should be usable by itself.

1. Create an index:
```
curl localhost:10101/index/i -d ''
```
2. Create a time field:
```
curl localhost:10101/index/i/field/f -d '{"options":{"type":"time","timeQuantum":"YMDH"}}'
```
3. Set a bit:
```
curl localhost:10101/index/i/query -d 'Set(2, f=1, 1999-12-31T00:00)'
```
4. Run a `Row` query with `from` but without `to`:
```
curl localhost:10101/index/i/query -d 'Row(f=1, from=1999-12-31T00:00)'
```
5. The HTTP response is not returned.

yuce:
## Overview

Fixes #1823 

## Pull request checklist

## Code review checklist

### DIFF
--- a/executor_test.go
+++ b/executor_test.go
@@ -1493,6 +1493,8 @@ func TestExecutor_Execute_Row_Range(t *testing.T) {
 		Set(2, f=10, 2001-01-01T00:00)`
 		readQueries := []string{
 			`Row(f=1, from=1999-12-31T00:00, to=2002-01-01T03:00)`,
+			`Row(f=1, from=1999-12-31T00:00)`,
+			`Row(f=1, to=2002-01-01T02:00)`,
 			`Clear( 2, f=1)`,
 			`Row(f=1, from=1999-12-31T00:00, to=2002-01-01T03:00)`,
 		}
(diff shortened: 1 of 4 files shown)


executor_test.go
```diff
@@ -1505,8 +1507,20 @@ func TestExecutor_Execute_Row_Range(t *testing.T) {
 			}
 		})
 
+		t.Run("From", func(t *testing.T) {
```
travisturner: At first, I was a little surprised to see the implied `toTime` default to now + 1, but I suppose I understand the logic there. With that in mind, I guess my only comment is that there isn't actually a test which shows that a `Row()` query with only a `from` excludes results that are more than 1 day into the future. I guess if there was a record set at date `2200-01-01T00:00` then a test could ensure that it does not come back in the result set.

yuce: I've updated the test to include a timestamp which is greater than the default end timestamp by 1 second to ensure that.

travisturner: I don't think that works on `f=10`; the query is checking `f=1`.
Also, I wonder if there's a case where `nextDayExclusive` ends up being the same as `toTime` since `time.Now()` is determined at different times. Perhaps add more than 1 second?

yuce: Yeah, with 1 second it passed on local, but failed on CI. Added 1 day, waiting for the CI job to finish.

yuce: The `nextDayExclusive.Format("2006-01-02T15:04")` is on the line with `f=10` but actually it affects the `Set(8, f=1, %s)` line. It seems to pass on the CI.

travisturner: Ah right. I see that now.

PR #1831, opened 2019-01-16
make sure more tests and benchmarks can have their temp dir set by flag

jaffee:
This is to allows the directory to be set to where a particular disk is mounted
during benchmarking.

## Pull request checklist

## Code review checklist

### DIFF
--- a/executor_test.go
+++ b/executor_test.go
@@ -16,7 +16,9 @@ package pilosa_test
 
 import (
 	"context"
+	"flag"
 	"fmt"
+	"io/ioutil"
 	"math/rand"
 	"reflect"
 	"strconv"
@@ -32,6 +34,20 @@ import (
 	"github.com/pilosa/pilosa/test"
 	"github.com/pkg/errors"
 )
+
+var (
+	TempDir *string
+)
+
+func init() {
+	tdflag := flag.Lookup("temp-dir")
+	if tdflag == nil {
+		TempDir = flag.String("temp-dir", "", "Directory in which to place temporary data (e.g. for benchmarking). Useful if you are trying to benchmark different storage configurations.")
+	} else {
+		s := tdflag.Value.String()
+		TempDir = &s
+	}
+}
 
 // Ensure a row query can be executed.
 func TestExecutor_Execute_Row(t *testing.T) {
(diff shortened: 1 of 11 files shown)


executor_test.go
```diff
@@ -33,6 +35,20 @@ import (
...
+var (
+	TempDir *string
+)
+
+func init() {
+	tdflag := flag.Lookup("temp-dir")
+	if tdflag == nil {
+		TempDir = flag.String("temp-dir", "", "Directory in which to place temporary data (e.g. for benchmarking). Useful if you are trying to benchmark different storage configurations.")
+	} else {
+		s := tdflag.Value.String()
```
jaffee: so, this whole init() is fun... `TempDir` (and its flag) are actually declared in two files. Both are _test files, so it's only around for tests, but one is in package pilosa, and the other is in pilosa_test. Declaring the same flag twice leads to a panic, so in the pilosa_test package, we guard by looking up the flag first to see if it's already defined, and then only define it if it isn't.

executor_test.go
```diff
@@ -2987,7 +3003,16 @@ func TestExecutor_Execute_SetRow(t *testing.T) {
...
+	c := test.MustNewCluster(b, 1)
+	var err error
+	c[0].Config.DataDir, err = ioutil.TempDir(*TempDir, "benchmarkExistence")
+	if err != nil {
+		b.Fatalf("getting temp dir: %v", err)
+	}
+	err = c.Start()
+	if err != nil {
+		b.Fatalf("starting cluster: %v", err)
+	}
```
jaffee: I didn't want to change any interfaces in our `test` package, so I'm setting up the data dir manually after newing the cluster instead of adding a parameter that we can pass in to `MustRunCluster` which will tell it where to put files.

executor_test.go
codysoyland: Could we avoid the init() call altogether by making funcs like `pilosa.GetTempDir()` and `pilosa.GetTempFile()`?

jaffee: I got rid of the init. I think the way I've done this is slightly better than defining a func to call at each invocation of `ioutil.Temp*`, because we need something that defines the flag at startup time. feel free to disagree though

codysoyland: I didn't actually realize you can call functions from the top level like that. Since that is essentially the same as init() in terms of startup side-effects, it's interesting that there is no linter that checks for that.

PR #1832, opened 2019-01-18
support advertise address and listen on 0.0.0.0

### LINKED ISSUE #1735: When binding to 0.0.0.0, Pilosa should not use that as its advertised address
### Description

This is somewhat related to #1718, but can be accomplished without it. Currently binding to 0.0.0.0 means that Pilosa will set its URI to 0.0.0.0 and so no other hosts will be able to contact it on that address. We should make an effort to look up an actual IP address which Pilosa is listening on once it's bound and set that as the URI instead. This is a somewhat fraught issue in the case of multiple NICs, but anything we can do will be strictly better than the alternative of just using 0.0.0.0. I believe memberlist has code to do this already, so we could try to steal that.

It's possible to stand up a Pilosa cluster with each node binding to 0.0.0.0, so that it's possible to talk to each node on it's own localhost, but the cluster also works.

travisturner:
## Overview

This PR adds support for advertise address by using a new config
option `advertise-addr`, or by defaulting its value to that
specified in `listen-addr`. Also replaces `bind` with `listen-addr`.

Also adds support for listening on 0.0.0.0 by trying to determine
the preferred outbound IP to use for the advertise address.

Fixes #1735

TODO:
- [x] Update the docs if the changes here are agreed upon

## Pull request checklist

## Code review checklist

### DIFF
--- a/ctl/server.go
+++ b/ctl/server.go
@@ -26,6 +26,7 @@ func BuildServerFlags(cmd *cobra.Command, srv *server.Command) {
 	flags := cmd.Flags()
 	flags.StringVarP(&srv.Config.DataDir, "data-dir", "d", srv.Config.DataDir, "Directory to store pilosa data files.")
 	flags.StringVarP(&srv.Config.Bind, "bind", "b", srv.Config.Bind, "Default URI on which pilosa should listen.")
+	flags.StringVar(&srv.Config.Advertise, "advertise", srv.Config.Advertise, "Address to advertise externally.")
 	flags.IntVarP(&srv.Config.MaxWritesPerRequest, "max-writes-per-request", "", srv.Config.MaxWritesPerRequest, "Number of write commands per request.")
 	flags.StringVar(&srv.Config.LogPath, "log-path", srv.Config.LogPath, "Log path")
 	flags.BoolVar(&srv.Config.Verbose, "verbose", srv.Config.Verbose, "Enable verbose logging")
@@ -49,6 +50,9 @@ func BuildServerFlags(cmd *cobra.Command, srv *server.Command) {
 
 	// Gossip
 	flags.StringVarP(&srv.Config.Gossip.Port, "gossip.port", "", srv.Config.Gossip.Port, "Port to which pilosa should bind for internal state sharing.")
+	flags.StringVarP(&srv.Config.Gossip.AdvertiseHost, "gossip.advertise-host", "", srv.Config.Gossip.AdvertiseHost, "Host on which memberlist should advertise.")
+	flags.StringVarP(&srv.Config.Gossip.AdvertisePort, "gossip.advertise-port", "", srv.Config.Gossip.AdvertisePort, "Port on which memberlist should advertise.")
+
 	flags.StringSliceVarP(&srv.Config.Gossip.Seeds, "gossip.seeds", "", srv.Config.Gossip.Seeds, "Host with which to seed the gossip membership.")
 	flags.StringVarP(&srv.Config.Gossip.Key, "gossip.key", "", srv.Config.Gossip.Key, "The path to file of the encryption key for gossip. The contents of the file should be either 16, 24, or 32 bytes to select AES-128, AES-192, or AES-256.")
 	flags.DurationVarP((*time.Duration)(&srv.Config.Gossip.StreamTimeout), "gossip.stream-timeout", "", (time.Duration)(srv.Config.Gossip.StreamTimeout), "Timeout for establishing a stream connection with a remote node for a full state sync.")
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -61,7 +61,7 @@ The config file is in the [toml format](https://github.com/toml-lang/toml) and h
 
 #### Bind
 
-* Description: host:port on which the Pilosa server will listen for requests. Host defaults to localhost and port to 10101. If `bind` is set to `0.0.0.0` then Pilosa will listen on all addresses.
+* Description: host:port on which the Pilosa server will listen for requests. Host defaults to localhost and port to 10101. If `bind` is set to `0.0.0.0` then Pilosa will listen on all available interfaces.
 * Flag: `--bind="localhost:10101"`
 * Env: `PILOSA_BIND="localhost:10101"`
 * Config:
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -36,6 +36,17 @@ The config file is in the [toml format](https://github.com/toml-lang/toml) and h
 
 ### All Options
 
+#### Advertise
+
+* Description: Address advertised by the server to other nodes in the cluster and to clients via the `/status` endpoint. Host defaults to the IP address represented by `bind` and port to 10101. If `bind` is set to `0.0.0.0` and `advertise` is not specified, then Pilosa will try to determine a reasonable, external IP address to use for `advertise`.
+* Flag: `--advertise="192.168.1.100:10101"`
+* Env: `PILOSA_BIND="192.168.1.100:10101"`
+* Config:
+
+    ```toml
+    advertise = 192.168.1.100:10101
+    ```
+
 #### Anti Entropy Interval
 
 * Description: Interval at which the cluster will run its anti-entropy routine which ensures that all replicas of each fragment are in sync.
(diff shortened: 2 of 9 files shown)


server/config.go
```diff
@@ -150,3 +167,197 @@ func NewConfig() *Config {
...
+		} else {
+			advHost = listenHost
+		}
+	}
+	return advScheme, advHost, advPort, nil
+}
+
+// OutboundIP gets the preferred outbound ip of this machine.
+func OutboundIP() net.IP {
+	conn, err := net.Dial("udp", "8.8.8.8:80")
```
jaffee: what's the idea behind using 8.8.8.8:80 here? I guess Dial for udp doesn't actually contact the remote host, and we want to make sure to use an address which is external so that we have a good chance of getting an interface with an externally routable address. 

I played around with using 192.0.0.8 which is the "IPv4 dummy address", but I don't know that this is actually better - it produced the same result as 8.8.8.8:80 on my machine.

server/config.go
```diff
@@ -107,8 +123,9 @@ type Config struct {
 // NewConfig returns an instance of Config with default options.
 func NewConfig() *Config {
 	c := &Config{
-		DataDir:             "~/.pilosa",
-		Bind:                ":10101",
+		DataDir:    "~/.pilosa",
+		ListenAddr: ":10101",
+		// AdvertiseAddr: "",
```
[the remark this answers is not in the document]
travisturner: well, I think because this was a place where defaults are being defined, we were originally listing all attributes, along with the default value, and commenting out those lines which use the go default value. but it doesn't look like this pattern has been maintained, and this is the second time someone has asked about this, so i'll just remove them.

server/config.go
```diff
@@ -150,3 +167,197 @@ func NewConfig() *Config {
...
+// ValidateAddrs controls the address fields in the Config object
+// and fills in any blanks.
+// The addresses fields must be guaranteed by the caller to either be
+// completely empty, or have both a host part and a port part
+// separated by a colon. In the latter case either can be empty to
+// indicate it's left unspecified.
+func (cfg *Config) ValidateAddrs(ctx context.Context) error {
+	// TODO (2.0): Deprecate Bind.
+	if cfg.Bind != "" {
+		cfg.ListenAddr = cfg.Bind
```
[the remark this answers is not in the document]
travisturner: I agree with this.
I originally had:
```
    // TODO (2.0): Deprecate Bind.
    if cfg.Bind != "" && cfg.ListenAddr == "" {
        cfg.ListenAddr = cfg.Bind
    }
```
but the problem is that the default for `cfg.ListenAddr` gets applied before this line, so it's never empty. Let me see if I can move this deprecation check earlier in the process.

server/config.go
travisturner: I really don't know if there's an ideal way to do this. I tried using `net.Interfaces()`, but that just seems like a random choice as to which IP to return at that point. This method seemed clever, and it's returning the IP that would be used if it actually made a connection to `8.8.8.8`. But with that said, I think that target IP could be any external address. Really the only limitation is that it not be something local like `127.0.0.1`, in which case `Dial` will use a local route. Do you think changing that to `192.0.0.8` is more clear?

jaffee: I don't want anyone scanning the code to think that Pilosa needs to be able to hit 8.8.8.8 in order to work. I actually think that keeping it a known external address is probably likely to work in the most situations. I could see a case where someone screwed up a route entry and 192.0.0.0/8 gets routed to the wrong place.

So I guess let's just add a comment explaining what's going on?

tgruben: +1 on the comment.  My gut says that hack will not work without a real address.  The one test i guess we should try is to what happens when the network is totally private and unable to route to the public addresses.

jaffee: it should work @tgruben `net.Dial` with a "udp" network doesn't actually send any data over the network AFAICT. It just looks up the source IP it would use if it were to send a packet.

codysoyland: What if we used the gossip seed IP address? That would ensure it chose the correct network in the case that the NIC used to connect to the internet is not the NIC used for Pilosa. Not sure how hard it is to get that config value from this code.

server/config.go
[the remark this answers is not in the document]
travisturner: I removed all these commented lines.

ctl/server.go
```diff
@@ -26,6 +26,8 @@ func BuildServerFlags(cmd *cobra.Command, srv *server.Command) {
 	flags := cmd.Flags()
 	flags.StringVarP(&srv.Config.DataDir, "data-dir", "d", srv.Config.DataDir, "Directory to store pilosa data files.")
 	flags.StringVarP(&srv.Config.Bind, "bind", "b", srv.Config.Bind, "Default URI on which pilosa should listen.")
```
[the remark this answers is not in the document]
travisturner: `Bind` is no longer deprecated. Removed `ListenAddr` instead.

server/config.go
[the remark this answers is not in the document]
travisturner: No longer applies. I removed `ListenAddr`.

server/config.go
```diff
@@ -150,3 +167,197 @@ func NewConfig() *Config {
...
+	if err != nil {
+		return "", "", errors.Wrap(err, "looking up address")
+	}
+	return addr, port, nil
+}
+
+// LookupAddr resolves the given address/host to an IP address. If
+// multiple addresses are resolved, it returns the first IPv4 address
+// available if there is one, otherwise the first address.
+func LookupAddr(ctx context.Context, resolver *net.Resolver, host string) (string, error) {
```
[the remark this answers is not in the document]
travisturner: I moved all these to an internal test file and un-exported several functions.

travisturner:
@jaffee I rebased the latest changes into the original commit because a lot of it was undoing what had been done in that first commit, so keeping it as separate commits just added more clutter.

server/config.go
jaffee: in the case of a single node cluster, that *might* result in the node advertising localhost which the client would then pick up and try to use? I think the 8.8.8.8 approach is probably slightly safer, but neither is going to be perfect.

codysoyland: I don't think this configuration is relevant to a single-node cluster. Additionally, if you are configuring a local cluster (multiple processes on one machine), and you are configuring your gossip seed as a port on localhost, wouldn't you want to advertise on localhost?

jaffee: I'm envisioning a situation where someone is standing up a single node server. They set up bind to 0.0.0.0 and don't set up advertise. For some reason, they pass localhost as the only seed. 

When the client connects, it will do a `/status` request, and see only localhost which it will then try to use when doing imports and such.

travisturner: @codysoyland @jaffee keep in mind this lookup only applies if the `bind` address has been specified to be `0.0.0.0`. It will never happen as a default.

travisturner: @codysoyland can you describe the config situation that you're worried about?
Is it that someone would specify:
```
bind: "0.0.0.0"
advertise: ""
gossip-seeds: ["localhost:14000"]
```

travisturner: sorry, @jaffee I didn't see your response before posting this, but i think my config example described what your concern was.

codysoyland: Yeah I wasn't actually considering that the client uses the advertise url to optimize direct writes to shards. I'm not sure how that problem should be approached considering it doesn't work when Pilosa is behind a load balancer. Maybe 8.8.8.8 is the best solution but I feel like there has to be a better way.

jaffee: I'm fine with this as is - I think it's significantly better than the current state and shouldn't block merging. I do agree though that we could probably do even better, but I'm not 100% sure what that looks like.

docs/configuration.md
```diff
@@ -50,7 +61,7 @@ The config file is in the [toml format](https://github.com/toml-lang/toml) and h
 
 #### Bind
 
-* Description: host:port on which the Pilosa server will listen for requests. Host defaults to localhost and port to 10101.
+* Description: host:port on which the Pilosa server will listen for requests. Host defaults to localhost and port to 10101. If `bind` is set to `0.0.0.0` then Pilosa will listen on all addresses.
```
jaffee:
```suggestion
* Description: host:port on which the Pilosa server will listen for requests. Host defaults to localhost and port to 10101. If `bind` is set to `0.0.0.0` then Pilosa will listen on all available interfaces.
```

travisturner:
@jaffee this needs another approval after adding your commit

PR #1834, opened 2019-01-18
prevent deadlock in replication logic on reopening a store

### LINKED ISSUE #1833: sporadic deadlocks observed in testing on CI

### What's going wrong?

Deadlock

### What was expected?

Less deadlock

Run tests often. It looks like the failure is in the translate tests for ReassignPrimaryStore.

Various Linuxes, CI, etc.

seebs:
Depending on where in the replicate() loop you are when a
store is closed or reassigned, it's possible for it to deadlock.
The deadlock would be that replicate has just successfully read an
entry from your PrimaryTranslateStore.Reader, when a new
PrimaryTranslateStore event happens. Then handlePrimaryTranslateStore
grabs the mutex, signals that the replication handler should
close, and waits for the replication handler to close. Meanwhile,
the replicate() loop tries to grab the mutex... and deadlocks.

Solution: Make the replicate() loop part that needs the mutex
a goroutine that signals on a channel, so we can put it in a select
along with checking for the replicationClosing signal (or the
context terminating). If one of those happens, replicate()
terminates, allowing monitorReplication() to return, which
causes the anonymous function which called it to call
repWG.Done(), allowing handlePrimaryTranslateStore to continue
and eventually release the mutex. At some later point, appendEntry
succeeds or fails, dumps its result status in a buffered
channel, and exits, and the buffered channel is garbage collected.

This is way simpler than it sounds, but it took me a while
to figure out how simple it was.

## Overview

Fixes #1833 

## Pull request checklist

(No tests because this is an existing test.)

## Code review checklist

### DIFF
--- a/translate.go
+++ b/translate.go
@@ -372,7 +372,6 @@ func (s *TranslateFile) monitorReplication() {
 		if err := s.replicate(ctx); err != nil {
 			s.logger.Printf("pilosa: replication error: %s", err)
 		}
-
 		select {
 		case <-ctx.Done():
 			return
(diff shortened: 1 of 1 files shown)


seebs:
I think we don't need to, because the only thing waiting on replicate completing should be the thing which just closed the replicateClosing channel, but I agree that this feels overcomplicated.

PR #1839, opened 2019-01-23
added convenience function to efficiently calculate size of a roaring bitmap in bytes

### LINKED ISSUE #1838: Best way to get Roaring.Bitmap size?

### Description
To have a handle on memory footprint when working with Roaring.Bitmaps - I'm monitoring it's size. Wonder what is the best way to get it? I'm trying to use Bitmap.Info() and then calculate Alloc size of all containers - but this method does too many allocations of bitmap and container info structs - is there something else I can be using? or is it possible to improve way to get Bitmap size on hot path without as many allocations?

Little to no garbage when trying to get Bitmap size

tgruben:
Fixes #1838

### DIFF
--- a/roaring/roaring_test.go
+++ b/roaring/roaring_test.go
@@ -37,6 +37,29 @@ func TestContainerCount(t *testing.T) {
 		t.Fatalf("Count != CountRange\n")
 	}
 }
+func TestSize(t *testing.T) {
+	//array
+	a := roaring.NewFileBitmap(0, 65535, 131072)
+	if a.Size() != 6 {
+		t.Fatalf("Size in bytes incorrect \n")
+	}
+
+	//bitmap
+	b := roaring.NewFileBitmap()
+	for i:=uint64(0);i<2048;i++{
+		b.DirectAdd(i)
+	}
+
+	if b.Size() != 4096 {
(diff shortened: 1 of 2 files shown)


roaring/roaring_test.go
```diff
@@ -37,6 +37,29 @@ func TestContainerCount(t *testing.T) {
...
+		t.Fatalf("Size in bytes incorrect \n")
+	}
+
+	//bitmap
+	b := roaring.NewFileBitmap()
+	for i := uint64(0); i < 2048; i++ {
+		b.DirectAdd(i)
+	}
+
+	if b.Size() != 4096 {
```
travisturner: this is actually an `array` container (not a `bitmap` container), so it's still not checking `bitmap`.

roaring/roaring_test.go
```diff
@@ -37,6 +37,29 @@ func TestContainerCount(t *testing.T) {
...
+		t.Fatalf("Size in bytes incorrect \n")
+	}
+
+	//bitmap
+	b := roaring.NewFileBitmap()
+	for i := uint64(0); i < 4096; i++ {
+		b.DirectAdd(i)
+	}
+
+	if b.Size() != 8192 {
```
travisturner: still not correct. it's still an array.

PR #1851, opened 2019-01-30
add from/to range arguments to Rows() call

### LINKED ISSUE #1783: Rows call on date field w/o standard view don't work
### Expected behavior

I guess `Rows` needs to take a date range and then give back all the rows which have set bits in the views that fall in that date range.

### Actual behavior

`Rows` doesn't have any notion of date and always queries the standard view. Now that we allow disabling the standard view, it doesn't work.

Import into a date field with the standard view disabled and then execute a Rows query.

### Success Criteria (implementation notes)

Basically we need to implement logic for merging the Rows results from multiple different views. Then for time quantum fields with no standard view, we'll merge together all rows from all the top-level quantum views. (e.g. if it's a YMD field, we just need to merge the results from each year... not the MD views)

travisturner:
## Overview

This PR adds `from` and `to` arguments to the `Rows()` call.
If the field specified is type `time`, then the `from` and `to`
can be used to restrict the results.

Fixes #1783

## Pull request checklist

## Code review checklist

### DIFF
--- a/docs/query-language.md
+++ b/docs/query-language.md
@@ -836,8 +836,7 @@ result of the previous request will start from the next available row.
 
 If the field is of type `time`, the `from` and `to` arguments can be provided
 to restrict the result to a specific time span. If `from` and `to` are
-not provided, the field must be configured with `noStandardView` set to
-`false`.
+not provided, the full range of existing data will be queried.
 
 **Result Type:** Object with `"rows" or "keys" and an array of integers or strings respectively.`
 
--- a/docs/query-language.md
+++ b/docs/query-language.md
@@ -812,10 +812,12 @@ Options(Row(f1=10), shards=[0, 2])
 {"attrs":{},"columns":[100, 2097152]}
 ```
 
+#### Rows
+
 **Spec:**
 
 ```
-Rows(<FIELD>, previous=<UINT|STRING>, limit=<UINT>, column=<UINT|STRING>)
+Rows(<FIELD>, previous=<UINT|STRING>, limit=<UINT>, column=<UINT|STRING>, from=<TIMESTAMP>, to=<TIMESTAMP>)
 ```
 
 **Description:**
@@ -832,6 +834,10 @@ is given, the number of rowIDs returned will be less than or equal to
 result sets. Results are always ordered, so setting `previous` as the last
 result of the previous request will start from the next available row.
 
+If the field is of type `time`, the `from` and `to` arguments can be provided
+to restrict the result to a specific time span. If `from` and `to` are
+not provided, the field must be configured with `noStandardView` set to
+`false`.
 
 **Result Type:** Object with `"rows" or "keys" and an array of integers or strings respectively.`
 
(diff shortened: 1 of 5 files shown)


executor.go
```diff
@@ -1152,16 +1152,59 @@ func (e *executor) executeRowsShard(_ context.Context, index string, fieldName s
...
+				return nil, errors.New("Row() 'from' arg must be a timestamp")
+			}
+		}
 
-	frag := e.Holder.fragment(index, fieldName, viewStandard, shard)
-	if frag == nil {
-		return make(RowIDs, 0), nil
+		// Parse "to" time, if set.
+		var toTime time.Time
+		if _, ok := c.Args["to"]; ok {
```
yuce: Does it make sense to limit `toTime` to today + 1 day like for the `Row` query?

yuce: .. of course as long as `to` is not explicitly specified.

docs/query-language.md
```diff
@@ -832,6 +834,10 @@ is given, the number of rowIDs returned will be less than or equal to
 result sets. Results are always ordered, so setting `previous` as the last
 result of the previous request will start from the next available row.
 
+If the field is of type `time`, the `from` and `to` arguments can be provided
+to restrict the result to a specific time span. If `from` and `to` are
+not provided, the field must be configured with `noStandardView` set to
+`false`.
```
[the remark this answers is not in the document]
travisturner: We can do this, but as I mentioned above, I'm concerned about the number of views that will be considered. I guess the worst case would be number of hours from time 0 to time max (or time today+1). I think that about 438k hours. Or 18k days. Or 600 months. Or 50 years. Right?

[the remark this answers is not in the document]
travisturner: I guess it's just checking the views that it knows about in memory, so maybe it's not a concern.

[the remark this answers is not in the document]
travisturner: Update:
0/now -> 2019 years. 24,228 months. 737,424 days. 17,698,176 hours. 
It actually takes ~12s for `viewsByTimeRange()` to generate the list of hours.

And if we were to use a "max time", we'd need to decide what that is. Actually using the maximum possible time is a very large date and not reasonable.

executor.go
jaffee: sounds like we'd need to rewrite things such that `viewsByTimeRange` can take the actual available views into account rather than naively generating all possible views in the range for the quantum. 

I'm fine with *requiring* both `from` and `to` if that's too much work right now.

I'm not fine with defaulting to now+1 day which I think is arbitrary/surprising behavior that will cause bugs for people.

yuce: I believe most of the users would use a timestamp for past or current time, and 1 day is simply something to account for time zone differences. Did we have any use case where we had to use a future timestamp?

In any case, the best solution is using already existing views if we can, instead of iterating on all possible views.

travisturner: ok, the latest commit first determines the min/max possible time range values based on the available views. then, the default time range uses those min/max values.

travisturner:
@jaffee I had to add another commit to fix the docs since we changed the logic.

PR #1855, opened 2019-01-31
Validate shard-WIP

### LINKED ISSUE #1808: ignore shards which node does not own
### Description

in some situations, e.g. after a resize, a node may have shard files which do not belong to it. Normally the node will open all files in its data directory regardless of whether it needs to use them this can lead to an explosion of used file descriptors.

### Success criteria (What criteria will consider this ticket closeable?

Pilosa does not open data files which it does not need

tgruben:
## Overview
In order to provide for potential read-only clusters, we needed to be able to point a pilosa cluster at some shared space and only load the shards applicable to each node. Integeration testing is working, but some internal testing needs to be added.  Just wanting a bit of feedback before i continue

Fixes #1808


field.go
```diff
@@ -207,6 +208,9 @@ func newField(path, index, name string, opts FieldOption) (*Field, error) {
 
 		remoteAvailableShards: roaring.NewBitmap(),
 
+		shardValidator: func(uint64) bool {
```
travisturner: what do you think about defining a `defaultShardValidator` somewhere and using that in place of this default code in 3 places?

holder.go
```diff
@@ -77,6 +77,8 @@ type Holder struct {
 	// The interval at which the cached row ids are persisted to disk.
 	cacheFlushInterval time.Duration
 
+	shardValidatorfn func(index string, shard uint64) bool
```
travisturner: I think I'd rather see `shardValidatorFunc` instead of `...fn`

jaffee:
this was merged in #1900

PR #1859, opened 2019-02-11
add server stats to /info endpoint

### LINKED ISSUE #651: Add endpoint to get host/infrastructure info from Pilosa
### Description
It would be useful for benchmarks to be able to programmatically get a description of the hardware that Pilosa is running on. It seems like the easiest way to do this is to support it in Pilosa itself, since that's the only thing that the benchmarking tools are guaranteed to have access to.

This small library (no deps outside of stdlib) https://github.com/cloudfoundry/gosigar provides a cross platform way of getting system memory and other stats.

We could add a separate endpoint to Pilosa for this, or roll the info into `/status` or another pre-existing endpoint potentially. We could implement it so that querying a single node returned the data for the whole cluster, or just make the client query nodes one by one.

Benchmarking tools are able to automatically get a description of cluster infrastructure and add it to their output.

seebs:
report the approximate hardware specs (CPU speed, cores, memory)
of the server in the /info endpoint. This may be useful when
benchmarking.

Fixes #651

(note: docs haven't been changed, and maybe they should, but right now we don't actually document /info at all, apparently?)

## Pull request checklist

## Code review checklist

### DIFF
--- a/api.go
+++ b/api.go
@@ -1177,8 +1177,18 @@ func (api *API) Version() string {
 
 // Info returns information about this server instance
 func (api *API) Info() serverInfo {
+	si := api.server.systemInfo
+	// we don't report errors on failures to get this information
+	physicalCores, logicalCores, _ := si.CPUCores()
+	mhz, _ := si.CPUMHz()
+	mem, _ := si.MemTotal()
 	return serverInfo{
-		ShardWidth: ShardWidth,
+		ShardWidth:       ShardWidth,
+		CPUPhysicalCores: physicalCores,
+		CPULogicalCores:  logicalCores,
+		CPUMHz:           mhz,
+		CPUType:          si.CPUModel(),
+		Memory:           mem,
 	}
 }
 
@@ -1213,7 +1223,12 @@ func (api *API) TranslateKeys(body io.Reader) ([]byte, error) {
 }
 
 type serverInfo struct {
-	ShardWidth uint64 `json:"shardWidth"`
+	ShardWidth       uint64 `json:"shardWidth"`
+	Memory           uint64 `json:"memory"`
+	CPUType          string `json:"cpuType"`
+	CPUPhysicalCores int    `json:"cpuPhysicalCores"`
+	CPULogicalCores  int    `json:"cpuLogicalCores"`
+	CPUMHz           int    `json:"cpuMHz"`
 }
 
 type apiMethod int
--- a/gopsutil/systeminfo.go
+++ b/gopsutil/systeminfo.go
@@ -16,8 +16,10 @@ package gopsutil
 
 import (
 	"runtime"
+	"strings"
 
 	"github.com/pilosa/pilosa"
+	"github.com/shirou/gopsutil/cpu"
 	"github.com/shirou/gopsutil/host"
 	"github.com/shirou/gopsutil/mem"
 )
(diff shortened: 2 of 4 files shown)


travisturner:
My initial feedback is that this doesn't follow the pattern of keeping dependencies out of the `pilosa` package (and instead injecting them via the `server` package).

See [gcnotify](https://github.com/pilosa/pilosa/blob/master/server/server.go#L282) as an example.

I don't have a strong opinion about this, but we did a lot of work to clean up the `pilosa` package so it might be worth the additional work to continue that pattern.

seebs:
revised to use the existing `gopsutil` functionality we already had handy.

api.go
```diff
@@ -1203,7 +1213,12 @@ func (api *API) TranslateKeys(body io.Reader) ([]byte, error) {
 }
 
 type serverInfo struct {
-	ShardWidth uint64 `json:"shardWidth"`
+	ShardWidth       uint64 `json:"shardWidth"`
+	Memory           uint64 `json:"memory"`
+	CPUType          string `json:"cpuType"`
+	CPUPhysicalCores int
```
travisturner: Any reason to omit the `json` tags here? As it is, the case of the last three aren't consistent.

seebs: Is "inattention" a "reason"?

And yeah, that should be fixed.

seebs: On the cores: I read the comments for Linux, which said I'd get one entry per core, with "cores" = 1. Apparently on the Mac, it will come back with a single thing listing more cores... And what's bugging me is that I'm pretty sure the i5 is dual-core in physical cores, and has hyperthreading, so it *should* be claiming 4 "threads" (aka logical cores). Huh. Welp, back to the drawing board.

seebs: weirder still: "2900" is the *expected* value for MHz (or the one I'd have expected). on linux i think this one would report 3300.

seebs:
Hah. And I added the json field names... and that broke the test, which I apparently didn't rerun.

seebs:
it looks like that CI failure was just the ongoing occasional races in memberlist.

jaffee:
should we try to get this merged? needs a rebase

seebs:
I think so. It's useful, and imagine uses it if it's available.

seebs:
the patch applied cleanly to master, so let's see whether CI passes!

gopsutil/systeminfo.go
```diff
@@ -40,6 +46,55 @@ func (s *systemInfo) Uptime() (uptime uint64, err error) {
 	return hostInfo.Uptime, nil
 }
 
+// cpuFrequencyMultipliers is a lookup table for prefixes on clock speeds.
+// This is probably overkill.
+var cpuFrequencyMultipliers = [256]int{
+	'M': 1,
+	'G': 1000,
+	'T': 1000 * 1000,
+}
```
jaffee: I did not know you could do this... not sure how I feel about it.

jaffee:
Travis' issue seems to be fixed. At least... my mac reports correct numbers and ours our about the same age.

gopsutil/systeminfo.go
seebs: This is two syntactic sugar at once. One is that you can use labeled initializers in slice/array constants, just like in struct ones, the other is that you can treat runes as their unicode values.

Very convenient, though.

PR #1863, opened 2019-02-18
avoid probable race when creating fragments

### LINKED ISSUE #1810: flock unavailable while opening fragment on 1.2
Similar to #1669 but probably not the same root cause since it's happening on 1.2.

While importing:
2019/01/02 15:57:44 pilosa.go:195: starting field import for test_bsi_field: importing values to nodes: doing import: Server error 500 Internal Server Error body:’importing existence columns: creating view: opening fragment: opening storage: flock: resource temporarily unavailable

Disabling the existence field seems to stop the problem.

This is on a 5 node cluster running imports from the pdk.

seebs:
Revised this based on re-reading the code and trying to test it. It turns out that the reuse of old items in the fragments map is probably harmless, but there was an actual underlying race condition that could result in trying to reopen a fragment immediately after creating it, rather than using the just-opened fragment.

## Pull request checklist

## Code review checklist

### DIFF
--- a/view.go
+++ b/view.go
@@ -21,6 +21,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/pilosa/pilosa/logger"
 	"github.com/pilosa/pilosa/pql"
@@ -180,10 +181,8 @@ func (v *view) fragmentPath(shard uint64) string {
 func (v *view) Fragment(shard uint64) *fragment {
 	v.mu.RLock()
 	defer v.mu.RUnlock()
-	return v.fragment(shard)
+	return v.fragments[shard]
 }
-
-func (v *view) fragment(shard uint64) *fragment { return v.fragments[shard] }
 
 // allFragments returns a list of all fragments in the view.
 func (v *view) allFragments() []*fragment {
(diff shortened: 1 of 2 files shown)


seebs:
> Do you think it's worth adding a test to cover this scenario like we did in https://github.com/pilosa/pilosa/pull/1675/files ?

yes, will do

seebs:
@travisturner, I think this is probably ready now.

So when I made the (working) patch, I got stumped trying to test it -- any test which would delay suitably to manufacture the failure under the old code would deadlock under the new code, or vice versa. Then I realized I don't need an actual lock, I just need enough delay to make the scheduler likely to run both goroutines. 10ms later, here's a test which produces the expected (but unwanted) failure with the old CreateFragmentIfNotExists, but passes with the new one.

seebs:
`staticcheck` is wiser than i am, and points out that you shouldn't call t.Fatalf() from a goroutine. Fixed using errgroup.

view_internal_test.go
```diff
@@ -118,3 +145,15 @@ type errorBroadcaster struct {
...
 	return errors.New("intentional error")
 }
+
+// delayBroadcaster is a broadcaster which delays briefly.
+type delayBroadcaster struct {
+	nopBroadcaster
+	delay time.Duration
+}
+
+// SendSync is an implementation of Broadcaster SendSync which always returns an error.
```
travisturner: This comment doesn't describe the actual behavior.

seebs: "i have checked all of the comments carefully. yes, i even found the error in the comments i missed, so i fixed that too."

[narrator] They had not checked all of the comments.

seebs:
rebased and pushed. you'll be glad to know that, while i did indeed edit the comment on errorBroadcaster to say it delays, while leaving the comment on delayBroadcaster alone, I fixed that before committing.

view_internal_test.go
travisturner: "nice narration"

[narrator] he secretly—and jealously—wished he, too, had a narrator.

seebs:
hmm, interesting question. i wonder... say two nodes simultaneously create shards, and notify each other. do they have to wait on any processing on the remote end before they can finish/ack the message, in some way? because if they did, if they were doing shards of the same views, they could end up each having the view locked, and unable to do things to it based on the other end's message, for instance.

i'll study this more tomorrow, but i think the existence of that attempt to fix a deadlock implies something.

possibly a fix would be to add the fragment to v.fragments whether or not the broadcast succeeds; after all, if we don't, the next time we shut down, we still end up coming up with that fragment defined, and never broadcast it.

jaffee:
I've combed through this a few times now, and I'm pretty well convinced that this particular instance won't cause a deadlock, but I do think that having something locked, and then talking to other nodes in such a way that they require locking is something to be avoided if possible.

If you'd like to discuss the previous deadlocks and how they worked, lmk

view_internal_test.go
```diff
@@ -78,43 +78,39 @@ func TestView_DeleteFragment(t *testing.T) {
...
+	err := creates.Wait()
 	if err != nil {
-		t.Fatal(err)
+		t.Fatalf("unexpected error: %v", err)
 	}
 }
 
-// errorBroadcaster is a broadcaster which always returns an error.
-type errorBroadcaster struct {
+// delayBroadcaster is a broadcaster which delays briefly.
```
jaffee: total nitpick... it's not really briefly right? "delayBroadcaster is a nopBroadcaster with a configurable delay."?

seebs: that is also true

view_internal_test.go
```diff
@@ -78,43 +78,39 @@ func TestView_DeleteFragment(t *testing.T) {
 }
 
 // Ensure view closes fragment after failed shard broadcast.
-func TestView_CreateFragmentError(t *testing.T) {
+func TestView_CreateFragmentRace(t *testing.T) {
+	var creates errgroup.Group
 	v := mustOpenView("i", "f", "v")
 	defer v.close()
 
 	// Use a broadcaster which intentionally fails.
```
jaffee: this comment is no longer correct

view_internal_test.go
```diff
@@ -78,43 +78,39 @@ func TestView_DeleteFragment(t *testing.T) {
...
+type delayBroadcaster struct {
 	nopBroadcaster
+	delay time.Duration
 }
 
-// SendSync is an implementation of Broadcaster SendSync which always returns an error.
-func (errorBroadcaster) SendSync(Message) error {
-	return errors.New("intentional error")
+// SendSync is an implementation of Broadcaster SendSync which delays for a
+// specified interval before succeeding.
```
jaffee: (without actually doing anything)

view.go
```diff
@@ -206,45 +205,45 @@ func (v *view) recalculateCaches() {
...
+			v.logger.Printf("broadcasting create shard: %v", err)
+		}
+		close(broadcastChan)
+	}()
+
+	// We want to wait until the broadcast is complete, but what if it
+	// takes a really long time? So we time out.
+	select {
+	case <-broadcastChan:
+	case <-time.After(50 * time.Millisecond):
```
jaffee: consider logging a warning here? sendSync is taking a long time, other nodes may not know about this shard, results may not be consistent until it finishes... etc

seebs: I'd be a bit concerned about it being spammy; I don't know how often we'd expect that to happen, or how often new shards get created. I'd say maybe... Make a version that does that and do some testing? But I'm not sure what to test it on, for workloads.

jaffee: make it a logger.Debug?

seebs:
Have fixed the two comments. I think for now I'm inclined not to report the broadcast errors, but that's a revisitable decision. I am also pondering the implications of this for a logging strategy to coalesce similar-and-repeated messages.

PR #1865, opened 2019-02-18
removed copy for pilosa roaring files

### LINKED ISSUE #1850: roaring.Bitmap.UnmarshalBinary - streaming variant

### Description
Right now to unmarshal bitmap from file requires to read it all into memory - that proves to be the highest memory consumer when working with big bitmaps. Would be good if it would be possible to unmarshal binary from Reader
No need to read entire bitmap file into memory to unmarshal it

tgruben:
## Overview

Due to the fact that pilosa roaring files do not mutate the internal bitmaps we can avoid a costly copy on import.   
Fixes #1850

### DIFF
--- a/api.go
+++ b/api.go
@@ -323,21 +323,21 @@ func (api *API) ImportRoaring(ctx context.Context, indexName, fieldName string, 
 					if len(viewData) == 0 {
 						return fmt.Errorf("no data to import for view: %s", viewName)
 					}
-					// must make a copy of data to operate on locally.
-					// field.importRoaring changes data
 					fileMagic := uint32(binary.LittleEndian.Uint16(viewData[0:2]))
-					if fileMagic == roaring.MagicNumber { // if pilosa roaring
+					if fileMagic == roaring.MagicNumber { // if pilosa roaring format
 						err = field.importRoaring(viewData, shard, viewName, req.Clear)
 						if err != nil {
-							return err
+							return errors.Wrap(err,"import pilosa roaring")
 						}
 
 					} else {
+					// must make a copy of data to operate on locally on standard roaring format.
+					// field.importRoaring changes the standard roaring run format to pilosa roaring
 						data := make([]byte, len(viewData))
 						copy(data, viewData)
 						err = field.importRoaring(data, shard, viewName, req.Clear)
 						if err != nil {
-							return err
+							return errors.Wrap(err,"import standard roaring")
 						}
 					}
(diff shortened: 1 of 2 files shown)


api.go
```diff
@@ -324,11 +325,20 @@ func (api *API) ImportRoaring(ctx context.Context, indexName, fieldName string,
...
-						return err
+					fileMagic := uint32(binary.LittleEndian.Uint16(viewData[0:2]))
+					if fileMagic == roaring.MagicNumber { // if pilosa roaring
+						err = field.importRoaring(viewData, shard, viewName, req.Clear)
+						if err != nil {
+							return err
+						}
+
+					} else {
+						data := make([]byte, len(viewData))
```
jaffee: let's move the comment from above down here and edit it a bit to explain what modifies the data

travisturner: any benefit in wrapping the errors too?

jaffee: definitely... good call

api.go
```diff
@@ -322,13 +323,22 @@ func (api *API) ImportRoaring(ctx context.Context, indexName, fieldName string,
...
-					data := make([]byte, len(viewData))
-					copy(data, viewData)
-					err = field.importRoaring(data, shard, viewName, req.Clear)
-					if err != nil {
-						return err
+					fileMagic := uint32(binary.LittleEndian.Uint16(viewData[0:2]))
+					if fileMagic == roaring.MagicNumber { // if pilosa roaring format
+						err = field.importRoaring(viewData, shard, viewName, req.Clear)
+						if err != nil {
+							return errors.Wrap(err, "import pilosa roaring")
```
travisturner:
```suggestion
							return errors.Wrap(err, "importing pilosa roaring")
```

api.go
```diff
@@ -322,13 +323,22 @@ func (api *API) ImportRoaring(ctx context.Context, indexName, fieldName string,
...
+						}
+
+					} else {
+						// must make a copy of data to operate on locally on standard roaring format.
+						// field.importRoaring changes the standard roaring run format to pilosa roaring
+						data := make([]byte, len(viewData))
+						copy(data, viewData)
+						err = field.importRoaring(data, shard, viewName, req.Clear)
+						if err != nil {
+							return errors.Wrap(err, "import standard roaring")
```
travisturner:
```suggestion
							return errors.Wrap(err, "importing standard roaring")
```

PR #1870, opened 2019-02-20
only use intersectionCount on single field groupBy

### LINKED ISSUE #1866: GroupBy on single field with filter is slow (intersect vs intersectionCount)
### Description

`GroupBy(Rows(field=some_field_with_a_million_rows), filter=Row(blah=2))` is very slow. However, the roughly equivalent query: 
`GroupBy(Rows(field=blah, previous=1, limit=1), Rows(field=some_field_with_a_million_rows))` is quite fast. 

This is because GroupBy always Intersects the filter with each Row in the first `Rows` query passed to it, and then does intersectionCount with each row in the last `Rows` query.

In the case of a single `Rows` query being passed with a filter, the GroupBy logic should use intersectionCount to combine the filter with each Row from the `Rows` query.

The two example queries should be roughly equivalent in performance (the performance of the second query)

tgruben:
## Overview
The special case of a single field groupby does not require a resultant bitmap so only the more performant intersectionCount needs to be used.

Fixes #1866

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -2856,7 +2856,10 @@ func newGroupByIterator(rowIDs []RowIDs, children []*pql.Call, filter *Row, inde
 	var fieldName string
 	var ok bool
 	ignorePrev := false
+	maxChild := 0
+
 	for i, call := range children {
+		maxChild = i
 		if fieldName, ok = call.Args["_field"].(string); !ok {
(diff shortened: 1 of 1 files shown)


tgruben:
I totally read this as a different issue.  Gonna have to think a bit now

PR #1871, opened 2019-02-22
1864 random import perf

### LINKED ISSUE #1864: Improve import performance when data is spread over many shards
### Description

Some Pilosa ingest workloads basically go record by record where each record gets an increasing column ID in Pilosa, and records don't get updated often (or at all), so that once a shard gets written out it doesn't get touched again. This is essentially the easy case - the client can buffer up a shard at a time, and write it all out at once, resulting in each fragment being snapshotted to disk a single time.

The case which is more difficult is if updates are happening randomly across the record space. Each time a single bit in any shard gets touched, the entire fragment which that bit is in needs to be snapshotted to disk (after some buffering period), which means that in many cases, snapshotting may occur with very few actual changes to the data. Pilosa nodes end up saturating disk I/O and import performance slows dramatically.

The proposed fix is to do something similar to the op log which is currently only used if one is setting bits through PQL. This is how we supported random updates at Umbel, but the overhead of PQL is not ideal. The basic idea is that as new writes come in, we can update things in memory which is quite cheap, and append the raw writes to a log so that they are durably persisted without much overhead. After some duration, or once a certain number of writes are outstanding, we can snapshot the fragment and clear the log.

Rough steps to implement this for imports are as follows: 
 1. As an import comes in to a fragment - decide whether to snapshot or not based on size of import and size of existing data.
2. If snapshotting - follow existing logic.
3. If not snapshotting, follow existing logic for doing in-memory updates, write the import data to some kind of append only log, update some kind counter to track how much updated data is in-memory, but not snapshotted.
4. snapshot on close
5. recovery logic for new append only log - e.g. on open, is there an op log, if so replay it.

Ingest workloads which have random writes scattered across the column space are significantly more performant.

jaffee:
## Overview

Fixes  #1864

For all imports except roaring imports, we build two slices of bitmap positions - one to set and one to clear. We reuse existing slice arguments when possible to avoid extra allocations.

After building these two slices, we call fragment.importPositions which decides whether or not this import should cause a snapshot or write the changed bits to the op log after applying them in-memory. This decision is based on the values of fragment.opN, fragment.MaxOpN, and the number of bits to be set and cleared.

Note: value imports don't use importPositions for the snapshotting path.

For bulkImportStandard, we no longer build a separate bitmap and Union it in. Instead we've created new roaring bitmap methods which are better optimized for bulk insertion. These methods keep track of the last container used and reuse in the event that the next bit to set has the same highbits - this avoids the overhead of looking up the same container repeatedly in some common cases. Additionally, we've added new operation types to the op log which write a batch of values all at once rather than having each value be a new operation which is separately written to disk.

We've also pulled btree Containers out of enterprise and directly into roaring. This was initially motivated by needing to generate large test cases for benchmarking more quickly, but is also helpful for overall performance in most cases.

This PR adds a significant amount of new API to roaring for performance reasons (and deprecates some old).

A number of benchmarks have been added or updated.

## Pull request checklist

## Code review checklist

### DIFF
--- a/fragment.go
+++ b/fragment.go
@@ -1468,23 +1468,6 @@ func (f *fragment) bulkImport(rowIDs, columnIDs []uint64, options *ImportOptions
 // bulkImportStandard performs a bulk import on a standard fragment. May mutate
 // its rowIDs and columnIDs arguments.
 func (f *fragment) bulkImportStandard(rowIDs, columnIDs []uint64, options *ImportOptions) (err error) {
-	var localBitmap *roaring.Bitmap
-	smallWrite := false
-	f.mu.Lock()
-	defer f.mu.Unlock() // TODO the big import path doesn't need to acquire the lock this soon.
-	if len(columnIDs)+f.opN < f.MaxOpN {
-		smallWrite = true
-		// If the number of operations is small, we'll write directly to local storage
-		localBitmap = f.storage
-	} else {
-		// Create a temporary bitmap which will be populated by rowIDs and columnIDs
-		// and then merged into the existing fragment's bitmap.
-		localBitmap = roaring.NewBitmap()
-
-		// Disconnect op writer so we don't append updates.
-		localBitmap.OpWriter = nil
-	}
-
 	// rowSet maintains the set of rowIDs present in this import. It allows the
 	// cache to be updated once per row, instead of once per bit. TODO: consider
 	// sorting by rowID/columnID first and avoiding the map allocation here. (we
@@ -1508,36 +1491,83 @@ func (f *fragment) bulkImportStandard(rowIDs, columnIDs []uint64, options *Impor
 		}
 	}
 	positions := columnIDs
-	var changedN int
+	f.mu.Lock()
+	defer f.mu.Unlock()
 	if options.Clear {
-		changedN, err = localBitmap.RemoveN(positions...) // TODO benchmark AddN behavior with sorted/unsorted positions
+		err = f.importPositions(nil, positions, rowSet)
 	} else {
-		changedN, err = localBitmap.AddN(positions...) // TODO benchmark RemoveN behavior with sorted/unsorted positions
+		err = f.importPositions(positions, nil, rowSet)
 	}
+	return errors.Wrap(err, "bulkImportStandard")
+}
+
+// importPositions takes slices of positions within the fragment to set and
+// clear in storage. One must also pass in the set of unique rows which are
+// affected by the set and clear operations. It is unprotected (f.mu must be
+// locked when calling it). No position should appear in both set and clear.
+//
+// importPositions tries to intelligently decide whether or not to do a full
+// snapshot of the fragment or just do in-memory updates while appending
+// operations to the op log.
+func (f *fragment) importPositions(set, clear []uint64, rowSet map[uint64]struct{}) error {
+	var setBitmap *roaring.Bitmap
+	var clearBitmap *roaring.Bitmap
+	smallWrite := false
+	if len(set)+len(clear)+f.opN < f.MaxOpN {
+		smallWrite = true
+		// If the number of operations is small, we'll write directly to local storage
+		setBitmap = f.storage
+		clearBitmap = f.storage
+	} else {
+		// Create a temporary bitmap which will be populated by rowIDs and columnIDs
+		// and then merged into the existing fragment's bitmap.
+		setBitmap = roaring.NewBitmap()
+		clearBitmap = roaring.NewBitmap()
+
+		// Disconnect op writer so we don't append updates.
+		setBitmap.OpWriter = nil
+		clearBitmap.OpWriter = nil
+	}
+
+	changedN, err := setBitmap.AddN(set...) // TODO benchmark Add/RemoveN behavior with sorted/unsorted positions
 	if err != nil {
 		return errors.Wrap(err, "adding positions")
 	}
-	f.stats.Count("ImportBit", int64(changedN), 1)
-	f.opN += changedN
+
+	if smallWrite {
+		f.stats.Count("ImportBit", int64(changedN), 1)
+		f.opN += changedN
+		changedN, err = clearBitmap.RemoveN(clear...)
+		f.stats.Count("ClearBit", int64(changedN), 1)
+		f.opN += changedN
+	} else {
+		// when doing big imports, we're going to difference the local bitmap,
+		// so we need to add rather than removing. TODO: figure out why we
+		// aren't adding/removing directly from storage.
+		_, err = clearBitmap.AddN(clear...)
+	}
+	if err != nil {
+		return errors.Wrap(err, "clearing positions")
+	}
 
 	var results *roaring.Bitmap
-	if !smallWrite {
-		// Merge localBitmap into fragment's existing data.
-		if options.Clear {
-			if f.storage.Count() > 0 {
-				results = f.storage.Difference(localBitmap)
-			} else {
-				results = roaring.NewBitmap()
-			}
-		} else {
-			if f.storage.Count() > 0 {
-				results = f.storage.Union(localBitmap)
-			} else {
-				results = localBitmap
-			}
-		}
-	} else {
+	if smallWrite {
 		results = f.storage
+	} else if beforeCnt := f.storage.Count(); !smallWrite && beforeCnt > 0 {
+		// Merge localBitmap into fragment's existing data.
+		if len(clear) > 0 {
+			f.storage = f.storage.Difference(clearBitmap)
+			nc := f.storage.Count()
+			f.stats.Count("ClearBit", int64(beforeCnt-nc), 1)
+			beforeCnt = nc
+			results = f.storage
+		}
+		if len(set) > 0 {
+			results = f.storage.Union(setBitmap) // TODO replace with UnionInPlace after https://github.com/pilosa/pilosa/issues/1875
+			f.stats.Count("ImportBit", int64(beforeCnt-f.storage.Count()), 1)
+		}
+	} else { // !smallWrite and nothing in storage
+		results = setBitmap
 	}
 
 	// Update cache counts for all affected rows.
--- a/fragment.go
+++ b/fragment.go
@@ -350,29 +350,40 @@ func (f *fragment) row(rowID uint64) *Row {
 	return f.unprotectedRow(rowID)
 }
 
+// unprotectedRow returns a row from the row cache if available or from storage
+// (updating the cache).
 func (f *fragment) unprotectedRow(rowID uint64) *Row {
 	r, ok := f.rowCache.Fetch(rowID)
 	if ok && r != nil {
 		return r
 	}
 
+	row := f.rowFromStorage(rowID)
+	f.rowCache.Add(rowID, row)
+	return row
+}
+
+// rowFromStorage clones a row data out of fragment storage and returns it as a
+// Row object.
+func (f *fragment) rowFromStorage(rowID uint64) *Row {
 	// Only use a subset of the containers.
 	// NOTE: The start & end ranges must be divisible by container width.
 	data := f.storage.OffsetRange(f.shard*ShardWidth, rowID*ShardWidth, (rowID+1)*ShardWidth)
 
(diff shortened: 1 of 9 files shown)


fragment_internal_test.go
```diff
@@ -713,6 +713,58 @@ func BenchmarkFragment_ImportValue(b *testing.B) {
...
+// the number of bits in the import, the number of rows in the fragment (which
+// is a pretty good proxy for fragment size on disk), and the MaxOpN on the
+// fragment which controls how many set bits occur before a snapshot is done. If
+// the number of bits in a given import is greater than MaxOpN, bulkImport will
+// always go through the standard snapshotting import path.
+func BenchmarkFragment_RepeatedSmallImports(b *testing.B) {
+	for _, numUpdates := range []int{100} {
+		for _, bitsPerUpdate := range []int{100, 1000} {
+			for _, numRows := range []int{1000, 100000, 1000000} {
+				// build the update data set all at once - this will get applied
```
seebs: hmm. suggestion: possibly move the creation of the slices out (we could just do a 100*1000 slice for each outside the loop, and reslice them at the beginning of the loop), and b.StopTimer()/b.StartTimer() around the population of the bits, just so they aren't being counted. Although I guess that doesn't matter as much since there's a b.Run() below?

seebs: nevermind just noticed this is changed by next commit.

fragment.go
```diff
@@ -1454,88 +1465,79 @@ func (f *fragment) bulkImport(rowIDs, columnIDs []uint64, options *ImportOptions
...
-		if err != nil {
-			return err
-		}
-
-		// Reduce the StatsD rate for high volume stats
-		f.stats.Count("ImportBit", 1, 0.0001)
+		columnIDs[i] = pos
 
 		// Add row to rowSet.
 		if i == 0 || rowID != lastRowID {
```
seebs: you could also use a lastRowID which can't validly occur -- I think doing lastRowID-- after creating it will make it a value that i don't think can actually meaningfully work in our system, and that simplifies the test here.

fragment.go
```diff
@@ -1454,88 +1465,79 @@ func (f *fragment) bulkImport(rowIDs, columnIDs []uint64, options *ImportOptions
...
-
-	// Merge localBitmap into fragment's existing data.
 	var results *roaring.Bitmap
-	if options.Clear {
-		if f.storage.Count() > 0 {
-			results = f.storage.Difference(localBitmap)
+	if !smallWrite {
+		// Merge localBitmap into fragment's existing data.
+		if options.Clear {
+			if f.storage.Count() > 0 {
```
seebs: I wonder if we should have a short-circuit `Any()` which we can use for cases where we just care whether the count is 0 or not-0.

fragment.go
```diff
@@ -1645,23 +1657,36 @@ func (f *fragment) bulkImportMutex(rowIDs, columnIDs []uint64) error {
...
 // importValue bulk imports a set of range-encoded values.
 func (f *fragment) importValue(columnIDs, values []uint64, bitDepth uint, clear bool) error {
-	f.mu.Lock()
-	defer f.mu.Unlock()
 	// Verify that there are an equal number of column ids and values.
 	if len(columnIDs) != len(values) {
 		return fmt.Errorf("mismatch of column/value len: %d != %d", len(columnIDs), len(values))
 	}
+	f.mu.RLock()
+	smallPath := false
```
seebs: would sorta prefer smallWrite just for naming consistency.

fragment.go
```diff
@@ -350,30 +350,41 @@ func (f *fragment) row(rowID uint64) *Row {
 	return f.unprotectedRow(rowID)
 }
 
+// unprotectedRow returns a row from the row cache if available or from storage
+// (updating the cache).
```
seebs: I'm curious why this didn't previously grab rows from storage. Is this a performance improvement, or a bug fix, or ??.

fragment.go
```diff
@@ -1508,36 +1491,83 @@ func (f *fragment) bulkImportStandard(rowIDs, columnIDs []uint64, options *Impor
...
+		_, err = clearBitmap.AddN(clear...)
+	}
+	if err != nil {
+		return errors.Wrap(err, "clearing positions")
+	}
+
+	var results *roaring.Bitmap
+	if smallWrite {
 		results = f.storage
+	} else if beforeCnt := f.storage.Count(); !smallWrite && beforeCnt > 0 {
```
seebs: I don't think we need the explicit `!smallWrite` here because this is the else from a test on smallWrite.

See also the previous question about an Any() operation.

fragment.go
```diff
@@ -1708,14 +1750,15 @@ func (f *fragment) importValue(columnIDs, values []uint64, bitDepth uint, clear
...
-			} else {
-				_, err = f.importSetValue(columnID, bitDepth, value, clear)
-			}
-			if err != nil {
+				ts, tc, err := f.positionsForValue(columnID, bitDepth, value, clear)
+				if err != nil {
+					return errors.Wrap(err, "getting positions for value")
+				}
+				toSet = append(toSet, ts...)
+				toClear = append(toClear, tc...)
```
seebs: I don't have a really better idea yet, but it feels to me like this is a lot of allocation-and-appending going on, and it might make more sense to build this more like `append`, something like:

`toSet, toClear, err = f.positionsForValue(columnID, bitDepth, value, clear, toSet, toClear)`

I'd also suggest making them both slightly larger than 50% of the size -- as is, we can guess at least one of them will need to grow, even if we don't know which one.

fragment.go
jaffee: there's actually no functional change here. I broke out the part of this method that copied the row out of storage so that I could use it elsewhere.

I have some thoughts around getting rid of the rowCache entirely, but not fully baked yet.

fragment.go
jaffee: I went ahead and did 5/8ths for the initial alloc. I think you're right about making it append style... will do that next.

fragment.go
```diff
@@ -1454,65 +1503,109 @@ func (f *fragment) bulkImport(rowIDs, columnIDs []uint64, options *ImportOptions
...
-		if f.storage.Count() > 0 {
-			results = f.storage.Union(localBitmap)
-		} else {
-			results = localBitmap
+		if len(set) > 0 {
+			results = f.storage.Union(setBitmap) // TODO replace with UnionInPlace after https://github.com/pilosa/pilosa/issues/1875
+			f.stats.Count("ImportBit", int64(beforeCnt-f.storage.Count()), 1)
 		}
+	} else { // !smallWrite and nothing in storage
+		results = setBitmap
```
seebs: what happens (or should happen) if the same column is present in both set and clear? if nothing else, maybe the `clearBitmap.AddN(clear...)` should be moved into the `len(clear) > 0` case, rather than happening up above on 1585.

fragment.go
```diff
@@ -1454,65 +1503,109 @@ func (f *fragment) bulkImport(rowIDs, columnIDs []uint64, options *ImportOptions
...
+
+// importPositions takes slices of positions within the fragment to set and
+// clear in storage. One must also pass in the set of unique rows which are
+// affected by the set and clear operations. It is unprotected (f.mu must be
+// locked when calling it). No position should appear in both set and clear.
+//
+// importPositions tries to intelligently decide whether or not to do a full
+// snapshot of the fragment or just do in-memory updates while appending
+// operations to the op log.
+func (f *fragment) importPositions(set, clear []uint64, rowSet map[uint64]struct{}) error {
```
seebs: reading this through again, i feel like these two paths are different enough that it might be cleaner/simpler to have them be two separate code flows. going back and forth, realizing that in smallWrite all of setBitmap, clearBitmap, and results end up being f.storage, and re-reading the logic to see why we sometimes have clearBitmap.AddN and sometimes RemoveN, seems complex. that or it's late in the day and i'm slow now.

jaffee: nah, you're totally right, it's a mess. I'll see about breaking it apart.

roaring/roaring.go
```diff
@@ -243,25 +243,25 @@ func (b *Bitmap) RemoveN(a ...uint64) (changed int, err error) {
...
+			a[changed] = v
+			changed++
+		}
+	}
+
 	if b.OpWriter != nil {
 		op := &op{
 			typ:    opTypeRemoveBatch,
-			values: a,
+			values: a[:changed],
```
seebs: hmm. okay, so, this creates a subtle semantic change: previously, if the op log write failed, we stopped *without* changing in-memory data. Now we'll be changing the in-memory data, but returning an error and a count of 0. I don't know whether that's a problem, but it feels not-ideal?

fragment.go
```diff
@@ -1668,11 +1671,15 @@ func (f *fragment) bulkImportMutex(rowIDs, columnIDs []uint64) error {
...
 		}
-		rowIDs[setIdx] = pos
-		setIdx++
+		colSet[columnID] = pos
 		rowSet[rowID] = struct{}{}
 	}
-	toSet := rowIDs[:setIdx]
+	i := 0
+	for _, pos := range colSet {
+		rowIDs[i] = pos
```
seebs: this confused the heck out of me for a while but i think i follow it. i have a vague feeling of impending doom but i think it's actually correct. only thing i notice is that if you had multiple writes to a column, you'd end up generating multiple clearbits of the same existing bit for it, but that doesn't break anything, and it's probably cheaper than setting up a thing parallel to the colSet tracking.

roaring/roaring.go
jaffee: great point... I felt kind of bad about it, but couldn't think of a good reason why.

What if we went back and unset/reset the bits after a write fail?

fragment.go
```diff
@@ -1596,7 +1596,7 @@ func (f *fragment) importPositions(set, clear []uint64, rowSet map[uint64]struct
 	var results *roaring.Bitmap
 	if smallWrite {
 		results = f.storage
-	} else if beforeCnt := f.storage.Count(); !smallWrite && beforeCnt > 0 {
+	} else if beforeCnt := f.storage.Count(); beforeCnt > 0 {
```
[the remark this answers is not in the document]
jaffee: we use beforeCnt later, but this whole thing probably isn't worth it just for the stat tracking purposes. I'm also thinking there is a more efficient way to implement AddN that means we don't need to build a separate bitmap and union it in which would then be able to give us the number of changed bits as a byproduct.

roaring/roaring.go
seebs: Hmm. We do have a record of what we changed, and we know whether they were add/remove operations, so we could in fact do that fairly easily -- just do the other operation to the changed values only. Sure, that seems sane.

fragment.go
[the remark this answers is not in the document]
seebs: Hmm. Maybe? I am inclined to leave it alone for now and maybe have a separate pass later for "check for places where we perform a count we don't need".

fragment.go
jaffee: much cleaner now :)

fragment.go
jaffee: the doc string for this method indicates that no position should appear in both set and clear. With the new logic, I think it would end up getting cleared, but I think it's best we leave it specifically undefined.

jaffee:
if you get a chance @seebs I made a few pretty trivial tweaks since your last review

roaring/roaring_test.go
```diff
@@ -1457,64 +1457,87 @@ const (
...
 }
 
 func BenchmarkContainerReverse(b *testing.B) {
-	for n := 0; n < b.N; n++ {
-		bm := roaring.NewFileBitmap()
-		for row := NumRows - 1; row >= 1; row-- {
-			for col := NumColums - 1; col >= 1; col-- {
-				bm.Add(row*pilosa.ShardWidth + (col * MaxContainerVal))
+	for i, bmMaker := range bmFuncs {
+		b.Run(fmt.Sprintf("%v", bmFuncNames[i]), func(b *testing.B) {
```
seebs: `fmt.Sprintf("%v", bmFuncNames[i])` => `bmFuncNames[i]`

jaffee: yep... unnecessary sprintfs are not our friend

PR #1882, opened 2019-03-01
wip cache generated timeview strings

tgruben:
## Overview

based on some profiling, almost half the cpu time associated with time imports is related to generating the view path strings.  this pr attempts to mitigate that cost a bit with a cache

Fixes #

### DIFF
--- a/time.go
+++ b/time.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


jaffee:
I think this is looking fairly reasonable - I would like to see a before/after benchmark though since this is doing a lot of granular locking and has to allocate the unitKey struct each time.

tgruben:
Not huge, but better anyway, i guess an around 8%

BenchmarkTImportNoCache-12    	      10	 182980616 ns/op	106356676 B/op	  190632 allocs/op
BenchmarkTImportCached-12     	      10	 169199810 ns/op	98846802 B/op	    3572 allocs/op

tgruben:
Ugh..allocations do have their perks.   I'm see what i can do, maybe we just build the string without format function

PR #1891, opened 2019-03-07
Lower filehandles

### LINKED ISSUE #1888: Don't fail when mmaps are exhausted
### Description

With lots of fragments - especially when time quantums are being used, we see errors like `creating fragment: opening fragment: opening storage: mmap: cannot allocate memory` this could be due to the actual memory available being insufficient, but it can also be due to vm.max_map_count being too low which can be queried by `sysctl vm.max_map_count`. (This is in `fragment.openStorage`)

In case of mmap failing, it should be possible to fall back to reading the file into memory normally (e.g. ioutil.ReadAll), and proceeding from there as usual. Some state will need to be kept so that when the fragment is closed it doesn't try to unmap the file.

Note:
There is a lot of potential follow-on work to this - handling file handle exhaustion more gracefully, keeping only "hot" fragments mapped or in-memory, while others get read in on-demand, etc. This ticket is not for that work, though we should keep it in mind.

When max_map_count is reached and the mmap call throws an error, Pilosa should log a warning and then continue by reading the file into memory the old fashioned way. Insufficient max_map_count should not cause Pilosa to misbehave or fail to make progress.

tgruben:
## Overview

In an effort to reduce open file handles, this close files after unmarshalling and opens the only on set/clear 
Fixes #1888

### DIFF
--- a/fragment.go
+++ b/fragment.go
@@ -130,7 +130,8 @@ type fragment struct {
 	// existing value (to clear) prior to setting a new value.
 	mutexVector vector
 
-	stats stats.StatsClient
+	stats    stats.StatsClient
+	nonemmap bool
 }
 
 // newFragment returns a new instance of Fragment.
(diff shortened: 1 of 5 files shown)


jaffee:
closing in favor of #1905 and #1903 — a number of these commits are incorporated.

PR #1897, opened 2019-03-12
Address UnionInPlace performance regressions

### LINKED ISSUE #1875: UnionInPlace major performance regression
Using `UnionInPlace` instead of `Union` should/could be a strict win in terms of reducing memory allocation, but with the current implementation it is not. There is actually a `TODO` in the code which discusses this, but the magnitude of the issue is not immediately apparent. I've observed allocations on a benchmark I was doing going from 12GB to over 400GB when switching from Union to UnionInPlace.

The TODO is here: 
```
            // Use a bitmap container for the target bitmap, and union everything
            // into it.
            //
            // TODO(rartoul): Add another conditional case for n < ArrayMaxSize
            // (or some fraction of that value) to avoid allocating expensive
            // bitmaps when unioning many low-density array containers, but this
            // will require writing a union in place algorithm for an array container
            // that accepts multiple different containers to union into it for
            // efficiency.
            container := target.Containers.Get(iKey)
            // If target already has a bitmap container for iKey then we can reuse that,
            // otherwise we have to allocate a new one or convert the existing container
            // into a bitmap container.
            if container == nil {
                buf := make([]uint64, bitmapN)
```

So if one is Unioning a large sparse bitmap into a bitmap with little or no existing data, many new bitmap type containers will be created which is memory intensive. Indeed, I was able to reduce allocations in my benchmark to around 2GB after switching the order of my `UnionInPlace` operation.

This issue is unlikely to be as obvious when Unioning many bitmaps at once which is the original use case for which `UnionInPlace` was developed.

To fix this, we'll need to implement `unionInto` methods on containers that maintain the target container as an array or RLE container until it's obviously necessary to convert it.

seebs:
## Overview

UnionInPlace was making a lot of copies and/or new bitmaps to simplify the merge logic. The logic for just grabbing the one container for a key was slightly broken by late changes to the UnionInPlace PR to simplify/clean up the logic to treat the target as a method receiver, which made the SummaryStats computation of whether a container was the "only" container slightly wrong. Thus, in the case where the receiver had no container, but one of the things being merged in did, we were creating a new container and unioning everything into it -- even if there was only one evereything to union.

Cleaned that up, also stopped coercing everything to bitmap right away. We coerce to bitmap if we need to, but special-case array/array merges to allow them to stay arrays if N stays small.

Fixes #1875

## Pull request checklist

## Code review checklist

### DIFF
--- a/roaring/roaring.go
+++ b/roaring/roaring.go
@@ -1773,6 +1773,9 @@ func (c *Container) optimize() {
 	}
 }
 
+// unionInPlace does not necessarily preserve container's N; it's expected
+// to be used when running a sequence of unions, after which you should
+// call Repair(). (As of this writing, that only matters for bitmaps.)
 func (c *Container) unionInPlace(other *Container) {
 	switch c.containerType {
 	case containerBitmap:
--- a/roaring/roaring.go
+++ b/roaring/roaring.go
@@ -520,13 +520,12 @@ func (b *Bitmap) Intersect(other *Bitmap) *Bitmap {
 
 // Union returns the bitwise union of b and others as a new bitmap.
 func (b *Bitmap) Union(others ...*Bitmap) *Bitmap {
-	output := NewBitmap()
 	if len(others) == 1 {
+		output := NewBitmap()
 		b.unionIntoTargetSingle(output, others[0])
 		return output
 	}
-
-	output.UnionInPlace(b)
+	output := b.Clone()
 	output.UnionInPlace(others...)
 	return output
 }
@@ -561,9 +560,8 @@ func (b *Bitmap) unionIntoTargetSingle(target *Bitmap, other *Bitmap) {
 	}
 }
 
-// unionIntoTarget stores the union of b and others into target. b and others will
-// be left unchanged (unless one of them is also target), but target will be modified
-// in place.
+// unionInPlace stores the union of b and others into b. The others will
+// be left unchanged.
 //
 // This function performs an n-way union of n bitmaps. It performs this in an
 // optimized manner looping through all the bitmaps and performing unions one
@@ -573,42 +571,13 @@ func (b *Bitmap) unionIntoTargetSingle(target *Bitmap, other *Bitmap) {
 // participate in the union. This significantly reduces allocations. In addition,
 // because we perform the unions one container at a time across all the bitmaps, we
 // can calculate summary statistics that allow us to make more efficient decisions
-// up front. For example, imagine trying to perform a union across the following three
-// bitsets:
-//
-//     1. Bitmap A: Single array container at key 0 with 400 values in it.
-//     2. Bitmap B: Single array container at key 0 with 500 values in it.
-//     3. Bitmap C: Single array container at key 0 with 3500 values in it.
-//
-// Naive approach:
-//
-//     1. Perform union of bitmap A and B, container by container
-//         a. 400 + 500 < ArrayMaxSize so likely we will choose to allocate a new array
-//            container and then perform a unionArrayArray operation to merge the two
-//            arrays into the new array container.
-//     2. Perform a union of the bitmap generated in the step above with bitmap C.
-//        900 + 3500 > ArrayMaxSize so we will need to upgrade to a bitset container which
-//        we will have to allocate, and then we will have to perform two unions into the
-//        new bitmap container: one for the array container generated in the previous step,
-//        and one for the bitset container in bitmap C.
-//
-// Approach taken by this function:
-//
-//     1. Detect that bitmaps A, B, and C all have containers for key 0.
-//     2. Estimate the resulting cardinality of the union of all their containers to be
-//        400 + 500 + 3500 > ArrayMaxSize and decide upfront to use a bitset for the target
-//        container. Note that this is just an approximation of the final cardinality and can
-//        be off by a wide margin if there is a lot of overlap between containers, but that is
-//        fine, we'll still get the same result at the end, we'll just be more biased towards
-//        using bitmap containers will still being able to use array containers when all the
-//        cardinalities are small.
-//     3. Union the containers from bitmaps A, B, and C into the new bitset container directly
-//        using fast bitwise operations.
-//
-// In the naive approach, we had to allocate two containers, whereas in the optimized approach
-// we only had to allocate one container, and we also had to perform less union operations. This
-// example is simplistic, but the impact in terms of CPU cycles and memory allocations achieved
-// by using the optimized alogorithm when unioning many large bitmaps can be huge.
+// up front. For instance, if we have a non-bitmap target container, but we
+// expect more than ArrayMaxSize bits, we can convert to bitmap preemptively.
+// This will sometimes be wrong (we can't really tell how many bits we'll have
+// after a union) but is probably close enough to be useful. This will save
+// some reallocations for cases where several consecutive ops have array
+// representations, and we expect to have to convert to a bitmap eventually;
+// we don't allocate larger and larger array slices before doing that.
 //
 // An additional optimization that this function makes is that it recognizes that even when
 // CPU support is present, performing the popcount() operation isn't free. Imagine a scenario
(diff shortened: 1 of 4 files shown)


jaffee:
I put this into fragment.ImportRoaring like: 
```
@@ -1763,8 +1763,11 @@ func (f *fragment) importRoaring(data []byte, clear bool) error {
 
 	if clear {
 		bm = f.storage.Difference(bm)
-	} else if f.storage.Any() {
-		bm = f.storage.Union(bm)
+	} else if f.storage.Containers.Size() >= bm.Containers.Size() {
+		f.storage.UnionInPlace(bm)
+		bm = f.storage
+	} else {
+		bm.UnionInPlace(f.storage)
 	}
 
 	for rowID := range rowSet {
```

Benchcmp looks good:
```
benchmark                                                     old ns/op     new ns/op     delta
BenchmarkImportRoaring/Rows2Cache_ranked-8                    6180105       6131043       -0.79%
BenchmarkImportRoaring/Rows50Cache_ranked-8                   17139516      16956038      -1.07%
BenchmarkImportRoaring/Rows1000Cache_ranked-8                 27238809      28194480      +3.51%
BenchmarkImportRoaring/Rows100000Cache_ranked-8               104570846     104415830     -0.15%
BenchmarkImportRoaringUpdate/ranked2Rows20Cols-8              524057        518185        -1.12%
BenchmarkImportRoaringUpdate/ranked2Rows1000Cols-8            542770        567699        +4.59%
BenchmarkImportRoaringUpdate/ranked2Rows50000Cols-8           7025988       6822882       -2.89%
BenchmarkImportRoaringUpdate/ranked2Rows500000Cols-8          7788524       6856893       -11.96%
BenchmarkImportRoaringUpdate/ranked50Rows20Cols-8             647889        690404        +6.56%
BenchmarkImportRoaringUpdate/ranked50Rows1000Cols-8           3140017       3125205       -0.47%
BenchmarkImportRoaringUpdate/ranked50Rows50000Cols-8          32562283      24226950      -25.60%
BenchmarkImportRoaringUpdate/ranked50Rows500000Cols-8         45576973      33472502      -26.56%
BenchmarkImportRoaringUpdate/ranked1000Rows20Cols-8           3167222       3147993       -0.61%
BenchmarkImportRoaringUpdate/ranked1000Rows1000Cols-8         3740793       3765449       +0.66%
BenchmarkImportRoaringUpdate/ranked1000Rows50000Cols-8        45418052      35601999      -21.61%
BenchmarkImportRoaringUpdate/ranked1000Rows500000Cols-8       63076642      49400175      -21.68%
BenchmarkImportRoaringUpdate/ranked100000Rows20Cols-8         4234861       4128218       -2.52%
BenchmarkImportRoaringUpdate/ranked100000Rows1000Cols-8       4965404       4858865       -2.15%
BenchmarkImportRoaringUpdate/ranked100000Rows50000Cols-8      117548306     99915298      -15.00%
BenchmarkImportRoaringUpdate/ranked100000Rows500000Cols-8     160903920     158869631     -1.26%
BenchmarkImportRoaringIntoLargeFragment-8                     497445827     395656499     -20.46%

benchmark                                                     old allocs     new allocs     delta
BenchmarkImportRoaring/Rows2Cache_ranked-8                    130            130            +0.00%
BenchmarkImportRoaring/Rows50Cache_ranked-8                   1746           1746           +0.00%
BenchmarkImportRoaring/Rows1000Cache_ranked-8                 33294          33293          -0.00%
BenchmarkImportRoaring/Rows100000Cache_ranked-8               256294         256308         +0.01%
BenchmarkImportRoaringUpdate/ranked2Rows20Cols-8              58             58             +0.00%
BenchmarkImportRoaringUpdate/ranked2Rows1000Cols-8            105            105            +0.00%
BenchmarkImportRoaringUpdate/ranked2Rows50000Cols-8           208            166            -20.19%
BenchmarkImportRoaringUpdate/ranked2Rows500000Cols-8          208            166            -20.19%
BenchmarkImportRoaringUpdate/ranked50Rows20Cols-8             130            130            +0.00%
BenchmarkImportRoaringUpdate/ranked50Rows1000Cols-8           1815           1815           +0.00%
BenchmarkImportRoaringUpdate/ranked50Rows50000Cols-8          10881          2596           -76.14%
BenchmarkImportRoaringUpdate/ranked50Rows500000Cols-8         11848          2595           -78.10%
BenchmarkImportRoaringUpdate/ranked1000Rows20Cols-8           133            133            +0.00%
BenchmarkImportRoaringUpdate/ranked1000Rows1000Cols-8         3471           3471           +0.00%
BenchmarkImportRoaringUpdate/ranked1000Rows50000Cols-8        100474         38935          -61.25%
BenchmarkImportRoaringUpdate/ranked1000Rows500000Cols-8       125769         50092          -60.17%
BenchmarkImportRoaringUpdate/ranked100000Rows20Cols-8         138            138            +0.00%
BenchmarkImportRoaringUpdate/ranked100000Rows1000Cols-8       3905           3905           +0.00%
BenchmarkImportRoaringUpdate/ranked100000Rows50000Cols-8      402957         162715         -59.62%
BenchmarkImportRoaringUpdate/ranked100000Rows500000Cols-8     574944         351759         -38.82%
BenchmarkImportRoaringIntoLargeFragment-8                     1567068        526777         -66.38%

benchmark                                                     old bytes     new bytes     delta
BenchmarkImportRoaring/Rows2Cache_ranked-8                    19331         19332         +0.01%
BenchmarkImportRoaring/Rows50Cache_ranked-8                   240952        240964        +0.00%
BenchmarkImportRoaring/Rows1000Cache_ranked-8                 5166511       5165697       -0.02%
BenchmarkImportRoaring/Rows100000Cache_ranked-8               40269989      40270966      +0.00%
BenchmarkImportRoaringUpdate/ranked2Rows20Cols-8              76712         76712         +0.00%
BenchmarkImportRoaringUpdate/ranked2Rows1000Cols-8            286856        286856        +0.00%
BenchmarkImportRoaringUpdate/ranked2Rows50000Cols-8           285213        282013        -1.12%
BenchmarkImportRoaringUpdate/ranked2Rows500000Cols-8          285287        281820        -1.22%
BenchmarkImportRoaringUpdate/ranked50Rows20Cols-8             141798        141793        -0.00%
BenchmarkImportRoaringUpdate/ranked50Rows1000Cols-8           3981197       3981218       +0.00%
BenchmarkImportRoaringUpdate/ranked50Rows50000Cols-8          7279204       2601867       -64.26%
BenchmarkImportRoaringUpdate/ranked50Rows500000Cols-8         11476959      3536559       -69.19%
BenchmarkImportRoaringUpdate/ranked1000Rows20Cols-8           50345         50341         -0.01%
BenchmarkImportRoaringUpdate/ranked1000Rows1000Cols-8         1702708       1702731       +0.00%
BenchmarkImportRoaringUpdate/ranked1000Rows50000Cols-8        12567697      6307264       -49.81%
BenchmarkImportRoaringUpdate/ranked1000Rows500000Cols-8       16307988      8458899       -48.13%
BenchmarkImportRoaringUpdate/ranked100000Rows20Cols-8         546803        546789        -0.00%
BenchmarkImportRoaringUpdate/ranked100000Rows1000Cols-8       2064089       2064080       -0.00%
BenchmarkImportRoaringUpdate/ranked100000Rows50000Cols-8      44150201      23329376      -47.16%
BenchmarkImportRoaringUpdate/ranked100000Rows500000Cols-8     64531238      41585229      -35.56%
BenchmarkImportRoaringIntoLargeFragment-8                     202440744     88946368      -56.06%
```

seebs:
@richardartoul, can you have a look at this change? I am not 100% confident that I understand this code well enough, and I might have missed something relevant. But basically, there were performance regressions in the case where we were merging very sparse things (either two array containers, or an array container merged with no container), and i think this fixes them, but I'd like second opinions on it.

richardartoul:
@seebs @jaffee Whoa this looks promising. Yeah I didn't have time to do this at the time, but looks like its a huge win!

I'll take a look through today or tomorrow if that works for you.

Also hopefully I didn't cause you any issues with the regression? I think the way I wrote the code is it still uses the old logic and you had to opt-in to the new path.

jaffee:
@richardartoul no worries—you did indeed structure it so that it was "opt-in". I only noticed it when I tried to switch over to UnionInPlace and had significant regressions in some benchmarks.

Appreciate your help!

On a mostly unrelated note, I heard that sync.Pool was getting some updates in Go 1.13 (e.g. not being cleared every GC). Might make it worth it for us to spend some time looking into that a bit more for container/array/bitmap/rle reuse.

seebs:
No issues to speak of.

My current theory is there's actually two parts to the performance regression. One is that the "only one container has this key" logic and the "target container has this key" logic were sort of muddled, and the net result was that I don't think we ever got the "only one container has this key" result -- if the key was present in one of the non-target containers, it was always going to report false. What was missing there was a way to distinguish "this key was found in exactly one other container, and absent in the base container".

Except now that I think about it, I think my design's clumsier than it should be. Better logic would be:

* If the key is absent from the target container, clone the first container we found, then compute summary stats on the iterators *after* that.
* If we find any, we have to do a union of them.
* If we don't, we're done.
* If the key is present in the target container, compute summary stats on all the other containers, including the one we start with, and do a union.

So I'll probably rework that.

My understanding is that sync.Pool is currently basically-useless, but may *become* useful again in 1.13. I think the best path would be to work a bit on our proposed work for doing magic to let us get away with not needing to keep three distinct slices in a container first, because a huge amount of the pooling would improve if it were easier to repurpose containers without necessarily reallocating them.

jaffee:
My understanding is that sync.Pool is currently basically-useless, but may become useful again in 
    1.13. I think the best path would be to work a bit on our proposed work for doing magic to let us 
    get away with not needing to keep three distinct slices in a container first, because a huge amount 
    of the pooling would improve if it were easier to repurpose containers without necessarily 
    reallocating them.

^ 100% agree on sync.Pool and starting with the pointer magic to get down to one or two slices.

roaring/roaring.go
```diff
@@ -705,93 +674,98 @@ func (b *Bitmap) unionInPlace(others ...*Bitmap) {
...
+			expectedN := int64(0)
 
-			// There was more than one container across the bitmaps with key iKey
-			// so we need to calculate a union.
+			// determine whether we have a target to union into.
+			tContainer := target.Containers.Get(iKey)
+			// if the target's full, short-circuit out.
+			if tContainer != nil {
+				if tContainer.n == maxContainerVal+1 {
+					bitmapIters.markItersWithCurrentKeyAsHandled(i, iKey)
```
richardartoul: So kind of a jerk move to ask this since I originally named this function, but looking at it now seems like the name should be `markItersWithKeyAsHandled` since you have to pass in the key, its not relying on any iterator state right?

roaring/roaring.go
```diff
@@ -705,93 +674,98 @@ func (b *Bitmap) unionInPlace(others ...*Bitmap) {
...
-			// otherwise we have to allocate a new one or convert the existing container
-			// into a bitmap container.
-			if container == nil {
-				buf := make([]uint64, bitmapN)
-				ob := buf[:bitmapN]
-				container = &Container{
-					bitmap:        ob,
-					n:             0,
-					containerType: containerBitmap,
+			expectedN += summaryStats.n
```
richardartoul: Are you double counting the target container's n? Seems like you include the target containers n when you do

```golang
expectedN = int64(tConainter.n)
```

and then when you do

```golang
summaryStats := bitmapIters[i:].calculateSummaryStats(iKey)
```

you're gonna include it again

richardartoul:
reviewing now

roaring/roaring.go
```diff
@@ -705,93 +674,98 @@ func (b *Bitmap) unionInPlace(others ...*Bitmap) {
...
+			var itersToUnion handledIters
+			if tContainer == nil {
+				if summaryStats.c == 1 {
+					// There's no target and we have only one container, we
+					// can just clone it instead of unioning
+					statsHit("unionInPlace/reuse")
+					target.Containers.Put(iKey, iContainer.Clone())
```
richardartoul: You know there is only one container, but do you know for sure its the current container? The summaryStats function will look ahead arbitrarily far I think

roaring/roaring.go
```diff
@@ -705,93 +674,98 @@ func (b *Bitmap) unionInPlace(others ...*Bitmap) {
...
+					bitmapIters[i].handled = true
+					continue
+				}
+				// If we expect enough items that we'll want to convert to bitmap,
+				// and the container we already found isn't a bitmap, make a bitmap
+				// and union everything into that. Otherwise clone the first
+				// container and union everything else into it.
+				if expectedN >= 512 && iContainer.containerType != containerBitmap {
```
richardartoul: I thought max array size was like 4096? Maybe I'm misremembering

roaring/roaring.go
```diff
@@ -705,93 +674,98 @@ func (b *Bitmap) unionInPlace(others ...*Bitmap) {
...
+					statsHit("unionInPlace/newBitmap")
+					tContainer = &Container{containerType: containerBitmap, bitmap: make([]uint64, bitmapN)}
```
richardartoul: could you just do:

```golang
tContainer = iContainer.toBitmap()
```
and then both branches can share the

```golang
itersToUnion = bitmapIters[i+1:]
```

line? Or is that not equivalent

roaring/roaring.go
```diff
@@ -705,93 +674,98 @@ func (b *Bitmap) unionInPlace(others ...*Bitmap) {
...
-				ob := buf[:bitmapN]
-				container = &Container{
-					bitmap:        ob,
-					n:             0,
-					containerType: containerBitmap,
+			expectedN += summaryStats.n
+			var itersToUnion handledIters
+			if tContainer == nil {
+				if summaryStats.c == 1 {
+					// There's no target and we have only one container, we
```
richardartoul: There is a lot of if/else nesting in this section which is fine, but it would be a little easier to follow if you began each lexical scope with a little comment like this (specifically you don't have them for a lot of the else statements. I'll leave some examples below.

roaring/roaring.go
```diff
@@ -705,93 +674,98 @@ func (b *Bitmap) unionInPlace(others ...*Bitmap) {
...
+					// can just clone it instead of unioning
+					statsHit("unionInPlace/reuse")
+					target.Containers.Put(iKey, iContainer.Clone())
+					bitmapIters[i].handled = true
+					continue
+				}
+				// If we expect enough items that we'll want to convert to bitmap,
```
richardartoul: could move this comment into the conditional

roaring/roaring.go
```diff
@@ -705,93 +674,98 @@ func (b *Bitmap) unionInPlace(others ...*Bitmap) {
...
+				// and the container we already found isn't a bitmap, make a bitmap
+				// and union everything into that. Otherwise clone the first
+				// container and union everything else into it.
+				if expectedN >= 512 && iContainer.containerType != containerBitmap {
+					statsHit("unionInPlace/newBitmap")
+					tContainer = &Container{containerType: containerBitmap, bitmap: make([]uint64, bitmapN)}
+					itersToUnion = bitmapIters[i:]
+				} else {
+					statsHit("unionInPlace/clone")
```
richardartoul: example comment: "target container does not exist and we don't expect the cardinality to be high enough to justify a bitmap, so just clone the current container and move on"

roaring/roaring.go
```diff
@@ -705,93 +674,98 @@ func (b *Bitmap) unionInPlace(others ...*Bitmap) {
...
+					tContainer = iContainer.Clone()
+					itersToUnion = bitmapIters[i+1:]
+				}
+			} else {
+				itersToUnion = bitmapIters[i:]
```
richardartoul: "target container already exists"

roaring/roaring.go
```diff
@@ -705,93 +674,98 @@ func (b *Bitmap) unionInPlace(others ...*Bitmap) {
...
+				if expectedN >= 512 && tContainer.containerType != containerBitmap {
+					statsHit("unionInPlace/convertToBitmap")
```
richardartoul: "target container already exists and the expected cardinality is high enough to justify a bitmap so we convert the existing container to a bitmap"

richardartoul: (comment to go below this if condition where you decide not to do anything:

"target container already exists but the expected cardinality is not high enough to justify force converting the container to a bitmap so we do nothing"

roaring/roaring.go
```diff
@@ -705,93 +674,98 @@ func (b *Bitmap) unionInPlace(others ...*Bitmap) {
...
 
-			// Once we've acquired a bitmap container (either by reusing the existing one
-			// or allocating a new one) then the last step is to iterate through all the
-			// other containers to see which ones have the same key, and union all of them
-			// into the target bitmap container. Only need to loop starting from i because
-			// anything previous to that has already been handled.
-			for j := i; j < len(bitmapIters); j++ {
-				jIter := bitmapIters[j]
-				jKey, jContainer := jIter.iter.Value()
+			// Now we union everything together. In many cases, this will mean
```
richardartoul: "Now we union all the containers with the same key as the current container together"

roaring/roaring.go
```diff
@@ -705,93 +674,98 @@ func (b *Bitmap) unionInPlace(others ...*Bitmap) {
...
+			// in, but we can specialize away from that in some cases.
+			//
+			// We iterate through the other Containers [plural] checking to see whether
+			// any of them have a Container [singular] with the same key to merge
+			// in. We only need to check i and higher, because anything lower would
+			// already be handled.
+			//
+			// If the target didn't previously have this index, we cloned the
+			// first container we found for it, so we will not then union that one
+			// in.
```
richardartoul: "so we will not then union that one in since we've already set `itersToUnion` to skip it

roaring/roaring.go
```diff
@@ -705,93 +674,98 @@ func (b *Bitmap) unionInPlace(others ...*Bitmap) {
...
-					if jContainer.isArray() {
-						unionBitmapArrayInPlace(container, jContainer)
-					} else if jContainer.isRun() {
-						unionBitmapRunInPlace(container, jContainer)
-					} else {
-						unionBitmapBitmapInPlace(container, jContainer)
-					}
-					bitmapIters[j].handled = true
+					tContainer.unionInPlace(jContainer)
+					itersToUnion[j].handled = true
```
richardartoul: Maybe a little over paranoid, but a comment here could be useful: "iter is not a pointer so we need to explicitly update the struct by indexing into the slice."

roaring/roaring.go
richardartoul: Oh I guess you know its the current container because thats why you were looking at this key in the first place yeah?

roaring/roaring.go
seebs: If I have done it right, the target container's N is only included when there was an existing target container, in which case the target container isn't one of the ones in bitmapIters (or a clone of them).

roaring/roaring.go
seebs: Yup!

roaring/roaring.go
seebs: 2048. But! It turns out that, if you do an array/array in-place merge of >500 items, it's *faster* to do this -- doing the conversion only if the expected N is that high results in a factor of two or more performance regression for use cases where we get about 1k items in a container.

roaring/roaring.go
seebs: We don't currently have a toBitmap(), I don't think? So we have a "convert this container in-place to a bitmap", and we have a Clone(). It might make more sense to add a CloneAsBitmap(), though.

roaring/roaring.go
jaffee: For the record, array max size is 4096, run max size is 2048.

seebs:
Yes, the UnionInPlaceProp test is still there.

Basically, the problem I was originally fixing was that, if you have really sparse values -- say, two containers with 1 value each -- the conversion to bitmap, which then results in a later conversion back to array, is really expensive. But as the density increases, it starts being worth it quite a bit sooner than ArrayMaxSize elements. I may actually look at this for the general UnionArrayArray case; it might be that it's faster there too for some N.

roaring/roaring.go
seebs: ... That is true, and also obvious, and now I'm not sure how I managed to get confused about it. Anyway, cutover is way lower than the max size for an array, just because the array merge is not very cheap. I'm not sure whether there's an easy way to make it cheaper.

jaffee: We could probably tweak ArrayMaxSize down to improve query performance at the cost of increased storage. Whether or not this would be worthwhile probably depends on the particular use case and workload. I'm inclined to believe we have easier and more general wins elsewhere though before we start tuning at that level.

seebs: Yeah. Might be fun to benchmark, but there's a lot of additional tradeoffs, etc., so.

roaring/roaring.go
```diff
@@ -1792,6 +1773,44 @@ func (c *Container) optimize() {
 	}
 }
 
+func (c *Container) unionInPlace(other *Container) {
```
jaffee: let's add a comment that `c.n` may not necessarily be correct after this operation and should be repaired.

PR #1901, opened 2019-03-15
make Containers smaller, especially when they have small contents

seebs:
Revised summary: This branch hides access to the array/runs/bitmap slices of a Container behind function calls, and actually implements them from a single shared pointer/len/cap tuple. Additional tests are added to make us less likely to grab the wrong slice, but not enabled by default. Additionally, in the case where the actual storage needed is small (array or run containers using 10 or fewer bytes of storage), we use part of the Container object as backing store. (This is effectively free; we need 22 bytes of data for the structure, and the nearest alloc size class is 32.) This spares us extra allocations in small cases.

A variant scheme which squishes things down to 16 bytes was evaluated, but it imposed additional runtime costs for bit ops and felt more brittle. This still gets us from the smallest container (an array of one item) taking up 88 total allocated bytes to 32 total allocated bytes, which should relieve memory pressure a fair bit.

Performance of single-bit writes is somewhat degraded (potentially as much as 20-25% in a tight loop), but performance of just about everything else improves significantly, including general system-wide improvements from dramatically reduced GC workload. (The garbage collector now scans only the first word of each container, instead of needing to scan the first 8 and evaluate three of them as pointers.)


seebs:
oh, and the copy of btree containers in enterprise/ is now out of sync, i'll have to fix that.

roaring/roaring.go
```diff
@@ -78,7 +83,7 @@ type Containers interface {
 	// PutContainerValues updates an existing container at key.
 	// If a container does not exist for key, a new one is allocated.
 	// TODO(2.0) make n  int32
-	PutContainerValues(key uint64, containerType byte, n int, mapped bool)
+	PutContainerValues(key uint64, typ ContainerType, n int, mapped bool)
```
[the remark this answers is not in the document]
seebs: We could in principle, but I don't think it actually breaks anything that isn't fixed already within this. Unless someone's actually implemented another Containers.

roaring/roaring.go
```diff
@@ -1421,10 +1419,92 @@ func (iv interval16) runlen() int32 {
 	return 1 + int32(iv.last-iv.start)
 }
 
-// newContainer returns a new instance of container.
-func NewContainer() *Container {
+// NewContainer returns a new instance of container. This trivial function
+// may later become more interesting.
+func NewContainer(typ ContainerType, n int32) *Container {
```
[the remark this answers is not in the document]
seebs: That makes sense.

roaring/roaring.go
```diff
@@ -1421,10 +1419,92 @@ func (iv interval16) runlen() int32 {
...
+// slice of intervals.
+func NewContainerRun(set []interval16) *Container {
+	c := &Container{typ: containerRun, oneSlice: *(*[]uint16)(unsafe.Pointer(&set))}
+	for _, run := range set {
+		c.n += int32(run.last-run.start) + 1
+	}
+	return c
+}
+
+// array yields the data viewed as a slice of intervals.
```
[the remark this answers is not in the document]
seebs: ... yes, yes it is.

roaring/roaring.go
```diff
@@ -1939,102 +2046,106 @@ func (c *Container) bitmapMax() uint16 {
...
+	return runs[len(runs)-1].last
 }
 
 // bitmapToArray converts from bitmap format to array format.
 func (c *Container) bitmapToArray() {
 	statsHit("bitmapToArray")
-	c.array = make([]uint16, 0, c.n)
-	c.containerType = containerArray
+	bitmap := c.bitmap()
+	c.setBitmap(nil)
```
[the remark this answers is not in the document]
seebs: It's optional now, and actually I think I've removed it in more recent versions.

When all three exist separately, it should ideally be:

1. extract bitmap.
2. setBitmap(nil)
3. setTyp(containnerArray)
4. setArray(arrayValues)

And any other order is an error.

roaring/roaring.go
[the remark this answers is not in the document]
jaffee: yeah I don't think it breaks any of our stuff, but technically this is an API change and by the rules of semver means we shpuld go to 2.0

roaring/container_stash.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+	// no array: start with our default 2-value array
+	if runs == nil {
+		c.pointer, c.len, c.cap = (*uint16)(unsafe.Pointer(&c.data[0])), 0, stashedRunSize
+		return
+	}
+	h := (*reflect.SliceHeader)(unsafe.Pointer(&runs))
+	if h.Data == uintptr(unsafe.Pointer(c.pointer)) {
+		// nothing to do but update length
+		c.len = int32(h.Len)
```
jaffee: I guess it's impossible that `cap` would change without the actual pointer changing?

seebs: it's impossible that it would *correctly* change without the actual pointer changing. i believe.

... i will ask #darkarts on gopher slack, though.

roaring/container_stash.go
```diff
@@ -0,0 +1,263 @@
...
+// setBitmap stores a set of uint64s as data.
+func (c *Container) setBitmap(bitmap []uint64) {
+	if roaringParanoia {
+		if c.typ != containerBitmap {
+			panic("attempt to write non-bitmap's bitmap")
+		}
+	}
+	h := (*reflect.SliceHeader)(unsafe.Pointer(&bitmap))
+	c.pointer, c.len, c.cap = (*uint16)(unsafe.Pointer(h.Data)), int32(h.Len), int32(h.Cap)
+	runtime.KeepAlive(&bitmap)
```
jaffee: I don't think I understand the use of KeepAlive—is the idea that as soon as you `unsafe.Pointer(&bitmap)` the bitmap could be GCd?

roaring/containers_btree.go
```diff
@@ -120,7 +122,7 @@ func (btc *bTreeContainers) Count() (n uint64) {
 	e, _ := btc.tree.Seek(0)
 	_, c, err := e.Next()
 	for err != io.EOF {
-		n += uint64(c.N())
+		n += uint64(c.n)
```
jaffee: wouldn't N() be inlined anyway?

roaring/roaring.go
```diff
@@ -695,11 +699,7 @@ func (b *Bitmap) unionInPlace(others ...*Bitmap) {
 				// range that a container can store, so instead of calculating a
 				// union we can generate an RLE container that represents the entire
 				// range.
-				tContainer = &Container{
-					runs:          []interval16{{start: 0, last: maxContainerVal}},
-					containerType: containerRun,
-					n:             maxContainerVal + 1,
-				}
+				tContainer = NewContainerRun([]interval16{{start: 0, last: maxContainerVal}})
```
jaffee: I wonder if it's worth having a special function for building a full run container (to avoid the slice allocation and conversion). Probably not necessary right now, but something to keep an eye out for.

roaring/roaring.go
```diff
@@ -953,7 +959,8 @@ func (b *Bitmap) WriteTo(w io.Writer) (n int64, err error) {
 	byte8 := make([]byte, 8)
 
 	// Build header before writing individual container blocks.
-	// Metadata for each container is 8+2+2+4 = sizeof(key) + sizeof(containerType)+sizeof(cardinality) + sizeof(file offset)
+	// Metadata for each container is 8+2+2+4 = sizeof(key) + sizeof(type)+sizeof(cardinality) + sizeof(file offset)
+	// Type is stored as 2 bytes, even though it's only got values 1..3.
```
jaffee: yes, I noticed quite a few opportunities for improvement in our roaring format.... TODO(2.0) maybe

roaring/container_stash.go
seebs: Basically, if there's no references to `bitmap` itself after line 147, the GC is in principle allowed to discard the bitmap slice header, because lifetime analysis lets it realize that the local variable isn't *actually* still in use. The reflect.SliceHeader doesn't count as having a live pointer, so it's possible that, if this was the only reference to the slice's backing store (the caller's copy could also be getting nixed by lifetyme analysis), the backing store gets dropped. It's not very likely, I think, but also runtime.KeepAlive is basically free; it effectively serves as a compiler hint.

roaring/containers_btree.go
seebs: Yes, probably. Context: I had a previous patch where *every* `.n` had to be a `.N()`, and I ended up reverting these fairly aggressively. So this change isn't really necessary, but it's also harmless.

roaring/roaring.go
seebs: It might well be. I think NewContainerRun was just a bit too large to inline, so the compiler probably can't do it. But it shouldn't come up wuch.

roaring/roaring.go
seebs: Yeah. I'm still pondering a hypothetical mixed 16-bit/32-bit container format, I think on the whole it's probably not worth it, but it might be worth thinking about.

roaring/container_stash.go
jaffee: I'm out of my comfort zone, but is it possible for something like `realloc` to make an allocation larger without moving it? I'm reading the docs but it's still unclear to me. Also don't know if there's something higher level (within Go) that would prevent this.

seebs: So far as I know, allocations have a fixed size, because they're in size spans based on their size, and the cap should always have reflected the actual size of the underlying allocation in the first place. I think.

roaring/roaring.go
```diff
@@ -2049,97 +2011,112 @@ func (c *Container) bitmapToRun() {
...
-			c.array = append(c.array, uint16(v))
+			array[n] = uint16(v)
+			n++
 		}
 	}
-	c.runs = nil
-	c.mapped = false
+	if roaringParanoia {
+		if n != c.n {
+			panic("run has fewer bits set than container.n")
```
jaffee: couldn't it have more as well?

roaring/roaring.go
```diff
@@ -2049,97 +2011,112 @@ func (c *Container) bitmapToRun() {
...
+		copy(runs, cRuns)
+		out = NewContainerRun(runs)
 	}
-	return other
+	if roaringParanoia {
+		if out.n != out.count() {
+			panic("cloned container has wrong n")
+		}
+	}
+	// this should probably never happen
```
jaffee: is this comment referring to the previous block?

roaring/roaring.go
```diff
@@ -2166,40 +2144,42 @@ func (c *Container) arrayWriteTo(w io.Writer) (n int64, err error) {
 	//}
 
 	// Write sizeof(uint16) * cardinality bytes.
-	nn, err := w.Write((*[0xFFFFFFF]byte)(unsafe.Pointer(&c.array[0]))[:2*c.n])
+	nn, err := w.Write((*[0xFFFFFFF]byte)(unsafe.Pointer(&array[0]))[: 2*c.n : 2*c.n])
```
jaffee: can you explain the significance of adding the capacity to the slicing op?

My read is that we previously weren't doing this, and so technically the slice's capacity value was 0xFFFFFFF or whatever. This didn't matter because all the containers were mapped, and so we weren't appending to the slices and thereby using capacity we didn't actually have.

roaring/roaring.go
seebs: no, because that would have been caught by the previous check a few lines up (which is where it is to avoid array/slice overflows).

roaring/roaring.go
seebs: Yes. it should be above it. I think it may originally have referred to a "return nil" because the individual cases returned, but I changed them to create a result to return because I wanted to share the error checking code.

roaring/roaring.go
seebs: Yup. So, it's harmless right now, but when I was doing the packed representation, storing that unreasonably large value in capacities resulted in overwriting other bits of the representation. But I'd also like to be a little insured against the compiler going insane and deciding that it can do some weird optimization because it "knows" that the capacity of the slice is large.

jaffee:
Pilosa's memory usage was 25% lower on a random 4GB index I had on my hard drive. I'm about ready to approve this if you can fix that one comment, and rebase.

seebs:
okay, have rebased (cleanly, apparently). so far as i know, no doc changes, since this is 100% internal. i am inclined to leave the last couple of small bugfix commits as history.

my impression is that overall, the performance tends to be slightly better just because of the reduction in garbage collector workload, and obviously we can get a lot further before running out of memory. (How much varies with container types; with all bitmap containers, it should only be about 1% less memory, but it'll still be less work for the garbage collector.)

PR #1902, opened 2019-03-17
Unbounded BSI w/ sign magnitude

### LINKED ISSUE #893: Unbounded BSI fields
### Description
When ingesting new data, it is often unknown ahead of time what range of values an integer might take. It would be convenient if one could create a BSI field without specifying a range up front. As values came in, the field would add rows to itself as necessary.

This flexibility would allow people to write less code/configuration for ingesting data.

Because ranges can start and end anywhere in positive or negative integers, we'd have to make some initial assumptions about the field, and have the ability to re-arrange things if necessary.

Not needing to define field ranges ahead of time when defining BSI fields.

benbjohnson:
## Overview

This pull request implements BSI with variable bit depth using a sign magnitude to indicate whether a value is positive or negative. This also rearranges the existence bit to be the first bit instead of the last bit.

The BSI executor tests are working and there are some minor fragment tests I'm still resolving but otherwise this is working. This does change the on-disk format so do we want to have Pilosa do a migration on upgrade? Essentially it would just need to rewrite BSI data by 1) moving the existence bit to the `0` position and, 2) shifting all the other bits up by `2` positions.

Fixes #893

## Pull request checklist

## Code review checklist


fragment_internal_test.go
```diff
@@ -275,7 +275,7 @@ func TestFragment_SetValue(t *testing.T) {
 		}
 
 		// Non-existent value.
-		if value, exists, err := f.value(100, 11); err != nil {
+		if value, exists, err := f.value(101, 11); err != nil {
```
seebs: this doesn't seem right -- if we set 100 up above, we should still have 100 here, no?

fragment.go
```diff
@@ -741,11 +790,26 @@ func (f *fragment) importSetValue(columnID uint64, bitDepth uint, value uint64,
...
+		}
 	}
-	if clear {
+
+	// Mark sign bit.
+	if p, err := f.pos(uint64(bsiSignBit), columnID); err != nil {
+		return changed, errors.Wrap(err, "getting not-null pos")
+	} else if value >= 0 || clear {
 		if c, err := f.storage.Remove(p); err != nil {
 			return changed, errors.Wrap(err, "removing not-null from storage")
```
seebs: it looks like this message should be about the sign bit now, not the not-null bit?

fragment.go
```diff
@@ -836,25 +914,35 @@ func (f *fragment) max(filter *Row, bitDepth uint) (max, count uint64, err error
...
-		ii := i - 1 // allow for uint range: (bitDepth-1) to 0
-		row := f.row(uint64(ii))
+	// Find lowest negative number w/o sign and negate, if no positives are available.
+	prow := consider.Difference(f.row(bsiSignBit))
+	if prow.Count() == 0 {
+		max, count = f.minUnsigned(consider, bitDepth)
+		return -max, count, nil
+	}
+
+	// Otherwise find highested positive number.
```
seebs: "highested" seems typoish

fragment_internal_test.go
jaffee: previously, we were passing bitDepth of 11 to a field whose bitDepth was set to 10. Since the `exists`  bitmap was at the end, this made it look like it didn't exist. Now that exists is always at the beginning, we're trying to get a value from a column (101) which never had a bit set.

seebs: ohh! i get it. okay, that makes more sense now. thanks.

fragment.go
jaffee: +1

fragment.go
```diff
@@ -803,29 +873,37 @@ func (f *fragment) min(filter *Row, bitDepth uint) (min, count uint64, err error
 		return 0, 0, nil
 	}
 
-	for i := bitDepth; i > uint(0); i-- {
-		ii := i - 1 // allow for uint range: (bitDepth-1) to 0
-		row := f.row(uint64(ii))
+	// Find highest negative number w/o sign and negate, if negatives are available.
```
jaffee: I found this comment confusing—aren't we finding the highest number *with* the sign bit set?

jaffee:
So... yes, I think it would be best if we had an automatic migration—can we do it in such a way that it will tolerate weird failures?

Like, I know we have .startup.log in the top level of the Pilosa data dir, so we can detect when Pilosa starts up that an upgrade has happened. But what if the node gets powered down while we're migrating fragments. Can we tell on a fragment-by-fragment basis which ones have been migrated and which haven't? I haven't looked at this deeply, but my instinct is that we probably can based on the available metadata (or, I guess we could add some new metadata, the absence of which would let us know—I think there is a spare bit or two in the roaring headers that we could appropriate).

So yeah... this doesn't seem too unsafe provided we take the usual precautions (e.g. writing a separate file and then `mv`ing it into place). Ben has a heck of a lot more experience with this kind of thing that I do, and I think I've just convinced myself that it's doable, so yes 👍 let's auto-upgrade to this new format.

fragment.go
seebs: It's the highest negative number, which is to say, the one closest to being positive.

seebs: er, wait.

it's the negative number which is highest, if you ignore its sign. okay, conceded, the comment is probably confusing.

benbjohnson:
@jaffee @seebs I attempted a couple different ways with upgrading the roaring file. I originally tried a sidecar file to store the version but that was pretty ugly.

I ended up using part of the `roaring.Bitmap` cookie which is a magic number in the first 2 bytes and the storage version (which is currently zero) in the next two bytes. I shortened the storage version to 1 byte and used the 4th byte to store user-defined flags (the user here being the `pilosa` package).

The `fragment.Open()` now checks if the flag is unset and will automatically upgrade if the flag is not set. The upgrade is pretty straightforward. It iterates over all the bits and anything in the `bitDepth` row gets moved to the `0` row and everything gets moved up `2` rows. It writes to a temporary file next to the existing file, flush & close, and then renames over top of the original atomically.

I'm still doing some rearranging on the upgrade code so it isn't totally compiling yet but it's close. I wanted to push up and get some feedback. Do you see any issues with this approach?

jaffee:
I haven't had a chance to review the code, but your description is more or
less exactly what I was thinking.

jaffee:
After some light review, I agree with the direction here.

I hadn't thought about how a view would now need to specify flags at fragment creation time, so I don't love that this leaks up to that level, but It makes sense, and I don't have a better suggestion.

seebs:
Hmm. So, thinking this through: If we plan to bump the storage version, we *could* just use the storage version itself as the indicator. If storage version is 0, we have old-BSI and need to upgrade and reopen. If storage version is 1, it's new and we don't have to do anything.

But for some reason that's bugging me, and I like the explicit separate flag better.

I didn't realize there was a space for a storage version separate from the magic number.

benbjohnson:
@seebs I originally used the `storageVersion` but the roaring format doesn't change so that seemed odd. Also, I found it confusing that not all fragments would need to be upgraded to the new version so that's why I made a "user-defined flags" byte that can be delegated to Pilosa for whatever.

jaffee:
I agree that the storage version serves a slightly different purpose and this is something higher level. I hope that we don't go through 256 storage versions...

benbjohnson:
@jaffee @seebs Debugging the autoupgrade was more problematic than I had originally hoped but I finally got it working and did a bunch of testing on it. BSI in general is kinda tricky to test.

I updated the docs too. This PR is ready to go but it could definitely use some extra eyes on it.

field.go
```diff
@@ -972,14 +988,26 @@ func (f *Field) Value(columnID uint64) (value int64, exists bool, err error) {
...
-		return false, ErrBSIGroupValueTooLow
-	} else if value > bsig.Max {
-		return false, ErrBSIGroupValueTooHigh
+	}
+
+	// Increase max value if the unsigned value is greater.
+	uvalue := uint64(value)
+	if value < 0 {
+		uvalue = uint64(-value)
+	} else if uvalue > uint64(bsig.Max) {
```
jaffee: if `value` is negative, we skip the check here? The comment above seems to indicate this is not correct.

field.go
```diff
@@ -1172,10 +1200,23 @@ func (f *Field) importValue(columnIDs []uint64, values []int64, options *ImportO
...
+		if value < 0 {
+			uvalue = uint64(-value)
+		}
+
+		if int64(uvalue) > bsig.Max {
+			if err := func() error {
+				f.mu.Lock()
+				defer f.mu.Unlock()
+				bsig.Max = int64(uvalue)
+				f.options.Max = int64(uvalue)
```
jaffee: since we're sort of caching bsig.Max locally, it seems like it might be possible for f.options.Max to be set higher by another import going on concurrently, and then accidentally reset lower in here.

field.go
```diff
@@ -1203,9 +1243,9 @@ func (f *Field) importValue(columnIDs []uint64, values []int64, options *ImportO
 			return errors.Wrap(err, "creating fragment")
 		}
 
-		baseValues := make([]uint64, len(data.Values))
+		baseValues := make([]int64, len(data.Values))
 		for i, value := range data.Values {
-			baseValues[i] = uint64(value - bsig.Min)
+			baseValues[i] = int64(value - bsig.Min)
```
jaffee: we're kind of repurposing `bsiGroup.Min` now, right? Like previously it actually meant the minimum value stored in the field, but not it means the "base" value which the which the positive or negative stored values are  added to?

If that's correct, it might be worth renaming (not sure if it's exposed anywhere) or at least adding a comment on the struct field and in key places of use.

field.go
jaffee: I think there is a race here if multiple imports are going on at once. Because f.options.Max might be changed concurrently, and we don't have a read lock on it since we're effectively caching the value with bsig.Max. Another goroutine might set it higher, and then we come along and set it higher than our previous known value, but lower than what the other routine just changed it to.

docs/data-model.md
```diff
@@ -145,7 +145,7 @@ curl localhost:10101/index/repository/field/quantity \
 
 ##### BSI Range-Encoding
```
seebs: Should we perhaps include a reference to the previous format, and a comment about the upgrade process? Seems like it'd be very nice for that to be present in case Something Goes Wrong.

field.go
seebs: yeah, i don't think we want that `else`, in this case?

field.go
```diff
@@ -972,14 +988,26 @@ func (f *Field) Value(columnID uint64) (value int64, exists bool, err error) {
...
+	uvalue := uint64(value)
+	if value < 0 {
+		uvalue = uint64(-value)
+	} else if uvalue > uint64(bsig.Max) {
+		if err := func() error {
+			f.mu.Lock()
+			defer f.mu.Unlock()
+			bsig.Max = int64(uvalue)
+			f.options.Max = int64(uvalue)
+			return f.saveMeta()
```
seebs: won't this terminate before actually saving the value, in this case?

field.go
```diff
@@ -989,7 +1017,7 @@ func (f *Field) SetValue(columnID uint64, value int64) (changed bool, err error)
 	}
 
 	// Determine base value to store.
-	baseValue := uint64(value - bsig.Min)
+	baseValue := int64(value - bsig.Min)
```
seebs: On thinking about this: I think this should possibly be happening up above. The maximum isn't the value furthest from zero, it's the value furthest from bsig.Min. I think.

fragment.go
```diff
@@ -827,41 +880,64 @@ func (f *fragment) importSetValue(columnID uint64, bitDepth uint, value uint64,
...
+	// Execute once for positive numbers and once for negative. Subtract the
+	// negative sum from the positive sum.
+	var psum, nsum int64
 	for i := uint(0); i < bitDepth; i++ {
-		row := f.row(uint64(i))
-		cnt = row.intersectionCount(consider)
-		sum += (1 << i) * cnt
+		row := f.row(uint64(bsiOffsetBit + i))
+
+		psum += int64((1 << i) * row.intersectionCount(prow))
```
seebs: this is sort of brilliant.

a thing has occurred to me, though: for large BSI ranges, for large column counts, it seems to me that there's a risk of overflow.

should we perhaps be squashing these? both in terms of filtering out intersection counts before doing this multiply, and possibly later in comparing psum and nsum; if they're both greater than zero, we can reduce them both by the lower of the two values.

fragment.go
```diff
@@ -871,30 +947,41 @@ func (f *fragment) min(filter *Row, bitDepth uint) (min, count uint64, err error
...
 	}
 
-	for i := bitDepth; i > uint(0); i-- {
-		ii := i - 1 // allow for uint range: (bitDepth-1) to 0
-		row := f.row(uint64(ii))
+	// If we have negative values, we should find the highest unsigned value
+	// from that set, then negate it, and return it. For example, if values
+	// (-1, -2) exist, they are stored unsigned (1,2) with a negative sign bit
+	// set. We take the highest of that set (2) and negate it and return it.
+	if row := f.row(bsiSignBit).Intersect(consider); row.Count() != 0 {
```
seebs: we have a row.Any() now.

it might make sense to add a row.IntersectAny(), for that matter, although that's a lot more work.

fragment.go
```diff
@@ -904,25 +991,35 @@ func (f *fragment) max(filter *Row, bitDepth uint) (max, count uint64, err error
 		return 0, 0, nil
 	}
 
-	for i := bitDepth; i > uint(0); i-- {
-		ii := i - 1 // allow for uint range: (bitDepth-1) to 0
-		row := f.row(uint64(ii))
+	// Find lowest negative number w/o sign and negate, if no positives are available.
+	pos := consider.Difference(f.row(bsiSignBit))
+	if pos.Count() == 0 {
```
seebs: this could also be an Any().

fragment.go
```diff
@@ -972,22 +1078,44 @@ func (f *fragment) rangeNEQ(bitDepth uint, predicate uint64) (*Row, error) {
...
+		upredicate = uint64(-predicate)
+	}
+
+	// If predicate is positive, return all positives less than predicate and all negatives.
+	if (predicate >= 0 && allowEquality) || (predicate >= -1 && !allowEquality) {
+		pos, err := f.rangeLTUnsigned(b.Difference(f.row(bsiSignBit)), bitDepth, upredicate, allowEquality)
+		if err != nil {
+			return nil, err
+		}
+		neg := b.Intersect(f.row(bsiSignBit))
```
seebs: Do we need this intersection? In the set value code, I believe we're clearing the sign bit for any row that we're clearing the existence bit for. If so, the sign bit is already effectively intersected with the existence bit.

fragment.go
```diff
@@ -1001,32 +1129,54 @@ func (f *fragment) rangeLT(bitDepth uint, predicate uint64, allowEquality bool)
...
+	if (predicate >= 0 && allowEquality) || (predicate >= -1 && !allowEquality) {
+		return f.rangeGTUnsigned(b.Difference(f.row(bsiSignBit)), bitDepth, upredicate, allowEquality)
+	}
+
+	// If predicate is negative, return all negatives less than than upredicate and all positives.
+	neg, err := f.rangeLTUnsigned(b.Intersect(f.row(bsiSignBit)), bitDepth, upredicate, allowEquality)
+	if err != nil {
+		return nil, err
+	}
+	pos := b.Difference(f.row(bsiSignBit))
```
seebs: this has me pondering the desireability of an any-negative-values flag for the field, because if there's no negative values, we could save a lot of time here.

fragment.go
```diff
@@ -1036,68 +1186,102 @@ func (f *fragment) rangeGT(bitDepth uint, predicate uint64, allowEquality bool)
...
-		row := f.row(uint64(i))
+		row := f.row(uint64(bsiOffsetBit + i))
 		bit1 := (predicateMin >> uint(i)) & 1
 		bit2 := (predicateMax >> uint(i)) & 1
 
 		// GTE predicateMin
 		// If bit is set then remove all unset columns not already kept.
 		if bit1 == 1 {
-			b = b.Difference(b.Difference(row).Difference(keep1))
+			filter = filter.Difference(filter.Difference(row).Difference(keep1))
```
seebs: filter.Intersect(keep1)?

fragment.go
```diff
@@ -1036,68 +1186,102 @@ func (f *fragment) rangeGT(bitDepth uint, predicate uint64, allowEquality bool)
...
 		} else {
 			// If bit is unset then add columns with set bit to keep.
 			// Don't bother to compute this on the final iteration.
 			if i > 0 {
-				keep1 = keep1.Union(b.Intersect(row))
+				keep1 = keep1.Union(filter.Intersect(row))
```
seebs: UnionInPlace?

fragment.go
```diff
@@ -2266,6 +2450,59 @@ func (f *fragment) rows(start uint64, filters ...rowFilter) []uint64 {
...
+		return "", nil
+	}
+
+	other := roaring.NewBitmap()
+	other.Flags = roaringFlagBSIv2
+	if err := func() error {
+		f.mu.Lock()
+		defer f.mu.Unlock()
+
+		var err error
```
seebs: I like that this seems reasonably clear, but I think it's probably going to be fairly slow. Idle thought: We could move the *containers* rather than the *bits*. Each container present in the old system has a correct position in the new system; if it's the existence bit, it's reduced by ShardWidth*bitDepth, otherwise it's increased by 2*ShardWidth. If we have a lock on the storage (so we know no one's modifying it until we're done), we don't even need to clone the containers, we can just insert them in the new storage file at the correct positions.

roaring/roaring.go
```diff
@@ -4180,7 +4186,8 @@ func readOfficialHeader(buf []byte) (size uint32, containerTyper func(index uint
 		return newType
 	}
 	containerTyper = cf
-	cookie := binary.LittleEndian.Uint32(buf)
+	cookie := binary.LittleEndian.Uint32(buf) & 0xFFFFFF
+	flags = buf[3]
```
seebs: should we be reading flags from the "official" header? i thought this was basically upstream Roaring format, which we wouldn't want to interpret differently.

fragment.go
jaffee: As long as it's not going to take, like, hours... I don't think we should spend much effort on optimizing the upgrade code since it isn't going to run over and over.

seebs: that's probably fair. i don't actually have a clear sense of how large the data sets are or how long it'll take.

benbjohnson:
@jaffee @seebs I ended up refactoring the `Min`/`Max into `Base`/`BitDepth` which makes way more sense but there were a lot of affected sections of the code and tests so it took way longer than expected. I fixed up the other review comments & added a note in the docs about the upgrade.

Tests are passing. Can you please re-review when you get a chance?

seebs:
Okay, I hate to do this, but my first intuitive response is "but I think we *want* the options to be spelled as min/max, even if the internal implementation is different. I don't think we want to break existing config or spelling, and min/max can be useful expressively, as well as letting people set a cap -- if we wanted min 0, max 100, it would actually still be desireable to reject or truncate values outside of that range.

So I think we may want the min and max to exist... But then maybe compute a base/depth from them in some way? Maybe just compute a base midway between them when initially creating the field, then thereafter compute the effective depth based on min/max distance from that base.

benbjohnson:
> _I think we want the options to be spelled as min/max, even if the internal implementation is different...it would actually still be desirable to reject or truncate values outside of that range._

I think that's probably a product question. Min/max existed as an implementation constraint but trying to keep them seems counterintuitive to me. /cc @travisturner 

We don't have an effective minimum or maximum now and it doesn't seem like most people would try to fiddle with the `base` settings to optimize storage. I think the compression in `roaring` would likely improve the size of bitmaps that are in a similar range even without `base` being set (e.g. if all your values are between `10000` and `11000` then a lot of the same upper bits will be set).


> _So I think we may want the min and max to exist... But then maybe compute a base/depth from them in some way? Maybe just compute a base midway between them when initially creating the field, then thereafter compute the effective depth based on min/max distance from that base._

This seem like it's adding a lot complexity without much benefit. Honestly, I'd rather just remove `base` altogether. That alone creates a bunch of complexity even without calculating the average and bit depth from min/max.

benbjohnson:
@jaffee @seebs I re-added the min/max constraints, rebased from `master`, and separated the new changes into its own commit this time (71a28a2).

benbjohnson:
@jaffee I thought I had checks when I refactored but lo and behold they aren't there. I fixed everything up in 4080337. Can you give it a look through please?

PR #1903, opened 2019-03-18
Mmap fail

jaffee:
## Overview

This seems to work—I imagine'd 20 billion bits into a 5 node cluster with Pilosa's max map count set to 1000. Memory usage was reported normally (i.e. not like mmap). 85k fragment files on each node, 24ish GB on disk. 40GB memory used.

There are some questionable design decisions here (hello globals!), so fire away.

Fixes #1890 

## Pull request checklist

## Code review checklist

### DIFF
--- a/fragment.go
+++ b/fragment.go
@@ -41,6 +41,7 @@ import (
 	"github.com/pilosa/pilosa/pql"
 	"github.com/pilosa/pilosa/roaring"
 	"github.com/pilosa/pilosa/stats"
+	"github.com/pilosa/pilosa/syswrap"
 	"github.com/pilosa/pilosa/tracing"
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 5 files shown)


fragment.go
```diff
@@ -218,24 +219,29 @@ func (f *fragment) openStorage() error {
...
-	if err != nil {
-		return fmt.Errorf("mmap: %s", err)
-	}
-	f.storageData = storageData
+	} else {
+		// Mmap the underlying file so it can be zero copied.
+		data, err := syswrap.Mmap(int(f.file.Fd()), 0, int(fi.Size()), syscall.PROT_READ, syscall.MAP_SHARED)
+		if err == syswrap.ErrMaxMapCountReached {
+			f.Logger.Debugf("maximum number of maps reached, reading file instead")
+			data, err = ioutil.ReadAll(file)
```
seebs: This doesn't seem insane, but I wonder whether we should, in this case, unmap all the containers so we can drop the data object, since that's likely a large allocation.

jaffee: I think this would not be trivial at the moment—since unmapping would mean copying everything. Would be nice to have some kind of streaming variant of unmarshalbinary that creates unmapped containers, but I'm not sure it would ever be as performant as doing pointer magic into a big byte slice.

My future goals for this would be to have the ability to only read the header, and then fetch rows from storage on-demand and cache them.

Or something like that... more thinking needed.

syswrap/mmap.go
```diff
@@ -13,16 +14,25 @@ var mapCount uint64
...
+}
 
 // Mmap increments the global map count, and then calls syscall.Mmap. It
 // decrements the map count and returns an error if the count was over the
 // limit. If syscall.Mmap returns an error it also decrements the count.
 func Mmap(fd int, offset int64, length int, prot int, flags int) (data []byte, err error) {
-	if newCount := atomic.AddUint64(&mapCount, 1); newCount > MaxMapCount {
+	mu.RLock()
+	defer mu.RUnlock()
+	if newCount := atomic.AddUint64(&mapCount, 1); newCount > maxMapCount {
```
seebs: I was going to say, don't we need this lock also on the other end? But of course we don't, we're not checking maxMapCount on the unmap.

PR #1906, opened 2019-03-20
implement global open file counter using syswrap

### LINKED ISSUE #1905: Allow Pilosa to have more fragments than the open file handle limit.
### Description
Currently, Pilosa opens each fragment file, tries to mmap it, and falls back to reading it if the max-map-count has been reached. It leaves the file open in both cases and small writes may append to the file (using the op log functionality) rather than causing a full snapshot of the fragment.

We'd like to avoid keeping all of these files open all the time as Pilosa can run up against system limits on open file handles which can be difficult to change. The file handle of an mmapped file can actually be closed after it's mapped.

Once closed, we need to ensure that we reopen the file if we need to write to the op log.

It will be important to understand the performance and overhead implications of reopening and closing files for each update. 

The proposed path for this work to take is:
1. close all fragments after mmapping or reading them, and re-open for appends/snapshots.
2. test the performance of this implementation vs the previous for fragment heavy workloads (e.g. random writes to YMDH fields)
3. If performance testing shows a measureable difference, implement some tracking and a global maximum limit on open files imposed via Pilosa configuration. After the limit is reached, new fragments will be opened/closed on demand rather than kept open all the time.

Future work might be to have some kind of LRU eviction cache on open files so that the files most likely to be written to stay open while the others are closed.

Pilosa can add more fragments regardless of OS limits on file handles. Performance on files up to approximately the OS limit is not negatively affected.

jaffee:
close files after using them if global max is passed.

I originally implemented this without the global count—just always closing files
when done with them, and reopening for new writes. This was crazy slow for that
one test that uses mustSetBits in a big loop. I modified the test to use
importRoaring and everything worked better (though still slower).

After adding the global counter, I ran the tests with that one test using
mustSetBits again, and the performance was similar to master. After completing
this PR, I ran the tests with the max limit set to 5—they still passed but were
much slower.

Fixes #1905 

## Pull request checklist

## Code review checklist

### DIFF
--- a/fragment.go
+++ b/fragment.go
@@ -187,6 +187,18 @@ func (f *fragment) Open() error {
 	return nil
 }
 
+func (f *fragment) reopen() (mustClose bool, err error) {
+	if f.file == nil {
+		// Open the data file to be mmap'd and used as an ops log.
+		f.file, mustClose, err = syswrap.OpenFile(f.path, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0666)
+		if err != nil {
+			return mustClose, fmt.Errorf("open file: %s", err)
+		}
+		f.storage.OpWriter = f.file
+	}
+	return mustClose, nil
+}
+
 // openStorage opens the storage bitmap.
 func (f *fragment) openStorage() error {
 	// Create a roaring bitmap to serve as storage for the shard.
@@ -194,11 +206,14 @@ func (f *fragment) openStorage() error {
 		f.storage = roaring.NewFileBitmap()
 	}
 	// Open the data file to be mmap'd and used as an ops log.
-	file, err := os.OpenFile(f.path, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0666)
+	file, mustClose, err := syswrap.OpenFile(f.path, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0666)
 	if err != nil {
 		return fmt.Errorf("open file: %s", err)
 	}
 	f.file = file
+	if mustClose {
+		defer f.safeClose()
+	}
 
 	// Lock the underlying file.
 	if err := syscall.Flock(int(f.file.Fd()), syscall.LOCK_EX|syscall.LOCK_NB); err != nil {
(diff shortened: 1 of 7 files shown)


fragment_internal_test.go
```diff
@@ -2455,6 +2465,14 @@ func (f *fragment) mustSetBits(rowID uint64, columnIDs ...uint64) {
 	}
 }
 
+func addToBitmap(bm *roaring.Bitmap, rowID uint64, columnIDs ...uint64) {
+	for i, c := range columnIDs {
+		columnIDs[i] = rowID*ShardWidth + c%ShardWidth
+	}
+	positions := columnIDs
```
seebs: i do not see why `positions` exists, at this point. is it okay that addToBitmap mutates the contents of the slice passed to it? (this will only matter in explicit `slice...` cases, i guess.)

fragment.go
```diff
@@ -480,6 +513,13 @@ func (f *fragment) unprotectedSetBit(rowID, columnID uint64) (changed bool, err
 func (f *fragment) clearBit(rowID, columnID uint64) (bool, error) {
 	f.mu.Lock()
 	defer f.mu.Unlock()
+	err, mustClose := f.reopen()
```
seebs: this pattern is bugging me, but i can't see a good way around it. you can't have a function which does a defer-for-a-parent, but this feels like something where if we want to change this logic, we have to change it in too many places. i note that, in nearly all cases, it also follows a mutex lock/unlock pair. maybe something like

`err, defer := f.assureFile(mustLock)`, then if err is non-nil, we return it as an error. and after that check, if defer is non-nil, we call it. but... hmm. this still feels clunky. i was going to say "what if that all went into a wrapper", but of course, these ARE the wrappers around the `unprotected` methods.

i have not started a "review" because i'm sleepy-headed and want to read this again tomorrow.

fragment_internal_test.go
jaffee: I was just trying to make it more clear that columnIDs no longer represented column ids and was now full of positions. Guess I should just have added a comment :)

fragment.go
jaffee: I had to change it several times while developing and it was clunky....

for what it's worth, I have some refactoring in my private TODO list which will greatly reduce the number of places we need to do this. I think setBit, clearBit, and others can be trivially rewritten to use `importPositions`

fragment_internal_test.go
```diff
@@ -2455,6 +2465,14 @@ func (f *fragment) mustSetBits(rowID uint64, columnIDs ...uint64) {
 	}
 }
 
+func addToBitmap(bm *roaring.Bitmap, rowID uint64, columnIDs ...uint64) {
+	// we'll reuse the columnIDs slice and fill it with positions for DirectAddN
+	for i, c := range columnIDs {
+		columnIDs[i] = rowID*ShardWidth + c%ShardWidth
+	}
+	bm.DirectAddN(positions...)
```
jaffee:
```suggestion
	bm.DirectAddN(columnIDs...)
```

PR #1917, opened 2019-03-28
fix importRoaring perf regression [no changelog]

jaffee:
## Overview

We accidentally used sliceContainers when we should have been using BTree containers in importRoaring. Coupled with the UnionInPlace change where we union into whichever bitmap has more containers, this caused an absolute massive slowdown in the case where the incoming bitmap had more containers, and the existing bits were all or mostly in containers which came before the incoming ones in the bitmap. 

This causes massive amounts of copying/and resizing in sliceContainers.

A benchmark is included that shows an instance of this. It's actually significantly less stark of a difference than what occurred with the taxi import which is where we first noticed this issue.

## Pull request checklist

## Code review checklist

### DIFF
--- a/fragment.go
+++ b/fragment.go
@@ -1798,7 +1798,7 @@ func (f *fragment) importValue(columnIDs, values []uint64, bitDepth uint, clear 
 func (f *fragment) importRoaring(data []byte, clear bool) error {
 	f.mu.Lock()
 	defer f.mu.Unlock()
-	bm := roaring.NewBitmap()
+	bm := roaring.NewBTreeBitmap()
 	err := bm.UnmarshalBinary(data)
 	if err != nil {
 		return err
--- a/fragment_internal_test.go
+++ b/fragment_internal_test.go
@@ -743,7 +743,7 @@ func BenchmarkFragment_RepeatedSmallImports(b *testing.B) {
 							f := mustOpenFragment("i", "f", viewStandard, 0, "")
 							f.MaxOpN = opN
 							defer f.Clean(b)
-							err := f.importRoaring(getZipfRowsSliceRoaring(uint64(numRows), 1), false)
+							err := f.importRoaring(getZipfRowsSliceRoaring(uint64(numRows), 1, 0, ShardWidth), false)
 							if err != nil {
 								b.Fatalf("importing base data for benchmark: %v", err)
 							}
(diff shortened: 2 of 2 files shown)


fragment.go
```diff
@@ -1798,7 +1798,7 @@ func (f *fragment) importValue(columnIDs, values []uint64, bitDepth uint, clear
 func (f *fragment) importRoaring(data []byte, clear bool) error {
 	f.mu.Lock()
 	defer f.mu.Unlock()
-	bm := roaring.NewBitmap()
+	bm := roaring.NewBTreeBitmap()
```
seebs: This seems reasonable, but I also wonder: Should the default be changed such that NewBitmap makes the kind we want in the first place?

fragment_internal_test.go
```diff
@@ -2334,12 +2359,12 @@ func getUpdataInto(f func(row, col uint64) bool, numRows, numCols uint64, seed i
 
 // getZipfRowsSliceStandard is the same as getZipfRowsSliceRoaring, but returns
 // row and column ids instead of a byte slice containing roaring bitmap data.
-func getZipfRowsSliceStandard(numRows uint64, seed int64) (rowIDs, columnIDs []uint64) {
+func getZipfRowsSliceStandard(numRows uint64, seed int64, startFrom, endAt uint64) (rowIDs, columnIDs []uint64) {
```
seebs: these should use the same name, whether it's startCol or startFrom, I think.

fragment.go
jaffee: the kind we want, in theory, changes based on the usage. After briefly reviewing our usage, I think the safest thing to do would be to change the default... however, I'd like to look into it a bit more, run benchmarks, etc. Can we save it for another PR?

fragment_internal_test.go
jaffee: yeah, across 4 args I used 3 different conventions... maybe not the best

fragment.go
seebs: yeah, that seems reasonable. i am not 100% sure that we want to actually have multiple kinds configured at once, but on reflection, it's a good idea to do that and test it that way, because that way we don't risk breaking it so that we couldn't later if we had a good reason to. (i'm thinking about some fancy stuff for distinguishing probably-read-only data from data that expects updates).

PR #1918, opened 2019-03-28
importValue should only consider the last instance of a column id

jaffee:
included test demonstrates bug

I know we have some thoughts in the pipeline for a better approach to importValue, so I didn't try to do anything fancy here.

The map of column ids is only used on the smallwrite path, so hopefully is never large.

## Pull request checklist

## Code review checklist

### DIFF
--- a/fragment.go
+++ b/fragment.go
@@ -1731,6 +1731,42 @@ func (f *fragment) bulkImportMutex(rowIDs, columnIDs []uint64) error {
 	return errors.Wrap(f.importPositions(toSet, toClear, rowSet), "importing positions")
 }
 
+func (f *fragment) importValueSmallWrite(columnIDs, values []uint64, bitDepth uint, clear bool) error {
+	// TODO figure out how to avoid re-allocating these each time. Probably
+	// possible to store them on the fragment with a capacity based on
+	// MaxOpN. For now, we know that the total number of bits to be
+	// set+cleared is len(values)*(bitDepth+1), so we make each slice
+	// slightly more than half of that to try to avoid reallocation.
+	toSet := make([]uint64, 0, len(columnIDs)*int(bitDepth+1)*(5/8))
+	toClear := make([]uint64, 0, len(columnIDs)*int(bitDepth+1)*(5/8))
+	colSet := make(map[uint64]struct{}, len(columnIDs))
+
+	if err := func() (err error) {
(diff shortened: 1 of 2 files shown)


fragment.go
```diff
@@ -1753,6 +1754,7 @@ func (f *fragment) importValue(columnIDs, values []uint64, bitDepth uint, clear
 		// slightly more than half of that to try to avoid reallocation.
 		toSet = make([]uint64, 0, len(columnIDs)*int(bitDepth+1)*(5/8))
 		toClear = make([]uint64, 0, len(columnIDs)*int(bitDepth+1)*(5/8))
+		colSet = make(map[uint64]struct{})
```
seebs: `, len(columnIDs)`? We might not need them all, but we won't need more.

fragment.go
```diff
@@ -1762,9 +1764,13 @@ func (f *fragment) importValue(columnIDs, values []uint64, bitDepth uint, clear
 	// Process every value.
 	// If an error occurs then reopen the storage.
 	if err := func() (err error) {
-		for i := range columnIDs {
+		for i := len(columnIDs) - 1; i >= 0; i-- {
 			columnID, value := columnIDs[i], values[i]
 			if smallWrite {
+				if _, ok := colSet[columnID]; ok {
```
seebs: should we make the corresponding change in the non-smallWrite case? it seems like it's just as applicable there.

just as a sanity check: we *can't* have timestamps here, right? so we're not losing timestamp data about previous writes.

jaffee: we don't actually need it in the large write case, because it goes value by value, setting the bits directly, rather than collecting all the set bits and all the clear bits for all the values, and then processing all set bits followed by all clear bits.

I should actually test this case, though. Will do that.

No timestamps in BSI—maybe one day, but I guess we'll cross that bridge...

seebs: Even so, given how expensive writes can be, it might be worth dropping the writes that would be overwritten by later writes, if that's a common thing.

jaffee: I'm not inclined to trade a (potentially large) allocation for speeding up some workloads (which I think are fairly rare). This might warrant more research, but since we have grand plans for improving value imports overall, I'd rather get this bug fixed without adding anything new to the large write path.

seebs: good point.

PR #1938, opened 2019-04-10
validate (and panic) on duplicate PQL arguments

### LINKED ISSUE #1868: Row query allows two fields

### What's going wrong?
Row query is allowed for two or more fields
```
$ curl coordinator:10101/index/main/query -X POST -d 'Count(Row(f1=15, f2=58098, from="2019-01-01T00:00", to="2019-02-01T00:00"))'
{"results":[1180031292]}
```
and result of this query is something unknown - it's neither intersect nor union of these two rows...confusing... 

### What was expected?
Error that explains that two fields can't be used in one Row()

travisturner:
## Overview

I don't like that this is handled via a `panic` in the ast logic, but I couldn't see any other way to handle this without adding error handing to the methods on `Query` in `pql/ast.go`. Suggestions welcome.

Fixes #1868 

## Pull request checklist

## Code review checklist


pql/ast.go
```diff
@@ -112,6 +114,15 @@ func (q *Query) addField(field string) {
 	}
 }
 
+// validateArgField ensures that field does not already
+// exist as a key in the Args map before adding the new
+// key/value.
+func (q *Query) validateArgField(elem *callStackElem) {
+	if _, exists := elem.call.Args[elem.lastField]; exists {
+		panic(fmt.Sprintf("multiple instances of argument '%s' provided", elem.lastField))
```
jaffee: I forget... does a panic in the parser then manifest itself as an error?

jaffee: we don't want to panic though, do we? I'd think we would want to return an informative error to the user who issued the query with a duplicate field

travisturner: I agree, but is seems everything in `pql/ast.go` is handled with panics on malformed inputs.

Doing proper errors will require some refactoring. For this issue in particular, because the arguments are stored in a map, this is the only place where we can tell that a duplicate argument was provided. Effectively, it's "last one wins" because the map entry gets overwritten.

I think we could either introduce error handling here, or we could use a different data structure for the arguments (i.e. one that would allow duplicate keys).

jaffee:
Could we do something nasty with `recover` and doing string matching on error messages and then return well-formed errors from Parse?

travisturner:
you ask for nasty, you get bdc4f3b07eb7bb4b1eee3a5c148e58fc640e1509

PR #1970, opened 2019-05-10
Fix filter calls in GroupBy not being translated

### LINKED ISSUE #1967: Group by filter doesn't work with field string keys
### What's going wrong?
When running a group by query with a filter expression providing a string key for in the field returns the following error:

`{"error":"executing: executing group by filter for shard 0: Row() error with arg for row: could not convert x of type string to uint64 in Call.UintArg"}`

### What was expected?

A normal group count response with the results filtered based on the supplied expression, similar to:

`{"results":[[{"group":[{"field":"test","rowKey":"y"},{"field":"test2","rowKey":"y"}],"count":1}]]}`

1. Create an index with string keys
1. Create a couple of fields with string keys
1. Add some rows
1. Run a group by query similar to  `GroupBy(Rows(field='test'),Rows(field='test2'), filter=Row(test='x'))`

Version 1.2.0

Solo server on Docker in MacOS

jaffee:
## Overview

Fixes #1967 

## Pull request checklist

## Code review checklist

### DIFF
--- a/executor_test.go
+++ b/executor_test.go
@@ -3262,23 +3262,35 @@ func TestExecutor_GroupByStrings(t *testing.T) {
 	}
 
 	tests := []struct {
-		query string
+		query    string
+		expected []pilosa.GroupCount
 	}{
 		{
+			query: "GroupBy(Rows(generals))",
+			expected: []pilosa.GroupCount{
+				{Group: []pilosa.FieldRow{{Field: "generals", RowID: 1, RowKey: "r1"}}, Count: 5},
+				{Group: []pilosa.FieldRow{{Field: "generals", RowID: 2, RowKey: "r2"}}, Count: 5},
+			},
+		},
+		{
 			query: "GroupBy(Rows(generals), filter=Row(generals=r2))",
+			expected: []pilosa.GroupCount{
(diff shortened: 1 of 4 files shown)


executor_test.go
```diff
@@ -3243,7 +3243,56 @@ func TestExecutor_Execute_Query_Error(t *testing.T) {
...
+				Query: tst.query,
+			})
+			if err != nil {
+				t.Fatalf("got an error %v", err)
+			}
+			results := r.Results[0].([]pilosa.GroupCount)
+			test.CheckGroupBy(t, tst.expected, results)
+		})
+	}
 }
```
benbjohnson: I personally don't like table tests—they make all the error line numbers show up in the same line and they're just more difficult to debug in general. The actual test body isn't very large either so you could expand this into two separate subtests for probably the same number of lines.

jaffee: They are a bit easier when using the `t.Run` subtest functionality—*if* you give them a descriptive name. In most cases I name them with their index which works fine until you have more than 5 or so, and then it's a bit tough to figure out which one failed.

I like that they encourage adding new cases since it's pretty low effort, but I do think we could do a better job of naming each case to aid in debugging.

We should probably add some more verbiage in CONTRIBUTING around how to write tests, and how our tests are organized. We have enough table tests right now that I don't think we'd want to step away from that pattern, but 🤷‍♂ I'm open to it.

PR #1974, opened 2019-05-13
WIP: Improve row cache (mostly by not doing it)

seebs:
The row cache turns out to use a lot of memory, especially in read-heavy loads. On relatively write-heavy loads, the row cache gets flushed moderately frequently, implicitly, by the periodic snapshots. On read-heavy loads, though, it grows without any upper bound but "the number of rows you have". Furthermore, to avoid corruption or data races, it has *copies* of the data. This doesn't just double the storage requirements -- in the case where a lot of the data is mmapped from files, the mapped pages can be dropped by the kernel and re-read later, but copies made from that data  can't be dropped like that.

In order to address this, we adopt copy-on-write semantics. (Note: This isn't actually fully implemented yet.) The basic idea is that containers acquire Freeze and Thaw methods; Freeze gives you a container that you can trust no one else will modify, and Thaw gives you a container you can safely modify. As of now, Freeze sets a flag, and Thaw does a copy.

To make this work, all operations that might mutate containers now return new containers (or pointers to them, anyway), and things like the add/remove operations on bitmaps now stash updated containers back into the bitmaps. Missing, as of yet, is the part where, when putting something in the row cache, we freeze its contents. I knew I was forgetting something.

As a secondary component of this work, I've also implemented (partially) support for nil containers as valid (but innately-frozen) containers, which can be queried in various ways. I'm not totally sure that was actually necessary, but it fits with the general pattern implied by following an `append`-like rule. Basically, since you can do `s = append(s, ...)` for a nil slice, I wanted `c = c.add(v)` to work for a nil container.

Fixes # 1968


roaring/container_stash.go
```diff
@@ -35,28 +36,85 @@ const (
...
+	c := &Container{typp: containerArray, len: 0, cap: stashedArraySize}
 	c.pointer = (*uint16)(unsafe.Pointer(&c.data[0]))
 	return c
 }
 
 // NewContainerBitmap makes a bitmap container using the provided bitmap, or
 // an empty one if provided bitmap is nil. If the provided bitmap is too short,
 // it will be padded.
-func NewContainerBitmap(n int32, bitmap []uint64) *Container {
+func NewContainerBitmap(bitmap []uint64) *Container {
```
jaffee: we can't change this signature without a 2.0 release

roaring/container_stash.go
```diff
@@ -197,81 +451,71 @@ func (c *Container) setRuns(runs []interval16) {
...
 	c.pointer, c.len, c.cap = (*uint16)(unsafe.Pointer(h.Data)), int32(h.Len), int32(h.Cap)
 	runtime.KeepAlive(&runs)
 }
 
 // Update updates the container
-func (c *Container) Update(typ byte, n int32, mapped bool) {
-	c.typ = typ
-	c.n = n
-	c.mapped = mapped
+func (c *Container) Update(typ byte, n int32, mapped bool) *Container {
```
jaffee: can't change this w/o a 2.0

holder.go
```diff
@@ -158,7 +158,7 @@ func (h *Holder) Open() error {
 			continue
 		}
 
-		h.Logger.Printf("opening index: %s", filepath.Base(fi.Name()))
+		h.Logger.Printf("opening index: %s [rowache]", filepath.Base(fi.Name()))
```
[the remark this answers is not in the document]
seebs: that shouldn't even still be there, that was just my sanity check that `make provision-pilosa` was updating it. whoops. i've committed a change to revert that, but will likely later just nuke the commit that added it.

roaring/btree.go
```diff
@@ -876,6 +876,42 @@ func (e *enumerator) Next() (k uint64, v *Container, err error) {
 	return k, v, err
 }
 
+// Every iterates over a tree.
+func (e *enumerator) Every(upd func(oldV *Container, exists bool) (newV *Container, write bool)) error {
+	if err := e.err; err != nil {
```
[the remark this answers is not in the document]
seebs: Copying the idiom from Next and Prev, which are probably doing it because possibly in the past they were using naked returns with the named parameters? In any event, I'm inclined to keep the idiom parallel with the upstream ones.

I'm not sold on this function. I sort of want to make it take a func returning an error, and yield the error, but that makes it incompatible with the update funcs used with Put.

fragment.go
```diff
@@ -413,17 +413,16 @@ func (f *fragment) rowFromStorage(rowID uint64) *Row {
...
+	// again. (This may be implemented by copying, or by copy-on-write
+	// semantics.)
+	data.Freeze()
 	row := &Row{
 		segments: []rowSegment{{
-			data:     *data.Clone(),
+			data:     *data,
 			shard:    f.shard,
-			writable: false, // this Row will probably be cached and shared, so it must be read only.
+			writable: true,
```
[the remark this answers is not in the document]
seebs: that is the intent, but i'm actually not 100% sure it's right. the issue is that i don't have a consistent distinction in mind between "containers are frozen" and "bitmap is frozen". we sort of want/need a parallel construct for bitmaps, where the bitmap-itself is "frozen", meaning that things that would add or remove containers cause it to be cloned.

so on reflection, i may need to fix that. but that implies a larger API shift.

roaring/container_stash.go
```diff
@@ -35,28 +36,85 @@ const (
 // less than 4,096 values, an array is often used. Containers with long runs of
 // integers would use run length encoding, and more random data usually uses
 // bitmap encoding.
+//
+// the names "nnn" and "typp" are temporary hacks to make failure to update
```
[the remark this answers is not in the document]
seebs: yes, yes we should. they predate my realization that i specifically want those to never be accessed directly, because accessing them directly breaks nil containers.

roaring/container_stash.go
seebs: Yeah. I could in theory not change the NewContainerBitmap calls, but since we need the changes to other things (e.g., optimize, Update), we probably need the 2.0 API change.

roaring/roaring.go
```diff
@@ -156,6 +161,21 @@ func NewBitmap(a ...uint64) *Bitmap {
...
+// type, which performs better in cases where we expect a contiguous block of
+// containers added in ascending order, such as when extracting a range from
+// another bitmap.
+func NewSliceBitmap(a ...uint64) *Bitmap {
+	b := &Bitmap{
+		Containers: newSliceContainers(),
+	}
+	// TODO: We have no way to report this. We aren't in a server context
+	// so we haven't got a logger, nothing is checking for nil returns
+	// from this...
```
jaffee: since the bitmap's OpWriter is nil, `AddN` won't ever return a non-nil error (with the code as it is now), so this is semi-ok... would be nice to prevent future possible mishaps though. Maybe it should panic on error since that *should* be an impossible scenario?

roaring/container_stash.go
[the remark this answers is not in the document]
jaffee: just a reminder that these should be renamed before merging

roaring/roaring.go
seebs: Huh. Hadn't thought of that. I'm inclined to leave it for now, I think.

roaring/container_stash.go
seebs: Re-reading this: This one may be genuinely impossible to avoid, in the case where we call Update on a frozen container.

But thinking about it: I think that's only used from PutContainerValues, and in fact, we only ever call that from the initial roaring unmarshal. The problem is that PutContainerValues is just plain wrong in a lot of cases -- it's assuming it's safe to just overwrite those fields, so it produces inconsistent states, which are later fixed up when we read in the offsets and attach the mmapped data.

roaring/container_stash.go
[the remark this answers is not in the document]
seebs: okay, went through and did that. I removed every reference to them outside container_stash.go.

I sort of want to do an experiment where I split out the Container implementation into another package just to make sure I know what the interface to it is, but that wouldn't actually be a good design overall; the Container concept is pretty deeply tied to Roaring. I just want to make sure I'm not leaking things unintentionally.

roaring/container_stash.go
seebs: Updated to note: We don't need the API change! Update can just refuse to run on frozen things, possibly panicing if roaringParanoia is on, but also we can just not call it in the case that's affected, and instead make a new function that does what we need. Also reverted this change, and I think I'm back to no API breakage, although I'm not totally sure.

PR #1981, opened 2019-05-28
WIP: A half-baked initial plugin scheme

seebs:
## Overview

**WIP. Do Not Merge.**

This is a crufty and half-baked plugin scheme. It has a number of obvious flaws, but it's sufficient to allow me to add new PQL commands which are implemented reasonably efficiently.

The argument validation is basically garbage, there's a lot of loose ends, and so on.

Also, fixed the minor issue that the previous patch for the parser's argument validation broke Shift completely.

(This also adds a negative-shift option, which doesn't work 100% correctly for the same reasons the existing one doesn't, but which is needed to allow neighbors to work.)


seebs:
this needs some reworking and i don't need CI on it right now, so closing.

PR #1988, opened 2019-06-03
WIP: address some startup speed and performance issues

seebs:
## Overview

What if startup loading of fields and views and fragments were parallel? Also, what if we used the shiny new roaring iterator used by ImportRoaring and the remap feature for initial roaring imports?

There's some opportunities for further enhancements here by custom-hacking the btree implementation to perform better in the case where every newly inserted key is sequentially greater than the highest previous key, but it's probably not worth the effort.

### DIFF
--- a/field.go
+++ b/field.go
@@ -35,6 +35,7 @@ import (
 	"github.com/pilosa/pilosa/stats"
 	"github.com/pilosa/pilosa/tracing"
 	"github.com/pkg/errors"
+	"golang.org/x/sync/errgroup"
 )
 
 // Default field settings.
@@ -431,6 +432,8 @@ func (f *Field) Open() error {
 	return nil
 }
 
+var fieldQueue = make(chan struct{}, 16)
+
 // openViews opens and initializes the views inside the field.
 func (f *Field) openViews() error {
 	file, err := os.Open(filepath.Join(f.path, "views"))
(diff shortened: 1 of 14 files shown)


seebs:
Note that this is a shiny new implementation, and the old one in this branch has been mostly-abandoned. (It had very small impacts, this one is more like a 2x-10x improvement for most targets.)

field.go
```diff
@@ -431,6 +432,8 @@ func (f *Field) Open() error {
 	return nil
 }
 
+var fieldQueue = make(chan struct{}, 16)
```
jaffee: should we base this on NumCPU?

view.go
```diff
@@ -126,29 +131,47 @@ func (v *view) openFragments() error {
...
-			v.logger.Debugf("WARNING: couldn't use non-integer file as shard in index/field/view %s/%s/%s: %s", v.index, v.field, v.name, fi.Name())
-			continue
-		}
-
-		v.logger.Debugf("open index/field/view/fragment: %s/%s/%s/%d", v.index, v.field, v.name, shard)
-		frag := v.newFragment(v.fragmentPath(shard), shard)
-		if err := frag.Open(); err != nil {
-			return fmt.Errorf("open fragment: shard=%d, err=%s", frag.shard, err)
+	eg, ctx := errgroup.WithContext(context.Background())
+	var mu sync.Mutex
```
jaffee: more descriptive name? `shardMapMu` maybe? (applies to other similar mutexes)

roaring/roaring.go
```diff
@@ -1554,78 +1589,62 @@ func (b *Bitmap) ImportRoaringBits(data []byte, clear bool, log bool, rowSize ui
...
-		case containerArray:
-			c.setArray((*[0xFFFFFFF]uint16)(unsafe.Pointer(&data[offset]))[:c.N():c.N()])
-			opsOffset = int(offset) + len(c.array())*2 // sizeof(uint32)
-		case containerBitmap:
-			c.setBitmap((*[0xFFFFFFF]uint64)(unsafe.Pointer(&data[offset]))[:bitmapN:bitmapN])
-			opsOffset = int(offset) + len(c.bitmap())*8 // sizeof(uint64)
-		}
+	itrKey, itrCType, itrN, itrLen, itrPointer, itrErr = itr.Next()
+	for itrErr == nil {
+		newC := &Container{
```
jaffee: it seems like previously if the bitmap already had Containers we would reuse them instead of allocating new ones... now it seems like we're allocating a new one every time?

seebs: we didn't actually reuse them, Reset() was dropping them all.

view.go
seebs: i stuck with the local just because it's only used locally on a pretty small scale.

field.go
seebs: i thought about it some, but didn't convince myself in either direction. it can be a future improvement.

PR #1990, opened 2019-06-04
use os.Rename semantically correctly

seebs:
So it's true that Rename's arguments are called oldname/newname, and
you want to rename from the previous name to the new name.

And it's true that we're calling Rename on oldPath and newPath.

But in our case, oldPath is the name the fragment file had before
the operation, and newPath is the name of the temporary file
created during the operation.


seebs:
I agree that is probably clearer.

seebs:
Updated with corrected names, that is way clearer.

PR #1993, opened 2019-06-05
False Positive nodeLeave events put cluster in an unusable state (Starting)

tgruben:
## Overview

It has been published that hashicorp's Memberlist, which we use for gossip, can produce nodeLeave events when a node is under heavy load.  Hashicorp have since published a paper detailing a new layer on top of memberlist which significantly lowers the rate of false positive node failure events.  Hashicorp implemented the solution in an application called serf, which can be embedded into applications as a library.  This PR is just a short term fix and should migrate to serf if possible.


cluster.go
```diff
@@ -1687,13 +1691,28 @@ func (c *cluster) considerTopology() error {
 	return nil
 }
 
+// band aid to protect against false nodeLeave events from memberlist
+// the test is the lightest weight endpoint of the node in question /version
+// TODO provide more robust solution to false nodeJoin events
+func confirmNodeDown(uri URI) bool {
+	for i := 0; i < confirmDownRetries; i++ {
+		resp, err := http.Get(uri.Scheme + uri.HostPort() + "/version")
```
travisturner: is there a default timeout on this? we might not want that to be too long, right?

cluster.go
```diff
@@ -1711,11 +1730,15 @@ func (c *cluster) ReceiveEvent(e *NodeEvent) (err error) {
...
-				err = c.unprotectedSetStateAndBroadcast(c.determineClusterState())
+			if confirmNodeDown(e.Node.URI) {
+				if c.removeNodeBasicSorted(e.Node.ID) {
+					c.Topology.nodeStates[e.Node.ID] = nodeStateDown
+					// put the cluster into STARTING if we've lost a number of nodes
+					// equal to or greater than ReplicaN
+					err = c.unprotectedSetStateAndBroadcast(c.determineClusterState())
+				}
+			} else {
+				c.logger.Printf("received node leave: %v", e.Node)
```
travisturner: this log line duplicates line 1709/1728 above.
perhaps change the wording to indicate that the confirmation check found that the node is still available.

cluster.go
tgruben: I added a timeout of 2 secs but didn't make it configurable

PR #1995, opened 2019-06-08
Add pilosa shard-distribution subcommand

travisturner:
## Overview

This PR adds a subcommand `shard-distribution` to Pilosa that will return the shard placement based on various cluster parameters (index name, number of nodes, replication, max shard).

If a `--host` is provided, it will base its calculations off of an existing pilosa cluster. If not, it will start an in-memory cluster to determine the shard distribution.

This PR also adds a new API method (and corresponding client methods), so I would like some feedback on the decisions made with those.

## Pull request checklist

## Code review checklist

### DIFF
--- a/cmd/shard_distribution.go
+++ b/cmd/shard_distribution.go
@@ -45,7 +45,6 @@ Pilosa cluster.
 	flags.IntVarP(&shardDistributioner.NumNodes, "num-nodes", "n", 1, "Number of nodes in the cluster")
 	flags.IntVarP(&shardDistributioner.MaxShard, "max-shard", "m", 0, "Maximum shard to include (0-indexed)")
 	flags.IntVarP(&shardDistributioner.ReplicaN, "replicas", "r", 1, "Number of replicas to include")
-	flags.StringVarP(&shardDistributioner.Path, "output-file", "o", "", "File to write shard distribution to - default stdout")
 	ctl.SetTLSConfig(flags, &shardDistributioner.TLS.CertificatePath, &shardDistributioner.TLS.CertificateKeyPath, &shardDistributioner.TLS.SkipVerify)
 
 	return shardDistributionCmd
--- a/cmd/shard_distribution.go
+++ b/cmd/shard_distribution.go
@@ -1,0 +1,52 @@
+// Copyright 2017 Pilosa Corp.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"context"
+	"io"
+
+	"github.com/spf13/cobra"
+
+	"github.com/pilosa/pilosa/ctl"
+)
+
+var shardDistributioner *ctl.ShardDistributionCommand
+
+func newShardDistributionCommand(stdin io.Reader, stdout, stderr io.Writer) *cobra.Command {
+	shardDistributioner = ctl.NewShardDistributionCommand(stdin, stdout, stderr)
+	shardDistributionCmd := &cobra.Command{
+		Use:   "shard-distribution",
+		Short: "Shard distribution data from pilosa.",
+		Long: `
+Returns the shard distribution based on the provided arguments, or on a running
+Pilosa cluster.
+`,
+		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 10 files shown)


cmd/shard_distribution.go
```diff
@@ -0,0 +1,52 @@
...
+Returns the shard distribution based on the provided arguments, or on a running
+Pilosa cluster.
+`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return shardDistributioner.Run(context.Background())
+		},
+	}
+	flags := shardDistributionCmd.Flags()
+
+	flags.StringVarP(&shardDistributioner.Host, "host", "", "", "host:port of Pilosa")
```
yuce: I'm still reviewing the PR, but we use `--bind` to set the host:port with the `server` command. It would be better to keep that convention.

cmd/shard_distribution.go
```diff
@@ -0,0 +1,52 @@
...
+	flags.StringVarP(&shardDistributioner.Index, "index", "i", "", "Pilosa index to consider")
```
yuce: If a non-existent index is specified, we still get a non-error result. Should it return an error?

cmd/shard_distribution.go
```diff
@@ -0,0 +1,52 @@
...
+	flags.IntVarP(&shardDistributioner.NumNodes, "num-nodes", "n", 1, "Number of nodes in the cluster")
+	flags.IntVarP(&shardDistributioner.MaxShard, "max-shard", "m", 0, "Maximum shard to include (0-indexed)")
+	flags.IntVarP(&shardDistributioner.ReplicaN, "replicas", "r", 1, "Number of replicas to include")
+	flags.StringVarP(&shardDistributioner.Path, "output-file", "o", "", "File to write shard distribution to - default stdout")
```
yuce: I don't have a strong opinion about this, making this `output` would be a bit more concise.

yuce:
I've tested the PR on a 2 node cluster using custom built Pilosa (defaults) with the following:

`node1.toml`:
```toml
data-dir = "/home/yuce/ramdisk/cr1"
bind = "127.0.0.1:10101"

[gossip]
port = 14000

[cluster]
replicas = 2
coordinator = true

[metric]
diagnostics = false
```
`node2.toml`:
```toml
data-dir = "/home/yuce/ramdisk/cr2"                                                                                                                                                                                                           
bind = "127.0.0.1:10102"

[gossip]
port = 15000
seeds = "127.0.0.1:14000"

[cluster]
replicas = 2

[metric]
diagnostics = false
```

Created the schema:
```
$ curl localhost:10101/index/i1 -d ''
$ curl localhost:10101/index/i1/field/f1 -d ''
```
Then:
```
$ ./pilosa shard-distribution --host :10101 --index i1
[[0],[0]]
```
That's expected, but:
```
$ curl localhost:10101/index/i1/query -d 'Set(922337203685477580, f1=1)'
{"results":[true]}
$ ./pilosa shard-distribution --host :10101 --index i1
[[0],[0]]
```
I would expect the shard `9223372036854775808 >> 20 = 8796093022208` in the output. Is that wrong?

cmd/shard_distribution.go
jaffee: IMO, `bind` would imply that this command which we are running is going to bind to a port and run a server

cmd/shard_distribution.go
jaffee: no, we want to be able to ask Pilosa for the shard distribution of any index name

yuce: Well, then I guess it should return an empty response, like `[]`.

jaffee: I think it should return the shard distribution up to the given max shard

yuce: What would be the max shard of a non-existent index?

cmd/shard_distribution.go
travisturner: agree. also note that the `pilosa export` subcommand uses `--host`

cmd/shard_distribution.go
travisturner: you provide the max-shard (default is 0)
the idea being that you could say "if i were to have an index called "foo" with 1000 shards, how would those shards be distributed on a 5-node cluster?" in that case, it's ok that "foo" doesn't exist.

cmd/shard_distribution.go
travisturner: i meant to remove the `output-file` argument.
initially, i was going to have the option to send the output to a file, but i'm not exactly sure who/why that would be useful, so i removed it for now. but i forgot to remove this instance.

travisturner:
@yuce there are a few issues that your example raised, all of which I need to address:

- the http arg getter `q.Get("maxShard")` is not acting as I expected
- I need to add a test that exercises the http args
- for an existing index, the logic is not using `availableShards`, but rather just 0-maxShard based on the value from `api.MaxShards()`. This could cause a problem in your example where it will return all shards within that very large range. I'm actually not sure what the user would expect in this case.

And finally, there's a test that is failing in CI which I cannot recreate locally, so I'm still blocked on that.

PR #1997, opened 2019-06-10
handle insertions correctly

### LINKED ISSUE #1991: rowCache introduced possibly-broken Update logic in sliceContainers
sliceContainers has code to shuffle containers around which doesn't update keys in parallel. I don't think it gets hit easily, but it should be fixed, but I'm not smart enough to do it today.

seebs:
The "Update" case for Slice containers is broken, and can
insert a container without inserting a key. Fix this by using
the existing insert/add logic.

## Overview

Silly bug, that I don't think was actually affecting us yet, because I don't think the Update function is actually being hit in current code paths, but it's still probably worth fixing.

Fixes #1991.


hackskills:
nice catch with the keys. Looks good to me!

PR #1999, opened 2019-06-10
WIP: importRoaring behavior changes

seebs:
## Overview

ImportRoaring performance is sporadically unexpectedly bad. This attempts to fix a few of the cases. One is a pathological case which makes a factor-of-10+ difference in the performance of writes of exactly MaxOpN bits, the others are smaller fixes but apply in other cases as well.

### DIFF
--- a/fragment.go
+++ b/fragment.go
@@ -225,14 +225,24 @@ func (f *fragment) reopen() (mustClose bool, err error) {
 // possibly remap it from the file, but we don't need a full unmarshal, just
 // an update of mapped pointers.
 //
-// readData: Do we need to read/mmap this data?
-// unmarshalData: Do we need to discard the existing bitmap?
+// unmarshalData is somewhat overloaded. it tells us whether or not we
+// need to actually create a bitmap from the data (if the data exists to
+// do this from).
 //
 // usually unmarshalData is only set to false when we're in the middle of
 // a snapshot, and unprotectedWriteToFragment just wrote the in-memory data
 // out.
+//
+// If we have existing storage data, and we successfully get new data,
+// we will unmap the existing storage data.
+//
+// This function's design is probably a problem -- it is trying to handle
+// both cases where there was existing data before, and cases where we
+// just wrote the data.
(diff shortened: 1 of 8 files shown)


seebs:
Note, the change to Containers.Reset() is clearly a breaking change, so this would need rework. It produced a nice improvement, though.

fragment.go
```diff
@@ -2007,7 +2013,7 @@ func (f *fragment) importRoaring(ctx context.Context, data []byte, clear bool) e
 	f.mu.Lock()
 	defer f.mu.Unlock()
 	span.Finish()
-	bm := roaring.NewBTreeBitmap()
+	bm := roaring.NewSliceBitmap()
```
jaffee: I think this will cause poor performance in the case where we Union `f.storage` into `bm` instead of the other way around (see line 2072)

seebs: That is possible, but that should only happen when the new bitmap is the larger of the two, in which case the stuff we're unioning in is comparatively small. So I suspect that will be a rareish case, and/or a small case.

jaffee:
I think we had a test or benchmark for this though..... if the new data was
all higher columns than the existing data but was slightly larger or
something... I'll dig and see if I can find/remember.

fragment.go
seebs: it turns out that this analysis became completely wrong when i switched to the "just use the bitmap we just wrote to storage as the bitmap" thing, and produced some 1000% or larger performance regressions in this patch set, so.

seebs:
There's an actual much larger change in here which isn't obvious from the initial description, because it's new since I first drafted this:

We don't reload things after a snapshot anymore, in general.

Basically, the existing practice was: (1) write the in-memory bitmap out to disk, (2) close and unmap the existing file, (3) open and mmap the existing file, (4) drop the existing in-memory bitmap and unmarshal it from the new file.

In a reasonably typical ingest workload, with a lot of snapshots happening, this was only about 10-15% of CPU time directly, but it was about *50%* of all allocations, and thus, all garbage collector workload.

The next step would be improving this to provide a way to redo the mmapping cleanly.

PR #2004, opened 2019-06-13
Added fuzzing code and readme.md to explain the fuzzer

asvetlik:
## Overview

A fuzzer.go file is being added to fuzz the Bitmap.UnmarshalBinary. A readme.md to explain the fuzzer operations is also being added.

## Pull request checklist

- [x] I have read the [contributing guide](https://github.com/pilosa/pilosa/blob/master/CONTRIBUTING.md).
- [x] I have agreed to the [Contributor License Agreement](https://cla-assistant.io/pilosa/pilosa).
- [x] I have updated the [documentation](https://github.com/pilosa/pilosa/tree/master/docs).
- [x] I have resolved any merge conflicts.
- [x] I have included tests that cover my changes.
- [x] All new and existing tests pass.
- [x] Make sure PR title conforms to convention in CHANGELOG.md.
- [x] Add appropriate changelog label to PR (if applicable).

## Code review checklist
This is the checklist that the reviewer will follow while reviewing your pull request. You do not need to do anything with this checklist, but be aware of what the reviewer will be looking for.

- [ ] Ensure that any changes to external docs have been included in this pull request.
- [ ] If the changes require that minor/major versions need to be updated, tag the PR appropriately.
- [ ] Ensure the new code is [properly commented](https://github.com/golang/go/wiki/CodeReviewComments#doc-comments) and follows [Idiomatic Go](https://dmitri.shuralyov.com/idiomatic-go).
- [ ] Check that tests have been written and that they cover the new functionality.
- [ ] Run tests and ensure they pass.
- [ ] Build and run the code, performing any applicable integration testing.
- [ ] Make sure PR title conforms to convention in CHANGELOG.md.
- [ ] Make sure PR is tagged with appropriate changelog label.

PR #2012, opened 2019-06-14
Added a test code to test the fuzzer bugs and fixed 2 of the bugs found in roaring

### LINKED ISSUE #2005: Slice Bounds Out of Range Roaring
### What's going wrong? 
I ran a fuzzer, which revealed the error `panic: runtime error: slice bounds out of range`. It cites line 3969 and line 4512 in roaring.go as the source of the error.
line 3969 : `_, _ = h.Write(data[13 : 13+op.value*8])`
line 4512 : `for i, buf := 0, data[pos:]; i < int(keyN); i, buf = i+1, buf[4:] {`

The fuzzer identified the crashers to be: `:000000` and `<000000000000000`. It prints out:
`panic: runtime error: slice bounds out of range`

`goroutine 1 [running]:`
`github.com/pilosa/pilosa/roaring.readOffsets(0xc000072180, 0x4010000, 0xc, 0x200000,` `0xc, 0x1, 0xc, 0xc000090030)`
	`/Users/ashleysvetlik/go/src/github.com/pilosa/pilosa/roaring/roaring.go:4512 +0x772`
`github.com/pilosa/pilosa/roaring.(*Bitmap).UnmarshalBinary(0xc000072180, 0x4010000,` `0xc, 0x200000, 0xee98140, 0xd5c8b1768a4)`
	`/Users/ashleysvetlik/go/src/github.com/pilosa/pilosa/roaring/roaring.go:4501 +0x453`
`github.com/pilosa/pilosa/roaring.Fuzz(0x4010000, 0xc, 0x200000, 0x4)`

and:
`panic: runtime error: slice bounds out of range`

`goroutine 1 [running]:`
`github.com/pilosa/pilosa/roaring.(*op).UnmarshalBinary(0xc000090d90, 0x4810008, 0xd,` `0x1ffff8, 0xc000090c00, 0x1c)`
	`/Users/ashleysvetlik/go/src/github.com/pilosa/pilosa/roaring/roaring.go:3969 +0x836`
`github.com/pilosa/pilosa/roaring.(*Bitmap).unmarshalPilosaRoaring(0xc000072180,` `0x4810000, 0x15, 0x200000, 0x100b358, 0x30)`
	`/Users/ashleysvetlik/go/src/github.com/pilosa/pilosa/roaring/roaring.go:1171 +0xbc0`
`github.com/pilosa/pilosa/roaring.(*Bitmap).UnmarshalBinary(0xc000072180, 0x4810000,` `0x15, 0x200000, 0x3252e770, 0xd5c72e5f99b)`
	`/Users/ashleysvetlik/go/src/github.com/pilosa/pilosa/roaring/roaring.go:4477 +0x582`
`github.com/pilosa/pilosa/roaring.Fuzz(0x4810000, 0x15, 0x200000, 0x4)`

asvetlik:
## Overview

In this pull request, I added a test code to test the Bitmap.UnmarshalBinary, which evaluates the :000000 and <000000000000000 bugs the fuzzer found. I then figured out what was causing these two bugs and made the edits in the roaring.go. The <000000000000000 bug is a result of the input being too big, which causes it to become negative when converted to an int (its bigger than the max int). The :000000 bug is a result of the variable pos being equal to the length of the data, which causes problems when the UnmarshalBinary tries to run readOffsets.

Fixes #2005 

## Pull request checklist

## Code review checklist


roaring/README.md
```diff
@@ -0,0 +1,30 @@
+# The Fuzzer
+
+For complete documentation on go-fuzz, please see: https://github.com/dvyukov/go-fuzz
+
+The fuzzer in relation to the roaring package checks the `Bitmap.UnmarshalBinary` function found in `roaring.go`. In order to use the fuzzer, you can follow these steps:
+
+`cd $GOPATH/src/github.com/pilosa/pilosa/roaring`
```
alanbernstein: This whole instruction section is a good example of something that could be incorporated into a makefile. Maybe as an "install-gofuzz" target, following the convention set there. This could be a new ticket though.

roaring/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+`go-fuzz-build ./`
+
+You must now make the workdir/corpus directory. This is achieved by:
+
+`mkdir workdir/corpus`
+
+The fuzzer needs some input to start the fuzzing with. Copy some sample fragments into the workdir/corpus folder. Once you have copied your sample inputs, you are ready to run the fuzzer:
```
alanbernstein: I don't imagine we will have many (other) newcomers looking at this, but just in case, this could be a bit more clear. I'd say "Pilosa fragments", and maybe add an example line like `cp ~/.pilosa/my-index/my-field/views/standard/fragments/0 workdir/corpus".

roaring/roaring.go
```diff
@@ -3961,6 +3961,9 @@ func (op *op) UnmarshalBinary(data []byte) error {
 	_, _ = h.Write(data[0:9])
 
 	if op.typ > 1 {
+		if 1152921504606847000 < int(op.value) {
```
alanbernstein: This looks like a [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)) - if we go with this as the fix, it would be great if this were instead a well-named constant, defined with the appropriate arithmetic, or at least a comment explaining where it comes from.

roaring/roaring.go
```diff
@@ -3961,6 +3961,12 @@ func (op *op) UnmarshalBinary(data []byte) error {
 	_, _ = h.Write(data[0:9])
 
 	if op.typ > 1 {
+		// The maximum integer value for a int64 is 9223372036854775807
+		maxInt := 9223372036854775807
+		maxOpValue := maxInt/8-13
+		if maxOpValue < int(op.value) {
```
alanbernstein: This is an improvement, but there are a couple more things we can do:

- I'd prefer `maxInt64` over `maxInt`, but moreso, I would suggest directly using `math.MaxInt64`, (see https://golang.org/pkg/math/). Then we just have one line `maxOpValue := math.MaxInt64/8-13`, and the preceding comment is not really needed.
- We can move the definition of this outside the function, especially since there is already a similar value defined above (`minOpSize`), it can go right next to that.  

This is no longer relevant here, but just good to know: the value in the previous commit, `1152921504606847000` is not quite right. This looks like the rounded value you might get if you compute `MaxInt/8-13` in double precision floating point, rather than integer arithmetic. When working with 64-bit integers, this will often be a problem - be careful to use integer arithmetic operations in a situation like this, where you want both the input and the output to be an integer.

asvetlik: The op.value isn't set until like 3958. Therefore the check has to go after that, right?

alanbernstein: That's right. I might have made an unclear suggestion: the definition of the bound value `maxOpValue := math.MaxInt64/8-13` can be moved, while keeping the check in the same place. `maxBatchSize` is a slightly more descriptive name.

roaring/roaring.go
```diff
@@ -3956,6 +3957,10 @@ func (op *op) UnmarshalBinary(data []byte) error {
 	// op.value will actually contain the length of values for batch ops
 	op.value = binary.LittleEndian.Uint64(data[1:9])
 
+	if math.MaxInt64/8-13 < int(op.value){
```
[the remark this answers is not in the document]
asvetlik: Why should it be `(math.MaxInt64-13)/8` and not `math.MaxInt64/8-13`? The operation later on is `13+op.value*8` so shouldn't it be the opposite of that: `math.MaxInt64/8-13`?

roaring/README.md
asvetlik: I made these changes, but they were pushed to pull request #2004.

alanbernstein:
One more CI check - it looks like fuzzer.go also needs a license header

asvetlik:
The license check should be taken care of once the other pull request (#2004) is merged in. This pull request contains the stuff the other one deals with.

PR #2014, opened 2019-06-18
WIP: importRoaring more directly

seebs:
## Overview

This is a WIP (don't merge) that changes importRoaring to use a new unmarshalling importer in roaring, rather than unmarshalling the imported bits into their own new bitmap and trying to merge them that way. The intent is to reduce the amount of overhead from unioning, etc., because we don't have to process the whole bitmap, only the parts the new bitmap would interact with. (Outside of the first import to a fragment, this should always be better.)


seebs:
the failures in testing are not super surprising to me, this worked under light load but was not debugged well yet.

PR #2024, opened 2019-06-25
Unmarshal3

seebs:
## Overview

Various roaring/unmarshalling/memory-mapping/etc stuff.

This is an attempt to address a series of issues related to the performance of importRoaring.

In no particular order:
* We allow roaring import as Add and Remove operations in the ops log. This lets us write many more bits of ops log much more quickly.
* We use that import/export functionality also to implement importRoaring itself. In the process, we get rid of the CountRange things used to rebuild the cache, since we can determine the number of bits added or removed from a row nearly as cheaply as we can determine whether or not any bits changed.
* When reopening storage, instead of unmarshalling a new bitmap and ditching all the existing containers, we iterate through updating the pointers of containers that match the characteristics of the just-written data to point to the storage, and marking them as mapped. Any mapped containers that don't match get their data copied, so any previous data can be unmapped safely.
* We add a snapshot queue. Under heavy load, this will allow many more than 10k bits of write ops to get dumped into an ops log before a snapshot happens. Under very light load, it will result in writing the ops log, then producing a new snapshot, which slows things down, but that's only under very light load. Under heavy load, performance improves a bunch.
* The various tests, especially in fragment_internal_test, now verify on cleanup of a fragment that the on-disk representation matches the in-memory representation. This is expensive and slow but provides higher confidence that all the rest of this actually works as planned.

Tagging Ben because of the most recent patch -- when I added official roaring support to this, I noticed that our implementation is masking out some of the container-count bits from official roaring files that have run containers. I think this change is safe but I would like a second opinion on my reasoning about not needing the flag bits.

### DIFF
--- a/fragment.go
+++ b/fragment.go
@@ -2669,6 +2669,13 @@ func filterWithRows(rows []uint64) rowFilter {
 // this container have been processed. The rows accumulated up to this point
 // (including this row if all filters passed) will be returned.
 func (f *fragment) rows(start uint64, filters ...rowFilter) []uint64 {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	return f.unprotectedRows(start, filters...)
+}
+
+// unprotectedRows calls rows without grabbing the mutex.
+func (f *fragment) unprotectedRows(start uint64, filters ...rowFilter) []uint64 {
 	startKey := rowToKey(start)
 	i, _ := f.storage.Containers.Iterator(startKey)
 	rows := make([]uint64, 0)
@@ -3096,9 +3103,10 @@ func newRowsVector(f *fragment) *rowsVector {
 
 // Get returns the rowID associated to the given colID.
 // Additionally, it returns true if a value was found,
-// otherwise it returns false.
+// otherwise it returns false. Ensure that you already
+// have the mutex before calling this.
 func (v *rowsVector) Get(colID uint64) (uint64, bool, error) {
-	rows := v.f.rows(0, filterColumn(colID))
+	rows := v.f.unprotectedRows(0, filterColumn(colID))
 	if len(rows) > 1 {
 		return 0, false, errors.New("found multiple row values for column")
 	} else if len(rows) == 1 {
@@ -3129,9 +3137,10 @@ func newBoolVector(f *fragment) *boolVector {
 
 // Get returns the rowID associated to the given colID.
 // Additionally, it returns true if a value was found,
-// otherwise it returns false.
+// otherwise it returns false. Ensure that you already
+// have the fragment mutex before calling this.
 func (v *boolVector) Get(colID uint64) (uint64, bool, error) {
-	rows := v.f.rows(0, filterColumn(colID))
+	rows := v.f.unprotectedRows(0, filterColumn(colID))
 	if len(rows) > 1 {
 		return 0, false, errors.New("found multiple row values for column")
 	} else if len(rows) == 1 {
--- a/fragment.go
+++ b/fragment.go
@@ -107,12 +107,21 @@ type fragment struct {
 	shard uint64
 
 	// File-backed storage
-	path        string
-	flags       byte // user-defined flags passed to roaring
-	file        *os.File
-	storage     *roaring.Bitmap
-	storageData []byte
-	opN         int // number of ops since snapshot
+	path               string
+	flags              byte // user-defined flags passed to roaring
+	file               *os.File
+	storage            *roaring.Bitmap
+	storageData        []byte
+	totalOpN           int64 // total opN values
+	totalOps           int64 // total ops (across all snapshots)
+	opN                int   // number of ops since snapshot (may be approximate for imports)
+	ops                int   // number of higher-level operations, as opposed to bit changes
+	snapshotsRequested int   // number of times we've requested a snapshot
+	snapshotsTaken     int   // number of actual snapshot operations
+	snapshotting       bool  // set to true when requesting a snapshot, set to false after snapshot completes
+	snapshotCond       sync.Cond
+	snapshotDelays     int
+	snapshotDelayTime  time.Duration
 
 	// Cache for row counts.
 	CacheType string // passed in by field
@@ -145,11 +154,13 @@ type fragment struct {
 	mutexVector vector
 
 	stats stats.StatsClient
+
+	snapshotQueue chan *fragment
 }
 
 // newFragment returns a new instance of Fragment.
 func newFragment(path, index, field, view string, shard uint64, flags byte) *fragment {
-	return &fragment{
+	f := &fragment{
 		path:      path,
 		index:     index,
 		field:     field,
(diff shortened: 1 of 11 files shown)


roaring/roaring.go
```diff
@@ -3993,29 +4244,45 @@ func (op *op) apply(b *Bitmap) (changed bool) {
...
+	buf := make([]byte, op.encodeSize())
 
 	// Write type and value.
 	buf[0] = byte(op.typ)
-	if op.typ <= 1 {
+	switch op.typ {
+	case 0, 1:
 		binary.LittleEndian.PutUint64(buf[1:9], op.value)
-	} else {
+	case 2, 3:
```
jaffee: shouldn't we use the op names instead of numbers?

roaring/roaring.go
```diff
@@ -4429,6 +4726,75 @@ func xorBitmapRun(a, b *Container) *Container {
...
 }
 
+// CompareEquality is used mostly in test cases to confirm that two bitmaps came
+// out the same. It does not expect corresponding opN, or OpWriter, but expects
+// identical bit contents. It does not expect identical representations; a bitmap
+// container can be identical to an array container. It returns a boolean value,
+// and also an explanation for a false value.
+func (b *Bitmap) BitwiseEqual(c *Bitmap) (bool, error) {
+	biter, _ := b.Containers.Iterator(0)
+	citer, _ := c.Containers.Iterator(0)
```
jaffee: any reason we're using the Containers Iterator instead of the bitmap iterator?

roaring/roaring.go
```diff
@@ -4066,7 +4343,25 @@ func (op *op) size() int {
 	if op.typ == opTypeAdd || op.typ == opTypeRemove {
 		return 1 + 8 + 4
 	}
-	return 1 + 8 + 4 + len(op.values)*8
+	if op.typ == opTypeAddBatch || op.typ == opTypeRemoveBatch {
+		return 1 + 8 + 4 + len(op.values)*8
+	}
+	// else it's presumably roaring?
```
jaffee: seems like it'd be better to explicitly check this, and have another case for unknown op type... or just check it under roaringParanoia and panic since we don't return an error here

fragment.go
```diff
@@ -164,11 +175,68 @@ func newFragment(path, index, field, view string, shard uint64, flags byte) *fra
...
+			// wait forever, but notice that we're waiting
+			f.snapshotQueue <- f
+			f.snapshotDelays++
+			f.snapshotDelayTime += time.Now().Sub(before)
+			if f.snapshotDelays >= 10 {
+				f.Logger.Printf("snapshotting %s: last ten delays took %v", f.path, f.snapshotDelayTime)
+				f.snapshotDelays = 0
+				f.snapshotDelayTime = 0
+			}
+		case <-time.After(5 * time.Second):
```
jaffee: this case needs to go away now?

fragment.go
```diff
@@ -164,11 +175,68 @@ func newFragment(path, index, field, view string, shard uint64, flags byte) *fra
...
+		select {
+		case f.snapshotQueue <- f:
+		default:
+			before := time.Now()
+			// wait forever, but notice that we're waiting
+			f.snapshotQueue <- f
+			f.snapshotDelays++
+			f.snapshotDelayTime += time.Now().Sub(before)
+			if f.snapshotDelays >= 10 {
+				f.Logger.Printf("snapshotting %s: last ten delays took %v", f.path, f.snapshotDelayTime)
```
jaffee: might be good for this message to indicate that we're talking about enqueuing delay rather than actual snapshotting

fragment.go
```diff
@@ -2669,6 +2669,8 @@ func filterWithRows(rows []uint64) rowFilter {
 // this container have been processed. The rows accumulated up to this point
 // (including this row if all filters passed) will be returned.
 func (f *fragment) rows(start uint64, filters ...rowFilter) []uint64 {
+	f.mu.Lock()
```
jaffee: RLock?

roaring/roaring.go
```diff
@@ -1145,6 +1145,17 @@ type baseRoaringIterator struct {
 	lastErr           error
 }
 
+// okay, then
+func (b *baseRoaringIterator) SilenceLint() {
+	// these are actually used in pilosaRoaringIterator or officialRoaringIterator
+	// but structcheck doesn't know that
```
jaffee: so structcheck doesn't handle the embedding properly?

roaring/btree.go
```diff
@@ -894,7 +894,7 @@ func (e *enumerator) Next() (k uint64, v *Container, err error) {
 }
 
 // Every iterates over a tree.
-func (e *enumerator) Every(upd func(oldV *Container, exists bool) (newV *Container, write bool)) error {
+func (e *enumerator) Every(upd func(key uint64, oldV *Container, exists bool) (newV *Container, write bool)) error {
```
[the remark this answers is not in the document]
seebs: Hmm. Well, if we ever exported an enumerator object, even if we didn't export the type, it sorta would. But also Every hasn't been there for long.

roaring/roaring.go
seebs: ... yes.

roaring/roaring.go
seebs: I didn't think of it. Although I think I prefer comparing by containers, because the existence of an unmatched non-empty container is a simple answer, and container equality is reasonably easy to check.

roaring/roaring.go
seebs: Good catch.

fragment.go
seebs: ... i was actually gonna file an issue for "we should use an RWMutex for this so we can RLock here", because I apparently never checked.

roaring/roaring.go
seebs: yeah.

seebs:
I absolutely concur with the complexity of openStorage, and want to rethink it at some point. I think it's actually correct now, but I don't *like* it.

But also, for instance, if we're *not* unmarshalling data, and didn't mmap it, we don't really need to do the ReadAll operation. And even if we are, if it's not mmapped, I think we're better off not treating it as mapped storage.

So basically, I want to revisit that, but part of revisiting it is probably replacing the implementation of `(*Bitmap) UnmarshalBinary(...)` to distinguish whether or not it's mmapped.

But this also feels like something I'll wreck if I try to do it right now, so.

benbjohnson:
I did an initial read and it's looking good overall. I need to go through the `openStorage()` and `closeStorage()` again. I didn't quite wrap my head around those yet. :)

seebs:
> I did an initial read and it's looking good overall. I need to go through the `openStorage()` and `closeStorage()` again. I didn't quite wrap my head around those yet. :)

I am not sure I did yet either. One of my upcoming goals will be to rewrite that stuff more clearly.

roaring/btree.go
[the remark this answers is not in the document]
jaffee: oh right... Every has been added since 1.3 so it doesn't matter

jaffee:
I'm gonna get this rebased and merged before it gets out of date. We're definitely still pending more testing before cutting a release with this code. @benbjohnson if you end up having any more comments on the open/close storage stuff, please do add them.

PR #2034, opened 2019-07-12
Worker pool

### LINKED ISSUE #1969: Queries interfere with gossip, causing cluster to die

### What's going wrong?

It seems that when we start querying cluster - it interferes\contends for resources with nodes ability to answer gossip probes and cluster jumps from NORMAL to STARTING and back frequently, causing queries to error out.
it is not caused by network being a bottleneck since traffic is low, around 50kbs, which is nothing for available capacity.

### What was expected?

Queries should not interfere with cluster ability to communicate. Think gossip communication and cluster ability to stay alive should come first when distributing available resources, would be nice to have a handle on this

my gossip configs are:
```
[gossip]
  key = \"\"
  port = \"14000\"
  probe-interval = \"4s\"
  probe-timeout = \"2s\"
  push-pull-interval = \"30s\"
  seeds = [\"10.8.222.185:14000\",\"10.8.249.29:14000\"]
  stream-timeout = \"10s\"
  suspicion-mult = 4
  to-the-dead-time = \"30s\"
```
Included merged logs from my 7 boxes ( started queries at 10:58 - immediately see issues with gissip, cluster starts to jump back and forth between states, this behavior persists until query load stops ):

[logs.log](https://github.com/pilosa/pilosa/files/3164866/logs.log)

commit - https://github.com/pilosa/pilosa/commit/5bcb00e11a746a988dfbca3f5b8c55d3c21fbd54

jaffee:
## Overview

    add worker pool to executor for local query processing

    Pilosa previously spawned a goroutine for each remote node that a
    query needed to be forwarded to, and then forwarded a single request
    containing all the shards that the query should operate on. It then
    spawned a goroutine *per local shard* to process the query
    locally. This was fine if there weren't too many shards, or too many
    queries coming in concurrently, but we found that it created issues
    when there were 100s or 1000s of shards per node, and dozens of
    queries arriving concurrently.

    Specifically, the memberlist "hiccup" issue is highly correlated with
    many goroutine scenarios, and after applying this patch, memberlist
    complaints in the logs were much decreased, and nodeLeave events under
    concurrent query load almost entirely eliminated.

    This patch creates a fixed size pool of goroutines to do local shard
    processing, and passes work to them through a channel, one job per
    query per shard. Handling of remote requests (forwarding queries) is
    unchanged.

    We set the pool size to NumCPU()+8 somewhat arbitrarily, but this
    seemed to work pretty well in our testing on 32 core machines. It's a
    pretty big improvement over launching a goroutine per shard per query
    which is what we were doing previously, so we can tune it more later
    if necessary.

I think I still need to add some kind of executor.Close() method to make sure we aren't leaking goroutines in tests.

Naming is totally up for debate.

Capacity of channels and number of workers hasn't been super well scienced, but probably won't make much of a difference.

Fixes #1969 (probably)

## Pull request checklist

## Code review checklist

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"runtime"
 	"sort"
 	"sync"
 	"time"
@@ -58,8 +57,9 @@ type executor struct {
 	// Stores key/id translation data.
 	TranslateStore TranslateStore
 
-	workersWG sync.WaitGroup
-	work      chan job
+	workersWG      sync.WaitGroup
+	workerPoolSize int
+	work           chan job
 }
 
 // executorOption is a functional option type for pilosa.Executor
@@ -72,16 +72,18 @@ func optExecutorInternalQueryClient(c InternalQueryClient) executorOption {
 	}
 }
 
+func optExecutorWorkerPoolSize(size int) executorOption {
+	return func(e *executor) error {
+		e.workerPoolSize = size
+		return nil
+	}
+}
+
 // newExecutor returns a new instance of Executor.
 func newExecutor(opts ...executorOption) *executor {
-	// this is somewhat arbitrary, though going less than
-	// runtime.NumCPU() would likely result in a loss of throughput.
-	workerPoolSize := runtime.NumCPU() + 8
 	e := &executor{
-		client: newNopInternalQueryClient(),
-
-		// capacity of this channel is unlikely to affect much
-		work: make(chan job, workerPoolSize),
+		client:         newNopInternalQueryClient(),
+		workerPoolSize: 2,
 	}
 	for _, opt := range opts {
 		err := opt(e)
@@ -89,7 +91,12 @@ func newExecutor(opts ...executorOption) *executor {
 			panic(err)
 		}
 	}
-	for i := 0; i < workerPoolSize; i++ {
+	// this channel cap doesn't necessarily have to be the same as
+	// workerPoolSize... any larger doesn't seem to have an effect in
+	// the few tests we've done at scale with concurrent query
+	// workloads. Possible that it could be smaller.
+	e.work = make(chan job, e.workerPoolSize)
+	for i := 0; i < e.workerPoolSize; i++ {
 		e.workersWG.Add(1)
 		go func() {
 			defer e.workersWG.Done()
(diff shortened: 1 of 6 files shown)


executor.go
```diff
@@ -71,13 +74,17 @@ func optExecutorInternalQueryClient(c InternalQueryClient) executorOption {
 func newExecutor(opts ...executorOption) *executor {
 	e := &executor{
 		client: newNopInternalQueryClient(),
+		work:   make(chan job, 2000),
```
travisturner: does the size of this buffer ever present a problem? i guess put another way, is there any downside to blocking the writes to this channel when the buffer is full? also, does changing the buffer size affect anything?

jaffee: I set it high for my testing to try to eliminate it as a variable. The more I think about it, the more I realize I don't know that much about channel performance characteristics.

There are going to be QUERY_CONCURRENCY goroutines putting things into this channel, and each one is going to put SHARDS_PER_NODE jobs into it per query in a tight loop. There are POOL_SIZE workers reading from it spending a widely variable amount of time on each job, but often ~1ms.

A large buffer means that the query handler goroutines which are populating the channel get to dump all their work in, and then wait on the result chan. A smaller channel means they might have to wait to dump their work into the work chan, but wait for less time on the result chan. So... setting this channel large is probably a waste of memory. Could it be totally unbuffered? Maybe... my instinct is it won't have much effect. I will test with setting it to POOL_SIZE, and unbuffered, and see what we get.

seebs:
The only thing that's concerning me (and we currently can't hit it) is that when we have cross-index queries, i think it becomes possible for all the workers to be blocked waiting for sub-queries that are in turn blocked waiting for workers. To fix that, I think: whenever a worker needs to make remote queries for some reason (for instance, cross-index queries, which I think would be the only case), it should probably do something where it launches them, then waits for them, but while waiting it runs other jobs from the work queue.

But I don't think we can or should fix that now, because we haven't got a way to verify the functionality until the cross-index queries are implemented and we can build a test case for it.

I would also suggest making the worker pool size configurable in some way.

jaffee:
I don't want to expose this config externally, but I did need to sort-of expose it to allow us to set it lower when creating test clusters. My thinking is that because we will likely need to revisit this as we implement Joins and other more complex queries, this will likely not be the right thing to expose, and changes to the config file seem like one more point of friction to avoid when doing upgrades, etc.

Issue #2011, opened 2019-06-13
Key Translation Version 2


benbjohnson:
## Overview

Key translation currently supports the mapping of string keys to an autoincrementing integer value that allows us to map onto Roaring bitmaps. The basic functionality works well but there are several limitations in the current implementation:

1. Monolithic design—all keys are stored in a single file and not broken up by index or field.
2. Slow initialization—initialization of the key file is single threaded.
3. Scaling—the key translation file is fully replicated to all nodes in the cluster.

## Current Design

The interface to the store currently exists as:

```go
type TranslateStore interface {
	TranslateColumnsToUint64(index string, values []string) ([]uint64, error)
	TranslateColumnToString(index string, values uint64) (string, error)

	TranslateRowsToUint64(index, field string, values []string) ([]uint64, error)
	TranslateRowToString(index, field string, values uint64) (string, error)

	Reader(ctx context.Context, off int64) (io.ReadCloser, error)
}

```

This provides translation called for both the index (column) and field (row). Translation from keys to integers can be done in bulk while translation from integers to keys must be done one at a time. This design was because of the nature of the caller to the store and the aim to reduce allocations for slices to and from the calls.

The `Reader()` provides a log-based stream of all changes so that replica stores can maintain the same data asynchronously.

### Per-index, per-field stores

Instead of separate column & row translation calls, the new proposed store would simply have a single pair of translation calls. There would then be one separate store for each index & field.

```go
type TranslateStore interface {
	TranslateToID(key string) (uint64, error)
	TranslateToKey(id uint64) (string, error)
}
```

We originally stored translation data in Badger so the duplicate `index` and `field` did not affect storage size because duplicate prefixes were trimmed internally. However, after moving to our current log-based RHH design, the `index` and `field` data does cause a lot of duplicate data. This new design should eliminate that issue.

By creating separate per-index & per-field stores, we can also drop data in bulk when an index or field is deleted. See [#1976](https://github.com/pilosa/pilosa/issues/1976).

### Change underlying storage

The current log-based RHH design has two advantages:

1. Logs can easily be replicated to replica nodes as a stream.
2. RHH is very fast and lookups are O(1).

However, as seen in [#1807](https://github.com/pilosa/pilosa/issues/1807), there is an upper limit to the size we can store. Logs are slow to replay since they are single threaded and memory still grows with size with our current RHH approach.

[...]

PR #2042, opened 2019-07-20
[WIP] TranslationStore Refactor

benbjohnson:
## Overview

This pull requests refactors the `TranslateStore` to be per-index & per-row and also moves the store implementation from `rhh` to `boltdb`.

_This is still a work in progress but it's already become a large PR so I thought I'd push something to get an initial review & see if anybody had ideas for breaking this up smaller._

## Pull request checklist

## Code review checklist


jaffee:
sorry Ben, I've been on vacation but will take a gander as well. don't let
me hold you up.

benbjohnson:
I pushed up some more changes that wires up a lot of the remaining pieces. There's still some outstanding replication test failures that I'm resolving.

benbjohnson:
I wanted to drop a note about progress here. The multiplexing of the translation streams has taken longer than expected but it's moving along. I added a replication manager on the `Holder` to manage the replica state (`holderTranslateStoreReplicator`) and I added a multiplexer on the primary side to consolidate all the index & field translation readers into one (`MultiTranslateEntryReader`). 

Since each index and field has their own offset now there's also a `TranslateOffsetMap` which tracks this and allows the remote to request all the streams at the appropriate states in one call. I have unit tests added to test these pieces but I'm working on the integration tests now.

One outstanding issue is what happens when an index or field is created/deleted. My current approach is to just stop replication and refetch using the most recent offset map. This mostly works well (assuming that we aren't creating/deleting indexes & fields constantly). There is some timing issues, however, to ensure that the replica doesn't request from the primary until the new index or field exists.

jaffee:
@benbjohnson I've been going through this (sorry I didn't get to it sooner!), and overall, I think it looks great. This approach seems to have simplified things for the most part, and I'm excited about the scalability wins.

One thing we discussed in https://github.com/pilosa/pilosa/issues/2011 that I'm not seeing/understanding is how the sharding of column ID translation is handled. I'm perfectly fine with this being tackled in a future PR, but just wasn't sure if I was missing something. Thanks!

benbjohnson:
@jaffee @travisturner I finished refactoring the translate replication and it seems to be working now. There are a few failing tests but they seem to be related to startup and not the translate store specifically. Can you all give this PR a review please? I'll get those few outstanding tests wrapped up.

benbjohnson:
I have all the tests passing now with f42b8af. There are, however, some old tests that seem to be flaky and are related to replication timing. I've fixed some of them by adding a `test.RetryUntil()` function but I'm still working through the others as they come up.

I need to rebase now that everything's working but after that it should be good to go. 🚀

jaffee:
wooohoo!

jaffee:
We should actually talk about how we want to go about incorporating these changes since they necessitate a 2.0 release from a Go API perspective (and from a data format perspective). I think we should merge it to master so this becomes the default tree that we're working on, but maybe add a 2.0-alpha tag or something, so there's a visual indication in the commit history and generated Pilosa version string of when 2.0 started happening? @codysoyland would love your input here (and of course yours as well @benbjohnson).

In some cases (e.g. if translation is not being used), it may be possible to upgrade seamlessly. In others, I think we should at least detect that the translate file is in the wrong place/format, and give a reasonable error (like, "you want 2.0? you can't handle 2.0. please downgrade to 1.x")

I also have had a tiny inkling that we could maybe pretty easily write a small utility to do an offline conversion from the old translatefile style to the new one. The idea being that a lot of the complexity is removed by saying: 
1. turn off Pilosa
2. backup your data directory
3. run this utility on your data directory
4. if it finishes happily, start Pilosa 2.0, if not restore your backup and file an issue.

...rather than needing to manage distributed state and recovery in the case of crashes or whatever, but it's possible that I don't have a good handle on what all could be involved. We definitely don't need to do anything like this before getting this merged.

benbjohnson:
@jaffee I have this all working except for one flaky test—`TestClusterTranslator`. The issue is that the translation lives on the primary node now so if it executes  `Set()` against the primary node then it passes but if it queries against the replica then it doesn't replicate back to the primary.

Anyway, I'm adding a redirect now so the replica will send the query to the primary, however, the next step of the translation store is to shard it so I don't want to put too much effort into it since it'll get replaced.

http/handler.go
```diff
@@ -306,12 +306,11 @@ func newRouter(handler *Handler) *mux.Router {
...
 	router.HandleFunc("/internal/fragment/nodes", handler.handleGetFragmentNodes).Methods("GET").Name("GetFragmentNodes")
 	router.HandleFunc("/internal/index/{index}/attr/diff", handler.handlePostIndexAttrDiff).Methods("POST").Name("PostIndexAttrDiff")
+	router.HandleFunc("/internal/translate/data", handler.handlePostTranslateData).Methods("POST").Name("PostTranslateData")
 	router.HandleFunc("/internal/index/{index}/field/{field}/attr/diff", handler.handlePostFieldAttrDiff).Methods("POST").Name("PostFieldAttrDiff")
 	router.HandleFunc("/internal/index/{index}/field/{field}/remote-available-shards/{shardID}", handler.handleDeleteRemoteAvailableShard).Methods("DELETE")
 	router.HandleFunc("/internal/nodes", handler.handleGetNodes).Methods("GET").Name("GetNodes")
 	router.HandleFunc("/internal/shards/max", handler.handleGetShardsMax).Methods("GET").Name("GetShardsMax") // TODO: deprecate, but it's being used by the client
-	router.HandleFunc("/internal/translate/data", handler.handleGetTranslateData).Methods("GET").Name("GetTranslateData")
-	router.HandleFunc("/internal/translate/keys", handler.handlePostTranslateKeys).Methods("POST").Name("PostTranslateKeys")
```
jaffee: oog, I did not notice this change before... I don't suppose there's any way we could continue to support the `/internal/translate/keys` endpoint in a backwards-compatible way? We can update the clients if needed, of course, but it'd be nice to avoid it if possible.

jaffee: with this latest change, I'm now seeing `translate keys: index not found` when I run an ingest which uses this endpoint. (same code works against master)

jaffee: nice, things seem to be working after that last push

benbjohnson:
@jaffee I pushed in a `Holder.hasV1TranslateKeysFile()` method and check on it `Open()`. I need to do some integration testing after all those changes I pushed up but tests are finally passing and hopefully this should be nearly ready to merge.

jaffee:
yeah, that's working for me, thanks! I'll try to give this another good pass, but I think we can probably get it merged when you're ready.

PR #2048, opened 2019-07-25
Add a worker pool for importRoaring jobs

### LINKED ISSUE #2047: Implement a worker pool for ingest jobs
### Description

Implement a working pool on `ImportRoaring` in order to control the number of import goroutines active at any one time. Similar to #2034 on the query side.

An ingest work pool has been implemented.

travisturner:
## Overview

This PR adds a worker pool to handle `importRoaring` jobs.
I'm not sure if having this as part of `API` is correct. Because `API` isn't really thought of as a running object (and doesn't have anything like `close()`) there may be a problem with closing this out while jobs are in flight.

Fixes #2047 

## Pull request checklist

## Code review checklist

### DIFF
--- a/api.go
+++ b/api.go
@@ -42,6 +42,9 @@ type API struct {
 	cluster *cluster
 	server  *Server
 
+	importWorkerPoolSize int
+	importWork           chan importJob
+
 	Serializer Serializer
 }
 
@@ -58,9 +61,18 @@ func OptAPIServer(s *Server) apiOption {
 	}
 }
 
+func OptAPIImportWorkerPoolSize(size int) apiOption {
+	return func(a *API) error {
+		a.importWorkerPoolSize = size
+		return nil
+	}
+}
+
 // NewAPI returns a new API instance.
 func NewAPI(opts ...apiOption) (*API, error) {
-	api := &API{}
+	api := &API{
+		importWorkerPoolSize: 2,
+	}
 
 	for _, opt := range opts {
 		err := opt(api)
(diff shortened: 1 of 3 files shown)


jaffee:
I think we do need to add a Close() on API or somehow bubble the job queue into something which has a Close method. As we see from the race test failure, just leaking the goroutines isn't really an option.

At this point, I don't really have a problem adding a `Close() error` to API.

The other problem that you pointed out is that this is just for ImportRoaring. This would be fine, except it doesn't include importing integer values right now. My instinct is that it would be fairly easy (if a bit messy) to have the pool handle both job types. Ultimately, I'd like to get to the point where we're using import roaring for value(bsi) imports though.

seebs:
I think there's another subtlety here, which is I believe we have conflicts between ImportRoaring and existence columns. Didn't that imply something like not supporting roaring imports when we have existence? Or maybe I'm confused.

travisturner:
@jaffee `api.ImportValue()` doesn't currently spawn a goroutine to run `field.importValue()` the way that `ImportRoaring` does. Does is still make sense to modify `ImportValue` to use the worker pool? I ask because I thought the goal of this was to reduce the number of goroutines.

jaffee:
My theory is that this is more about the number of runnable goroutines rather than the total number. If there are hundreds or thousands of goroutines that have work to do (e.g. aren't waiting on I/O or channels or whatever), the scheduler has to do more work, and memberlist goroutines may see much less run time.

Ingest definitely isn't as a big an issue as the query side though since each query was launching numShards+numNodes goroutines before, whereas each ingest is just numReplicas. However, with dozens, or hundreds of concurrent ingesters, it still may be a problem. The idea with the pool would be to convert hundreds of concurrent ingester goroutines, each trying to do it's own piece of work to NumCPU goroutines doing the actual work, and the hundreds just waiting to be woken up by results on the channel.

We've seen before though that excess memory allocation seems to play into the memberlist stability issues, so it could be that we need to figure out how to make ingest allocate less memory which would be a separate issue.

My gut right now is to leave this as-is since the use cases we're looking at which trigger the memberlist issues don't have many integer fields anyway. We can get a feel for whether this helps before we go much further.

PR #2051, opened 2019-07-31
Refactor consistent hashing implementation

### LINKED ISSUE #2035: Revisit consistent hashing implementation for shard distribution
Pilosa currently uses the [Jump Consistent Hashing Algorithm](https://arxiv.org/pdf/1406.2294.pdf) to distribute shards to nodes in the cluster. 

The main strength of using a consistent hash is that it minimizes the amount of data transfer when a node is added or removed from the cluster. In particular, the Jump hash outperforms other consistent hashing algorithms by enforcing a restriction that the nodes are ordered, so nodes can only be added or removed if they are the last one/s.

However, a node in Pilosa can be removed or added even if it is in the middle. As an example, if we have node0, node1, node2, adding a node3 in Jump hash causes around 1/3 of data from the first three nodes to be placed in node3. However, removing node1 causes node2 to be the new "second node", so the data originally in node1 (before node3 was even added) will all go to node2, and all the data in previously in node2 goes to node3. This defeats the purpose of a consistent hash.

There are a few ways we can fix this. One way is to explicitly map the node order to a node. In the above example, we can map the value 2 to node2, so that when node1 is removed, we would send the node mapped to the highest value (node3) to the value 1. This might be too complicated to implement though. Moreover, we're currently seeing bad performance for the jump hash with a 130% ratio for max : average distribution of shards per node (for 1000-10000 shards on 10 nodes).

Another way is to use a different hashing algorithm altogether. Either use a new consistent hashing algorithm that still minimizes data transfer but does not depend on node order, or a regular hash that does not minimize data transfer on node add/leave events, but balances number of shards per node reasonably well for large number of shards and nodes.

shaqque:
## Overview

This PR refactors the Jump consistent hashing implementation to allow using other consistent hashing algorithms. Shard distribution to nodes now happen on a per-index basis, which means different indexes can be distributed using different algorithms.

All pre-existing indexes will default to the current Jump algorithm, while new indexes will be allocated using the default shard distribution algorithm set when the server starts/restarts.

Fixes #2035 

## Pull request checklist

## Code review checklist

### DIFF
--- a/api_test.go
+++ b/api_test.go
@@ -28,16 +28,17 @@ import (
 )
 
 func TestAPI_Import(t *testing.T) {
+	partitionN := 2
 	c := test.MustRunCluster(t, 2,
 		[]server.CommandOption{
 			server.OptCommandServerOptions(
 				pilosa.OptServerNodeID("node0"),
-				pilosa.OptServerClusterHasher(&offsetModHasher{}),
+				pilosa.OptServerShardDistributor(newOffsetModDistributor(partitionN)),
 			)},
 		[]server.CommandOption{
 			server.OptCommandServerOptions(
 				pilosa.OptServerNodeID("node1"),
-				pilosa.OptServerClusterHasher(&offsetModHasher{}),
+				pilosa.OptServerShardDistributor(newOffsetModDistributor(partitionN)),
 			)},
 	)
 	defer c.Close()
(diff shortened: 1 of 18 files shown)


shaqque:
@jaffee the tests are failing because the protobuf files are not yet generated. Unfortunately, I have the latest version of protoc/protobuf on my computer which adds trash fields like `XXX_blahblah`. Otherwise, all tests pass locally on my computer with the generated protobuf files and this PR should be ready.

An older version of protoc/protobuf should generate the protobuf files without the trash fields.

PR #2061, opened 2019-09-12
Initial extension architecture v0

seebs:
## Overview

What if we could have extensions?

That would be neat.

This slightly-sprawling pull request:

* Defines an extension API, and adds some support for it, allowing a hypothetical extension to add a new PQL query.
* Supports precomputing expressions on the coordinating node, and replacing the calls that generated them with references to the precomputed value. (This is necessary to make cross-index queries work.)
* Supports cross-index queries.
* Adds a more efficient Roaring format for transmitting rows to and from nodes, both in requests and responses.
* Adds a sample extension, `Some()`, which can return a pseudo-random selection of bits from a provided bitmap.
* Messes with, possibly improving, argument validation logic for PQL.
* Also, as a separate commit but it's in here anyway, adds some profiling logic to our existing opentracing stuff, so a query made with `?profile=true` will respond with a nested tree of span durations for operations performed while executing the query.

(The profiling thing could probably be split out easily enough, it just became a lot more interesting when I started having cross-shard experiments to run and performance questions about them.)


jaffee:
I'm gonna look at this, but based on our discussion this morning, I think we should cut a 1.4 with current master before merging this and a few other things for a soon-to-follow-but-probably-less-stable 1.5.

seebs:
Agreed. I might also have a notion for cleaning this up further, which needs more investigation, but might allow dropping most of the "many different op types" stuff.

PR #2074, opened 2019-10-11
use labeled targets for break statements

seebs:
break in a select in a for terminates the current case of the
select, but does not terminate the for loop. The worker queue
implementations for opening indexes/fields/views all suffered
from the same issue here.

Also fix a `<= 0` on a uint value.

All hail staticcheck.


codysoyland:
Could you also update `.circleci/config.yml` to use golangci-lint 1.20.0?

PR #2084, opened 2019-11-06
Add logic that clears fragment.rowCache after importValue

### LINKED ISSUE #2083: import to integer field seems to result in incorrect values
### What's going wrong?
Integer values imported with `pilosa import` seem to be getting set incorrectly when they come after a `Row()` query. I haven't dug into this, but I wanted to report it as it seems to be a problem. See the example below for what's happening.

### What was expected?
Separate imports result in the same values as a single import.

This works as expected:
```
curl localhost:10101/index/t -d '{"options": {"keys": false, "trackExistence": false}}'
curl localhost:10101/index/t/field/f -d '{"options": {"type": "int", "keys": false}}'
echo "2,1" > tmp.csv; pilosa import -i t -f f tmp.csv
echo "1000,1" > tmp.csv; pilosa import -i t -f f tmp.csv
curl localhost:10101/index/t/query -d 'Row(f>0)'
```
The result is:
```
{"results":[{"attrs":{},"columns":[2,1000]}]}
```

However, when a `Row` query is run between the two imports, the second import seems to get applied incorrectly.
```
curl localhost:10101/index/t2 -d '{"options": {"keys": false, "trackExistence": false}}'
curl localhost:10101/index/t2/field/f2 -d '{"options": {"type": "int", "keys": false}}'
echo "2,1" > tmp.csv; pilosa import -i t2 -f f2 tmp.csv
curl localhost:10101/index/t2/query -d 'Row(f2>0)'
echo "1000,1" > tmp.csv; pilosa import -i t2 -f f2 tmp.csv
curl localhost:10101/index/t2/query -d 'Row(f2>0)'
curl localhost:10101/index/t2/query -d 'Row(f2<0)'
```
The result here is:
```
{"results":[{"attrs":{},"columns":[2]}]}
{"results":[{"attrs":{},"columns":[1000]}]}
```

travisturner:
## Overview

It seems that `fragment.importValue` doesn't clear `fragment.rowCache` the way that other mutating methods do (see: `fragment.importRoaring`, `fragment.unprotectedSetBit`, etc.).

This PR is pretty heavy-handed in that it clears the rowCache for all rows of the BSI value, as well as the existence and sign rows. There's likely a more elegant way to handle this, but I'm putting this here to get get feedback.

TODO:
- [x] If we can settle on the appropriate solution for this, I will add tests at the `fragment` level (currently, the test that exercises this is in the `http` package).

Fixes #2083 

## Pull request checklist

## Code review checklist

### DIFF
--- a/field.go
+++ b/field.go
@@ -1292,34 +1292,12 @@ func (f *Field) importValue(columnIDs []uint64, values []int64, options *ImportO
 		return errors.Wrap(ErrBSIGroupNotFound, f.name)
 	}
 
-	// Find the lowest/highest values.
+	// We want to determine the required bit depth, in case the field doesn't
+	// have as many bits currently as would be needed to represent these values,
+	// but only if the values are in-range for the field.
 	var min, max int64
-	for i, value := range values {
-		if i == 0 || value < min {
-			min = value
-		}
-		if i == 0 || value > max {
-			max = value
-		}
-	}
-
-	// Determine the highest bit depth required by the min & max.
-	requiredDepth := bitDepthInt64(min - bsig.Base)
-	if v := bitDepthInt64(max - bsig.Base); v > requiredDepth {
-		requiredDepth = v
-	}
-
-	// Increase bit depth if required.
-	if requiredDepth > bsig.BitDepth {
-		if err := func() error {
-			f.mu.Lock()
-			defer f.mu.Unlock()
-			bsig.BitDepth = requiredDepth
-			f.options.BitDepth = requiredDepth
-			return f.saveMeta()
-		}(); err != nil {
-			return errors.Wrap(err, "increasing bsi bit depth")
-		}
+	if len(values) > 0 {
+		min, max = values[0], values[0]
 	}
 
 	// Split import data by fragment.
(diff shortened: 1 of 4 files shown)


seebs:
This is probably right but it shouldn't be... Examining more closely.

Hmm. Okay, this shouldn't be necessary, but I think I see what's happening. The rationale for not doing this, historically, is that we're going to snapshot anyway, and the snapshot remakes the cache.

Except I think right now, on snapshot, it looks like we *don't* remake the cache sometimes. Specifically, in the case where we're not unmarshalling the storage, just remapping the storage. But in this case, we should. This is fixed in the other branch I have reworking openStorage, but I think it'd be sufficient to move `f.rowCache = &simpleCache{make(map[uint64]*Row)}` from around line 414 of fragment.go (and also 367) up to pretty much the top of openStorage.

travisturner:
@seebs moving the rowCache reset code as you suggested doesn't work because `openStorage` isn't called between the first and second imports. I'll keep reviewing this to get a better understanding of what's happening.

travisturner:
I added a test to this at the fragment level. It's ready for review.

seebs:
Update: I found another more subtle problem, although I don't think it matters in any real cases. The computation in importValueSmallWrite of `rowSet` is used to do two things in importPositions; it's used to clear the (top N type) cache, and it's used to clear the rowCache, and the problem is that we're not clearing the top bit; we're going up to `bitDepth+1`, but we really need `bitDepth+bsiOffsetBit`.

But I don't think we ever use a non-None cache type for BSI fields, so the distinction doesn't matter.

But this also means that my analysis previously was wrong -- I was looking only at the large write case, but this one is failing in the small write case.

seebs:
Sorry to do this to you, but now that I understand the bug better, I rebased (and squashed) the two-part patch for the BSI depth in the other branch, could we use that version?

travisturner:
I added a test in #2088

PR #2171, opened 2022-09-27
fix: updating code to meet linting requirements

CLoZengineer:


client/batch_test.go
```diff
@@ -1662,8 +1663,7 @@ func mutexClearRegression(t *testing.T, c *test.Cluster, client *Client) {
 		t.Fatalf("getting batch: %v", err)
 	}
 
-	col := uint64(0)
-	row := uint64(1)
+	var col, row uint64
```
travisturner, line 1666: perhaps the value didn't matter, but it's worth checking that `row` doesn't need to be `1`

client/batch_test.go
```diff
@@ -1746,14 +1745,17 @@ func mutexNilClearID(t *testing.T, c *test.Cluster, client *Client) {
...
-	b.Add(
+	err = b.Add(
 		Row{
 			ID:     items[0],
 			Values: []interface{}{nil},
 			Clears: map[int]interface{}{0: nil},
 		},
 	)
-	b.Import()
+	assert.Nil(t, err)
```
travisturner, line 1755: assert has a `NoError()` method that we should probably use instead of `Nil()`
The difference would be that it takes a typed error. I think it's better to use that explicit method.

idk/ingest.go
```diff
@@ -1500,8 +1500,6 @@ func (m *Main) batchFromSchema(schema []Field) ([]Recordizer, pilosaclient.Recor
 					case DeleteSentinel:
 						if hasMutex { //need to clear the mutex
 							rec.Clears[valIdx] = nil //? maybe
-						} else { //TODO(twg) set fields not supported
```
travisturner, line 1503: @tgruben anything we need to do here before losing this comment?

rbf/cursor_test.go
```diff
@@ -1060,7 +1062,7 @@ func TestCursor_RemoveCells(t *testing.T) {
 	//f, err := os.OpenFile("before.dot", os.O_CREATE|os.O_TRUNC|os.O_WRONLY, 066)
 }
 
-//These aren't test i'm just using to generate graphs to look at structure
+// These aren't test i'm just using to generate graphs to look at structure
```
travisturner, line 1065: This comment makes me think perhaps this function shouldn't be part of our test suite.

travisturner (commented):
Thanks for doing this! I love when we clean up code!
I left a few comments.

authn/authenticate.go
```diff
@@ -25,6 +25,8 @@ import (
 	"golang.org/x/oauth2"
 )
 
+type AuthContextKey string
```
travisturner, line 28: (not sure where my comment went) Let's get in the habit of adding go-docs to any exported type/method/function that we add or modify in the code.

client/batch_test.go
CLoZengineer: in both cases row is immediately assigned a new value before being used

client/batch_test.go
```diff
@@ -1720,8 +1720,7 @@ func mutexNilClearID(t *testing.T, c *test.Cluster, client *Client) {
 		t.Fatalf("getting batch: %v", err)
 	}
 
-	col := uint64(0)
-	row := uint64(1)
+	var col, row uint64
```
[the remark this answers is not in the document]
CLoZengineer: in both cases row is immediately assigned a new value before being used

client/batch_test.go
CLoZengineer: Updated to NoError

rbf/cursor_test.go
CLoZengineer: I'd agree with you, but also want to keep the scope of this work to linting issues.
@kcrodgers24 is looking at our tests.
There's a lot to do, so want to focus it on completing different layers.

authn/authenticate.go
CLoZengineer: Updated

kcrodgers24 (approved):
Approved pending Todd's comments on that one idk/ingest.go TODO item.

PR #2174, opened 2022-10-11
chore: adds golangci config

CLoZengineer:


golangci.yml
```diff
@@ -0,0 +1,2310 @@
...
+    right-to-left-override: false
+    left-to-right-isolate: false
+    right-to-left-isolate: false
+    first-strong-isolate: false
+    pop-directional-isolate: false
+
+  cyclop:
+    # The maximal code complexity to report.
+    # Default: 10
+    max-complexity: 10
```
seebs, line 142: this seems insane. cyclomatic complexity isn't a great metric at the best of times, but *ten*? That's very solidly into "this is broken up into so many tiny pieces that reasoning about the tiny pieces becomes overwhelming" territory, I think.

golangci.yml
```diff
@@ -0,0 +1,2310 @@
...
+
+    # Section configuration to compare against.
+    # Section names are case-insensitive and may contain parameters in ().
+    # The default order of sections is `standard > default > custom > blank > dot`,
+    # If `custom-order` is `true`, it follows the order of `sections` option.
+    # Default: ["standard", "default"]
+    sections:
+      - standard                       # Standard section: captures all standard packages.
+      - default                        # Default section: contains all imports that could not be matched to another section type.
+      - prefix(github.com/org/project) # Custom section: groups all imports with the specified Prefix.
```
seebs, line 388: ...pretty sure "org/project" should be our org and project, not the placeholders.

golangci.yml
```diff
@@ -0,0 +1,2310 @@
...
+    # Skip generated files.
+    # Default: true
+    skip-generated: false
+
+    # Enable custom order of sections.
+    # If `true`, make the section order the same as the order of `sections`.
+    # Default: false
+    custom-order: true
+
+  gocognit:
```
seebs, line 401: do we benefit from having both this *and* the cyclomatic thing? my default preference would be "neither" really.

golangci.yml
```diff
@@ -0,0 +1,2310 @@
...
+      - HACK # marks hack-around that should be removed before merging
+
+  gofmt:
+    # Simplify code: gofmt with `-s` option.
+    # Default: true
+    simplify: false
+    # Apply the rewrite rules to the source before reformatting.
+    # https://pkg.go.dev/cmd/gofmt
+    # Default: []
+    rewrite-rules:
```
seebs, line 593: I don't think we should use rewrite rules in general, that just seems risky.

golangci.yml
```diff
@@ -0,0 +1,2310 @@
...
+    # database/sql is always checked
+    # Default: []
+    packages:
+      - github.com/jmoiron/sqlx
+
+  staticcheck:
+    # Select the Go version to target.
+    # Default: "1.13"
+    # Deprecated: use the global `run.go` instead.
+    go: "1.15"
```
seebs, line 1643: this isn't specific to this, but it seems to me like we have a lot of different places where we're specifying this, but also we specify different versions, but also there's this "deprecated: use the global run.go instead" warning?

seebs (changes requested):
I'm concerned about the sheer volume of new linting here and want some idea what this reports and how much of what it reports is noise. Also, I'd really like to have all the things that say "deprecated, use [...]" be not-used, because that just seems like asking for trouble.

CLoZengineer:
Agree with you @seebs. I just started with the reference and I've been pairing back.
I tend to find that with config files, if you start blank and add, you end up missing a lot of options. But when you pare them away, you're actively thinking of them before you do.

golangci.yml
CLoZengineer: I think this will come down to a case per case basis.
But I'd like to keep this at the default for now, and we will adjust it as the code quality is tightened up.

golangci.yml
```diff
@@ -0,0 +1,2310 @@
...
+  errcheck:
+    # Report about not checking of errors in type assertions: `a := b.(MyStruct)`.
+    # Such cases aren't reported by default.
+    # Default: false
+    check-type-assertions: true
+
+    # report about assignment of errors to blank identifier: `num, _ := strconv.Atoi(numStr)`.
+    # Such cases aren't reported by default.
+    # Default: false
+    check-blank: true
```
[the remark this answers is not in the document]
CLoZengineer: updated

golangci.yml
CLoZengineer: as we're starting this I'd rather maximize signal, and deal with noise later.
Though to your previous comments, I updated this to its default of 30"

golangci.yml
CLoZengineer: removed

golangci.yml
CLoZengineer: removed this, and other instances of run.go deprecations

seebs (approved):
This seems like a reasonable starting point to me. I'm assuming that this won't result in a situation where we can't merge any changes until we make hundreds of large and intrusive changes, though, if I'm wrong, don't merge it yet. :)

PR #2182, opened 2022-10-21
chore: adds release workflow with basic build step

CLoZengineer:


CLoZengineer:
Merging this without review as it needs to be in master to be able to be run for development.
Meta GitHub workaround.

PR #2185, opened 2022-10-25
feat: updates shardwidth.Exponent to single constant

### LINKED ISSUE #2184: Convert shardwidth to code constant
Currently the shardwidth exponent is set via build tags.
The effect could be accomplished via a constant in code

AC:

1. Exponent constant is set in code without the use of build tags
2. Go doc includes warning of data corruption if the size is changed
3. Go doc includes explanation of the values effect

CLoZengineer:
- removes build time constant in favor of single constant with relevant godoc and warnings
- removes code for build time config from makefile

closes #2184

### DIFF
--- a/shardwidth/helper.go
+++ b/shardwidth/helper.go
@@ -6,6 +6,13 @@ import (
 	"math/bits"
 )
 
+// Exponent controls the size of each shard
+//
+// # Warnings
+// - changing this value WILL corrupt any data sets created with a different value **
+// - both server and client must be compiled with the same Exponent **
+const Exponent = 20
+
 // FindNextShard returns the index of the first item which is not in the
 // same shard as i. The index it returns may be equal to the length of the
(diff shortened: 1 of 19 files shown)


shardwidth/helper.go
```diff
@@ -6,6 +6,13 @@ import (
 	"math/bits"
 )
 
+// Exponent controls the size of each shard
+//
+// # Warnings
+// - changing this value WILL corrupt any data sets created with a different value **
```
seebs, line 12: maybe `**changing this value...value**` so markdown will probably think it's bold?

CLoZengineer: This formatting should work everywhere, and I'm hoping with a `Warnings` header, we won't need to bold the actual warning items

seebs (approved):
some concerns about risk of bitrot but honestly i don't think the use case for which we had shardwidths before comes up much? we should merge this upstream too.

PR #2208, opened 2023-01-20
Add Drop Database and Drop Table support

travisturner:


dax/computer/service/computer.go
```diff
@@ -123,27 +119,21 @@ type CommandConfig struct {
...
-		ssSvc = snapshotter.New(snapshotter.Config{
-			DataDir: cfg.SnapshotterConfig.DataDir,
-			Logger:  cfg.Logger,
-		})
+	var ssSvc computer.SnapshotService
+	if cfg.ComputerConfig.SnapshotterDir != "" {
+		ssSvc = snapshotter.New(cfg.ComputerConfig.SnapshotterDir, cfg.Logger)
+	} else {
+		ssSvc = computer.NewNopSnapshotterService()
+		cfg.Logger.Warnf("No snapshotter configured.")
```
jaffee, line 136: if you want a no-op you should have to explicitly ask for it so we don't have folks accidentally running w/o snapshot/writelog and not knowing it.

...and by folks I mean us

dax/test/dax/dax_test.go
```diff
@@ -507,6 +514,134 @@ func TestDAXIntegration(t *testing.T) {
...
+			numSlash := strings.Count(name, "/")
+			for i := 0; i < numSlash; i++ {
+				fmt.Print(" ")
+			}
+			fmt.Printf("%s\n", name)
+			return nil
+		})
+		if err != nil {
+			t.Fatalf("walking dir: %v", err)
+		}
```
jaffee, line 643: finish this test... not actually checking anything. also need to clean up the duplicated walk func

jaffee (changes requested):
the requested changes are in the comments

dax/computer/service/computer.go
travisturner: I changed this to accept `"NULL"` or error on `""`

dax/test/dax/dax_test.go
travisturner: just added some checks to see if the directories are empty or not

dax/computer/service/computer.go
travisturner: changed it to case-insensitive "off"

jaffee (approved):

PR #2210, opened 2023-01-20
get fb_exec_requests underlying storage to use less memory

paddyjok:
### Overview

This PR caps the number of sql requests stored in memory and caps the amount of memory used to store sql test and plan text.

### DIFF
--- a/systemlayer/executionrequests.go
+++ b/systemlayer/executionrequests.go
@@ -9,16 +9,21 @@ import (
 )
 
 const (
-	keepLastRequests int = 2000
-	truncateTextAt   int = 4096
+	keepLastRequests = 2000
+	truncateTextAt   = 4096
 )
 
 // ExecutionRequests is an internal struct that keeps a list of sql execution requests
 // this data allows visbility into queries that have been run and are running
 type ExecutionRequests struct {
 	sync.RWMutex
+	// requestsList being used a FIFO queue here
+	// to track both number of stored requests and
+	// which one to delete next (always 0)
 	requestsList []string
-	requests     map[string]*pilosa.ExecutionRequest
+	// the actual rquests being stored - we use a map
+	// so we can look them up by request id which is a uuid
+	requests map[string]*pilosa.ExecutionRequest
 }
 
 // Ensure type implements interface.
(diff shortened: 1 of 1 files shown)


systemlayer/executionrequests.go
```diff
@@ -8,32 +8,58 @@ import (
 	pilosa "github.com/featurebasedb/featurebase/v3"
 )
 
+const (
+	keepLastRequests int = 2000
+	truncateTextAt   int = 4096
```
jaffee, line 13: these can be untyped constants which will give us a bit more flexibility. (just remove "int") https://go.dev/blog/constants

systemlayer/executionrequests.go
```diff
@@ -8,32 +8,58 @@ import (
...
+	requests     map[string]*pilosa.ExecutionRequest
 }
 
 // Ensure type implements interface.
 var _ pilosa.ExecutionRequestsAPI = (*ExecutionRequests)(nil)
 
 func NewExecutionRequestsAPI() *ExecutionRequests {
 	return &ExecutionRequests{
-		requests: make(map[string]*pilosa.ExecutionRequest),
+		requestsList: make([]string, 0),
```
jaffee, line 29: This will allocate the requestsList once up front, rather than reallocating as it grows up to its max size. Not a big deal either way, but also makes it explicit that this is the size we expect it to be.
```suggestion
		requestsList: make([]string, 0, keepLastRequests),
```

jaffee: actually nevermind... because of the way we're appending below we're going to keep needing to reallocate. We'd have to implement a true ring buffer tracking the current index to avoid that.

systemlayer/executionrequests.go
```diff
@@ -8,32 +8,58 @@ import (
...
+	// we're going to add a new request so clear out oldest request if we've hit
+	// the threshhold
+	if len(e.requestsList) >= keepLastRequests {
+		// get the oldest request
+		delId := e.requestsList[0]
+
+		// make sure it's not this request, unlikely, but
+		// we've all been bitten by 'this should never happen' before
+		// and if it isn't delete it from the map and the list
+		if delId != requestID {
```
jaffee, line 48: why do we care if the oldest request has the same ID as this request? Won't we just add this request to the map and list again below?

systemlayer/executionrequests.go
```diff
@@ -8,32 +8,58 @@ import (
...
+			delete(e.requests, delId)
+			e.requestsList = e.requestsList[1:]
+		}
+	}
+	// add the request to the end
+	e.requestsList = append(e.requestsList, requestID)
+	//update the request
```
jaffee, line 55:
```suggestion
	// update the request
```

systemlayer/executionrequests.go
paddyjok: Good point - was being overly cautious...

systemlayer/executionrequests.go
jaffee: TODO: pretty sure we can get the linter to enforce this.  I've made a note

systemlayer/executionrequests.go
jaffee: I think the current code might actually be worse as if all the request IDs were the same somehow the list grows indefinitely again.

systemlayer/executionrequests.go
jaffee: yeah agree

jaffee (dismissed):

jaffee (dismissed):

jaffee (approved):

PR #2213, opened 2023-01-21
WIP: json marshal coded errors for http

travisturner:
@jaffee @paddyjok I took a stab at how we might json-encode/decode a `codedError` so it can cross an http boundary without losing its code/message.

This commit also shows a single example of the http handler and its corresponding client modified to use this. This results in the client method returning a `codedError` which can be compared with the `errors.Is(someCode)` method.

Looking for feedback.

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -3,6 +3,8 @@
 package errors
 
 import (
+	"encoding/json"
+
 	"github.com/pkg/errors"
 )
 
@@ -12,8 +14,8 @@ type Code string
 
 func New(code Code, message string) error {
 	return errors.WithStack(codedError{
-		code:    code,
-		message: message,
+		Code:    code,
+		Message: message,
 	})
 }
 
@@ -33,7 +35,7 @@ func Errorf(format string, args ...interface{}) error {
 // an error Code instead of an error.
 func Is(err error, target Code) bool {
 	match := codedError{
-		code: target,
+		Code: target,
 	}
 	return errors.Is(err, match)
 }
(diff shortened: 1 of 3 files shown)


dax/mds/client/client.go
```diff
@@ -159,7 +159,7 @@ func (c *Client) DatabaseByName(ctx context.Context, orgID dax.OrganizationID, n
 
 	if resp.StatusCode != http.StatusOK {
 		b, _ := io.ReadAll(resp.Body)
-		return nil, errors.Errorf("status code: %d: %s", resp.StatusCode, b)
+		return nil, errors.Wrapf(errors.UnmarshalJSON(b), "status code: %d", resp.StatusCode)
```
jaffee, line 162: why don't we have errors.UnmarshalJSON take a reader and wrap the ReadAll up in there?

jaffee (approved):
this seems pretty reasonable to me

dax/mds/client/client.go
travisturner: that's a good idea. I'll make that change and then convert the rest of the methods in our service apis to use this.

errors/errors.go
```diff
@@ -87,3 +90,50 @@ func (ce codedError) Is(err error) bool {
...
 	ErrUncoded Code = "Uncoded"
 )
+
+// MarshalJSON returns the provided error as a json object (as a string)
+// representing a codedError. If err is not already a codedError, the json
+// object will still represent a codedError but its `code` value will be empty.
+// Note: an empty code here is intentional and is different from code
+// `errors.Uncoded` which is a valid code; it just means the developer returned
+// a codedError but didn't bother to choose (or create) a useful error code.
+func MarshalJSON(err error) string {
```
paddyjok, line 100: 👍

paddyjok (commented):
This looks reasonable. How stringent are we planning to be on this? Given we're predicating behavioral flow on errors in some cases, should not having an proper coded error be a 'develop time error' type situation?

errors/errors.go
travisturner: This is a good question. I have a few thoughts on how I imagined this evolving:

- add a test which verifies that the methods implemented by a service's http handler and corresponding client include this encoding (the goal is to prevent a future developer from adding a method which does not container the encoding)
- start logging (as a warning) any time this encoding path receives a non `codedError` so we can see what's getting missed
- over time, try to address those things being logged with the goal of eventually converting that log to a developer error (as you mentioned)
- go through the code and replace existing `UncodedError`s with an actual code.

PR #2214, opened 2023-01-21
Fix CLI "exit" command

travisturner:
When we added a line feed to the user input, we broke the check for "exit" (because after that change we were getting "exit\n". This commit moves the exit check before the line feed append.


jaffee (approved):
yum

PR #2216, opened 2023-01-23
Fb-1816 : CharIndex str func

HHans09:
SQL string function CharIndex to identify the position of the substring in the input string. Position starts at 0

### DIFF
--- a/sql3/planner/inbuiltfunctionsstring.go
+++ b/sql3/planner/inbuiltfunctionsstring.go
@@ -214,7 +214,6 @@ func (p *ExecutionPlanner) analyseFunctionFormat(call *parser.Call, scope parser
 	if len(call.Args) < 1 {
 		return nil, sql3.NewErrCallParameterCountMismatch(call.Rparen.Line, call.Rparen.Column, call.Name.Name, 1, len(call.Args))
 	}
-
 	if !typeIsString(call.Args[0].DataType()) && !typeIsVoid(call.Args[0].DataType()) {
 		return nil, sql3.NewErrStringExpressionExpected(call.Args[0].Pos().Line, call.Args[0].Pos().Column)
 	}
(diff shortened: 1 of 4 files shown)


sql3/planner/inbuiltfunctionsstring.go
```diff
@@ -644,3 +668,65 @@ func (n *callPlanExpression) EvaluateReplicate(currentRow []interface{}) (interf
 	return strings.Repeat(stringArg, int(intArg)), nil
 
 }
+
+// function to get the index of substring in the given string
```
rachithrr, line 672: The exported functions should follow this format
``` EvalutateCharIndex ... [and then describe what it does and any information that would be useful for a user or another developer to know] ```
More info about that here: https://go.dev/blog/godoc

File: sql3/planner/inbuiltfunctionsstring.go [hunk 2]
```diff
@@ -644,3 +668,65 @@ func (n *callPlanExpression) EvaluateReplicate(currentRow []interface{}) (interf
...
+	argEval, err := n.args[0].Evaluate(currentRow)
+	if err != nil {
+		return nil, err
+	}
+	if argEval == nil {
+		return nil, nil
+	}
+	subString, ok := argEval.(string)
+	if !ok {
+		return nil, sql3.NewErrInternalf("unexpected type converion %T", argEval)
```
rachithrr, line 684: fb-1817 adds a new function that wraps around ErrInternalF. Rebase once that is merged.

sql3/planner/inbuiltfunctionsstring.go
```diff
@@ -644,3 +668,65 @@ func (n *callPlanExpression) EvaluateReplicate(currentRow []interface{}) (interf
...
+		return nil, nil
+	}
+	inputString, ok := argEval.(string)
+	if !ok {
+		return nil, sql3.NewErrInternalf("unexpected type converion %T", argEval)
+	}
+
+	if len(n.args) == 2 {
+		res := strings.Index(inputString, subString)
+		if res > -1 {
```
rachithrr, line 701: do we need this if condition? 
// return int64(res), nil

sql3/test/defs/defs_string_functions.go
```diff
@@ -820,5 +820,87 @@ var stringScalarFunctionsTests = TableTest{
...
+				"select charindex('is','this is great','you')",
+			),
+			ExpErr: "integer expression expected",
+		},
+		{
+			name: "CharIndexIncorrectTypeofArgumentsforInputString",
+			SQLs: sqls(
+				"select charindex('is',23,3)",
+			),
+			ExpErr: "string expression expected",
```
rachithrr, line 843: add compareexactordered

sql3/planner/inbuiltfunctionsstring.go
HHans09: Updated

sql3/planner/inbuiltfunctionsstring.go
HHans09: yes, when position is not found then it has to return -1

sql3/test/defs/defs_string_functions.go
HHans09: Do we need to add compareexactordered when we are expecting errors?

travisturner: no, you don't need it for errors

sql3/planner/inbuiltfunctionsstring.go
rachithrr: The strings.Index() returns -1 if not found, which is what we are returning in line 705.

sql3/planner/inbuiltfunctionsstring.go
HHans09: Rebased and updated

rachithrr (approved):
looks good to me

PR #2219, opened 2023-01-24
Remove MDS and replace it with Controller

travisturner:
This commit removes the MDS layer (and package) and shifts Controller package into its place.

I'll need to make the relevant updates to the cloud repos before we merge this, but I thought I'd go ahead and put this here for review. It's 99% renaming and moving packages. The other 1% is some logic change in the Controller: I moved some of the interface implementation injection from the top-level `controller` package to the `controller/service` package (which is really where it's supposed to be since that is the equivalent of the `server/server` package).


jaffee (dismissed):
bah bah bah bah.... I'm lovin' it

jaffee (approved):

PR #2220, opened 2023-01-24
FB-1817: Implement FORMAT()

rachithrr:


sql3/planner/inbuiltfunctionsstring.go
```diff
@@ -644,3 +660,34 @@ func (n *callPlanExpression) EvaluateReplicate(currentRow []interface{}) (interf
 	return strings.Repeat(stringArg, int(intArg)), nil
 
 }
+
+// implements fmt.Sprintf
```
travisturner, line 664: Just for reference, exported methods (and really any function, type, or method which is complex enough to benefit from having a description) should have godocs formatted like:
```
// EvalutateFormat ... [and then describe what it does and any information that would be useful for a user or another developer to know]
```
More info about that here: https://go.dev/blog/godoc

travisturner (commented):
Hey @rachithrr 
I like that you took the initiative to replace the internal error with a more explicit error. Thanks!

I just left a comment about godoc formatting.

sql3/test/defs/defs_string_functions.go
```diff
@@ -820,5 +820,86 @@ var stringScalarFunctionsTests = TableTest{
...
+			name: "FormatBoolean",
+			SQLs: sqls(
+				"select format('is this %t?', true)",
+			),
+			ExpHdrs: hdrs(
+				hdr("", fldTypeString),
+			),
+			ExpRows: rows(
+				row(string("is this true?")),
+			),
```
travisturner, line 846: some of these tests are missing the line:
```
Compare: CompareExactOrdered,
```
without a `Compare` type specified for non-error tests, the tests aren't getting exercised.

sql3/planner/inbuiltfunctionsstring.go
```diff
@@ -644,3 +660,34 @@ func (n *callPlanExpression) EvaluateReplicate(currentRow []interface{}) (interf
...
+
+	// loop, since args can be of any length.
+	for _, arg := range n.args[1:] {
+		argEval, err = arg.Evaluate(currentRow)
+		if err != nil {
+			return nil, err
+		}
+		if argEval == nil {
+			return nil, nil
+		}
```
travisturner, line 689: I would argue that this `nil` check should be removed completely.
The `FormatNullArg` test related to this does not do what I would expect; in other words I would expect a `nil` argument to cause the entire string to be returned as `nil`.
If we following `fmt.Sprintf()` logic, that test should return `"this = %!s(<nil>)"`.

travisturner (changes requested):

sql3/planner/inbuiltfunctionsstring.go
travisturner: Alternatively, if we don't want to support `nil` arguments, then that nil check could return an error.

rachithrr: The idea here is that, like all other functions, If any arg is null, we return NULL. But, I see that this might not be the case for Format() function. Will change to return error then.

travisturner (commented):
This looks good to me.
Can you add a test case covering the NullNotAllowed error case?
Other than that, I think it's ready.

travisturner (approved):

PR #2222, opened 2023-01-25
Fix pre sort cmd

jaffee:
this adds parallel processing support to the ndjson pre sorter and fixes several bugs


ctl/pre_sort.go
```diff
@@ -23,27 +26,37 @@ type PreSortCommand struct {
 
 	Table string `json:"table"`
 
-	PrimaryKeyFields []string `json:"primary-key-fields"`
+	PrimaryKeyFields    []string `json:"primary-key-fields"`
+	PrimaryKeySeparator string
```
travisturner, line 30: does this need a json tag?

jaffee: actually I think all these tags are completely spurious... I put them in out of habit because commandeer will read them... but we don't use commandeer for FB subcommands

jaffee: removed

travisturner (dismissed):
This seems to work for me as expected in my rather minimal testing.

I would vote on calling this subcommand `presort` over `pre_sort` but that's just a preference.

Also, note that other sub-commands use hyphens, so at the very least `pre-sort`

jaffee:
name changed

travisturner (approved):

PR #2224, opened 2023-01-25
timestamp fixes

paddyjok:
## Overview

This PR fixes:

1. string representations of timestamps in CSV files now ingest the timestamp, and not 0 as it was previously
2. min and max on timestamps (where these are pushed down into PQL min/max operators

The requisite testing has been added.

### DIFF
--- a/sql3/planner/oppqlaggregate.go
+++ b/sql3/planner/oppqlaggregate.go
@@ -263,20 +263,33 @@ func (i *pqlAggregateRowIter) Next(ctx context.Context) (types.Row, error) {
 			i.resultValue = int64(actualResult)
 
 		case pilosa.ValCount:
-			if actualResult.DecimalVal == nil {
+			switch i.aggregate.Type().(type) {
+			case *parser.DataTypeInt:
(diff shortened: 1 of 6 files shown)


sql3/planner/oppqlaggregate.go
```diff
@@ -263,20 +263,33 @@ func (i *pqlAggregateRowIter) Next(ctx context.Context) (types.Row, error) {
...
 				}
-			} else {
+
+			case *parser.DataTypeDecimal:
 				if i.aggregate.AggType() == types.AGGREGATE_AVG {
-					average := actualResult.DecimalVal.Float64() / float64(actualResult.Count)
-					i.resultValue = pql.NewDecimal(int64(average*10000), 4)
+					if actualResult.DecimalVal == nil {
+						average := float64(actualResult.Val) / float64(actualResult.Count)
+						i.resultValue = pql.NewDecimal(int64(average*10000), 4)
```
tgruben, line 279: Should the scale be obtained from the schema?

paddyjok: Yeah - good point.

tgruben (approved):

PR #2225, opened 2023-01-25
try to make sure we're releasing large allocations

jaffee:
this seems to work to stop all those persistent allocations we were seeing. There are two main things here:

1. When truncating SQL and SQL plan strings for storage, copy them to new fixed size memory allocations so we don't retain the entire underlying allocation.
2. make the requestsList a true ring buffer rather than tacking onto the end and reslicing off the beginning... I believe this was also retaining the underly memory allocation


j3m7 (approved):

paddyjok (approved):
lgtm

PR #2226, opened 2023-01-25
bulk import support for parquet files

tgruben:


sql3/planner/compilebulkinsert.go
```diff
@@ -203,6 +203,14 @@ func (p *ExecutionPlanner) analyzeBulkInsertStatement(stmt *parser.BulkInsertSta
 				return sql3.NewErrIntegerLiteral(im.MapExpr.Pos().Line, im.MapExpr.Pos().Column)
 			}
 		}
+	case "PARQUET":
+		// for paquet the map expressions need to be string values
```
travisturner, line 207:
```suggestion
		// for parquet the map expressions need to be string values
```

sql3/planner/compilebulkinsert.go
```diff
@@ -203,6 +203,14 @@ func (p *ExecutionPlanner) analyzeBulkInsertStatement(stmt *parser.BulkInsertSta
...
+		// that represent the offsets in the source file
+		for _, im := range stmt.MapList {
+			if !(im.MapExpr.IsLiteral() && typeIsString(im.MapExpr.DataType())) {
+				return sql3.NewErrIntegerLiteral(im.MapExpr.Pos().Line, im.MapExpr.Pos().Column)
```
travisturner, line 211: I don't quite understand the "must be string" logic, but does that mean this error should be `NewErrStringLiteral`?

sql3/sql_complex_test.go
```diff
@@ -4,13 +4,21 @@ package sql3_test
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
```
travisturner, line 7: you probably got this warning, but this package is deprecated.

sql3/sql_complex_test.go
```diff
@@ -2631,3 +2638,216 @@ func TestPlanner_BulkInsert_FB1831(t *testing.T) {
...
+	return nil
+}
+
+func TestPlanner_BulkInsertParquet_File(t *testing.T) {
+	// check that can pull parquet file from URL and load
+	c := test.MustRunCluster(t, 1)
+	defer c.Close()
+
+	_, _, err := sql_test.MustQueryRows(t, c.GetNode(0).Server, `create table j1 (_id ID, a INT, b DECIMAL(2), c STRING);`)
+	if err != nil {
```
travisturner, line 2695: I would prefer we use testify `assert.NoError(t, err)` (along with any other testify methods) in tests going forward.

sql3/sql_complex_test.go
```diff
@@ -2631,3 +2638,216 @@ func TestPlanner_BulkInsert_FB1831(t *testing.T) {
...
 	}
 }
+
+type tb struct {
+	Name  string
+	Type  arrow.DataType
+	Value interface{}
+}
+
+func simpleParquetMaker(f *os.File, numRows int64, input []tb) error {
```
travisturner, line 2648: we could make this a `t.Helper()`, use `assert.NoError(t, err)` within, and don't bother returning an error on this helper function

travisturner (commented):
This looks pretty slick.
I know you didn't ask for my review, but I wanted to look through it, and ended up leaving some comments.

One other thing I wondered is if we could share the same `MustRunCluster` for all three of these tests to avoid adding more startups to the tests. Is there a risk with that?

sql3/planner/compilebulkinsert.go
tgruben: good ole copy paste get's me every time

sql3/sql_complex_test.go
tgruben: forgot all about testify, will update

travisturner:
I think I might have misunderstood how `MustRunCluster()` works. Is it already shared? You may have had it configured correctly to begin with. Sorry.

tgruben:
thanks @travisturner , your suggestions made it look a lot cleaner

travisturner (approved):
I just imported my first parquet file!

The only thing I would add, and it's very minor, is that you could use the `TempFile` function in the `featurebase/testhook` package to create the temp files, and then you don't have to explicitly remove them.

But approved either way.

PR #2229, opened 2023-01-26
FB-1905: Test Consumer Close Timeout

jrbrinlee1:
This is the testing for changes made for SUP-288. It configures the kafka consumer in such a way that it hangs after exceeding max.poll.interval.ms. The test confirms that we don't hang indefinitely which was the requirement.

### DIFF
--- a/idk/kafka/cmd_test.go
+++ b/idk/kafka/cmd_test.go
@@ -179,13 +179,32 @@ func TestConfigOptions(t *testing.T) {
 		t.Fatalf("getting main: %v", err)
 	}
 
-	m.KafkaDebug = "consumer"
-	m.KafkaClientId = "blah"
-	m.SkipOld = true
-	m.KafkaGroupInstanceId = "instId"
-	m.KafkaMaxPollInterval = "30"
-	m.KafkaSessionTimeout = "20"
-	m.KafkaSocketKeepaliveEnable = "true"
+	debug := "consumer"
+	clientID := "blah"
+	autoOffsetReset := true
+	groupInstanceID := "instId"
+	maxPollIntervalMs := "30"
+	sessionTimeoutMs := "20"
+	socketKeepaliveEnable := "true"
+
+	m.KafkaDebug = debug
+	m.KafkaClientId = clientID
+	m.SkipOld = autoOffsetReset
+	m.KafkaGroupInstanceId = groupInstanceID
+	m.KafkaMaxPollInterval = maxPollIntervalMs
+	m.KafkaSessionTimeout = sessionTimeoutMs
+	m.KafkaSocketKeepaliveEnable = socketKeepaliveEnable
+
+	//kafka config name, expected value
+	tests := [][2]interface{}{
+		{"debug", debug},
+		{"client.id", clientID},
+		{"auto.offset.reset", "latest"},
+		{"group.instance.id", groupInstanceID},
+		{"max.poll.interval.ms", maxPollIntervalMs},
+		{"session.timeout.ms", sessionTimeoutMs},
+		{"socket.keepalive.enable", socketKeepaliveEnable},
+	}
 
 	// call NewSource to get and open source which calls some setup
 	// stuff that we want. We don't expect it to open successfully
@@ -198,26 +217,10 @@ func TestConfigOptions(t *testing.T) {
 
 	cfg := src.(*Source).ConfigMap
 
-	if val, err := cfg.Get("debug", nil); err != nil || val.(string) != "consumer" {
-		t.Fatalf("unexpected val for debug val: %v, err: %v", val, err)
-	}
-	if val, err := cfg.Get("client.id", nil); err != nil || val.(string) != "blah" {
-		t.Fatalf("unexpected val for client.id val: %v, err: %v", val, err)
-	}
-	if val, err := cfg.Get("auto.offset.reset", nil); err != nil || val.(string) != "latest" {
-		t.Fatalf("unexpected val for auto.offset.reset val: %v, err: %v", val, err)
-	}
-	if val, err := cfg.Get("group.instance.id", nil); err != nil || val.(string) != "isntId" {
-		t.Fatalf("unexpected group.instance.id val: %v, err: %v", val, err)
-	}
-	if val, err := cfg.Get("max.poll.interval.ms", nil); err != nil || val.(string) != "30" {
-		t.Fatalf("unexpected val for max.poll.interval.ms val: %v, err: %v", val, err)
-	}
-	if val, err := cfg.Get("session.timeout.ms", nil); err != nil || val.(string) != "20" {
-		t.Fatalf("unexpected val for session.timeout.ms val: %v, err: %v", val, err)
-	}
-	if val, err := cfg.Get("socket.keepalive.enable", nil); err != nil || val.(string) != "true" {
-		t.Fatalf("unexpected val for socket.keepalive.enable val: %v, err: %v", val, err)
+	for _, pair := range tests {
+		if val, err := cfg.Get(pair[0].(string), nil); err != nil || val.(string) != fmt.Sprintf("%v", pair[1]) {
+			t.Fatalf("unexpected val for %s val: %v, err: %v", pair[0].(string), val, err)
+		}
 	}
 }
 
--- a/idk/kafka/cmd_test.go
+++ b/idk/kafka/cmd_test.go
@@ -1443,9 +1443,9 @@ This test:
     a. the consumer did leave the group due to exceeding max.poll.interval.ms
     b. we close the consumer and log that we weren't able to confirm it closed properly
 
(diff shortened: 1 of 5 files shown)


idk/kafka/cmd_test.go
```diff
@@ -682,3 +699,107 @@ func tDoHTTPPost(t *testing.T, url, contentType, body string) string {
...
+	}
+
+	consumerError := make(chan error)
+	go func() {
+		consumerErr := consumer.Run()
+		consumerError <- consumerErr
+	}()
+
+	//wait some time for ingest - takes ~5 seconds locally
+	time.Sleep(5 * time.Second)
```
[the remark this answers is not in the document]
jrbrinlee1: Thanks for taking a look at this and good point. I'm writing the consumer logs to disk so I could read the log file until I see a message that indicates some number of messages has been consumer. As I can't think of any other options, I'll make that change.

jrbrinlee1:
This is ready for review for anyone that is going through and looking PRs.

idk/kafka/cmd_test.go
```diff
@@ -682,3 +702,122 @@ func tDoHTTPPost(t *testing.T, url, contentType, body string) string {
...
+	start := time.Now()
+
+	for !complete {
+		buf, _ := os.ReadFile(log_path)
+		s := string(buf)
+		if strings.Contains(s, "records processed 0-> (5000)") {
+			complete = true
+		}
+		elapsed := time.Since(start)
+		if elapsed > 15*1000*1000*1000 { //15 seconds
```
seebs, line 779: `15 * time.Second`

idk/kafka/cmd_test.go
```diff
@@ -682,3 +702,122 @@ func tDoHTTPPost(t *testing.T, url, contentType, body string) string {
...
+we also see the consumer hang when it's trying to close properly. See SUP-288.
+This test:
+ 1. configures the consumer such that it will hang if max.poll.interval.ms is exceeded.
+ 2. checks that
+    a. the consumer did leave the group due to exceeding max.poll.interval.ms
+    b. we close the consumer and log that we weren't able to confirm it closed properly
+
+It could be possible this test fails even though there isn't necessarily an issue. This
+would be the case if a fix comes along to the kafka-go lib that prevents the consumer
+from hanging during the call it's consumer.Close() func.
```
seebs, line 717: its ("it's" is the contraction of "it is", "its" is like "his" or "her/hers")

idk/kafka/cmd_test.go
```diff
@@ -682,3 +702,122 @@ func tDoHTTPPost(t *testing.T, url, contentType, body string) string {
 
 	return string(bod)
 }
+
+/*
+In some cases, when we see max.poll.interval.ms exceeded on the kafka consumer,
+we also see the consumer hang when it's trying to close properly. See SUP-288.
+This test:
+ 1. configures the consumer such that it will hang if max.poll.interval.ms is exceeded.
```
seebs, line 710: I would like to see a bit more of an explanation on this -- what about our configuration would cause it to hang? Why would it hang? I don't know enough about the consumer to recognize how we did this.

idk/kafka/cmd_test.go
```diff
@@ -682,3 +702,122 @@ func tDoHTTPPost(t *testing.T, url, contentType, body string) string {
...
+	}()
+
+	// wait until all 5000 messages to be consumed
+	complete := false
+	start := time.Now()
+
+	for !complete {
+		buf, _ := os.ReadFile(log_path)
+		s := string(buf)
+		if strings.Contains(s, "records processed 0-> (5000)") {
```
seebs, line 775: So, what happens if someone decides this is taking too long and edits `genNumRecords` to be 2000? This presumably never happens. Can we make this contingent on that behavior? Or is there some other way we can have this actually directly relate to the consumer's behavior, like having the consumer close a channel when it's done?

idk/kafka/cmd_test.go
```diff
@@ -682,3 +702,122 @@ func tDoHTTPPost(t *testing.T, url, contentType, body string) string {
...
+			complete = true
+		}
+		elapsed := time.Since(start)
+		if elapsed > 15*1000*1000*1000 { //15 seconds
+			t.Fatalf("unable to read from consumer log file")
+		}
+	}
+
+	// take a transaction lock on the database which drives
+	// max.poll.interval.ms timeout to occur
```
seebs, line 785: Why does it do that? I guess because it blocks writes, and thus the consumer can't do writing?

idk/kafka/cmd_test.go
```diff
@@ -682,3 +702,122 @@ func tDoHTTPPost(t *testing.T, url, contentType, body string) string {
...
+
+	// wait for the consumer to exit or 30 seconds to pass
+	select {
+	case <-consumerError:
+		buf, err := os.ReadFile(log_path)
+		if err != nil {
+			t.Errorf("issue reading consumer log: %s", err)
+		}
+		s := string(buf)
+		if !strings.Contains(s, "Application maximum poll interval") && !strings.Contains(s, "Unable to properly close consumer") {
```
seebs, line 817: is the intent here that it must contain both, or that it must contain either? I believe this will only report a test-failure if both strings are missing, but I'd have thought we wanted to fail if either was missing.

idk/kafka/cmd_test.go
```diff
@@ -682,3 +702,122 @@ func tDoHTTPPost(t *testing.T, url, contentType, body string) string {
...
+	if err != nil {
+		t.Fatalf("issues creating a consumer")
+	}
+	configureTestFlags(consumer)
+	consumer.Topics = []string{topic}
+	consumer.BatchSize = 1000
+	consumer.IDField = "user_id"
+	consumer.KafkaGroupInstanceId = "some_id"
+	consumer.ConsumerCloseTimeout = 3
+	consumer.KafkaMaxPollInterval = "15000"
```
seebs, line 742: I assume this is ... milliseconds as a string? That is certainly a novel data type. :P

I would be inclined to pick a significantly shorter timeout in testing because in theory none of this should take more than a second or two, right?

seebs (changes requested):
This looks basically reasonable but I am gonna be nitpicky on some of the details. The only thing I'm actually really solid on is that the multiplication should be `time.Second` since that's what the constant exists for, but I'd like to understand the answers to the other questions.

idk/kafka/cmd_test.go
jrbrinlee1: So this is a somewhat unanswered question. The consumer hangs in the librdkafka library that this https://github.com/confluentinc/confluent-kafka-go depends on. Which of course we depend on. A decision was made to accept that the consumer could hang and that we would just close it unsuccessfully if it did. So to answer this, we'd need to explore librdkafka as far as I know.

idk/kafka/cmd_test.go
jrbrinlee1: Yep. It blocks write, so the consumer waits to be able to write. During that time, it doesn't poll for more messages. And exceeds this interval.

idk/kafka/cmd_test.go
jrbrinlee1: So.. we need the consumer to stay alive so that it can, later in the test, timeout because it doesn't poll for messages. But I can format the string to include genNumRecords and I think that is a good call out!

idk/kafka/cmd_test.go
jrbrinlee1: So, I remember playing around with all the timeouts & what not. Dug into it a bit more and here is what I found.

- Session timeouts > 6 seconds give the following error:
```
%7|1678230252.773|REJOIN|rdkafka#consumer-1| [thrd:main]: Group "defaultgroup": Rejoining group without an assignment: JoinGroup error: Broker: Invalid session timeout
```
- There could be a few reasons for this but I haven't dug into it much further.
- max poll interval needs to be > that session timeout so it's at 6.5 seconds.
- the transaction lock needs to be a couple seconds > that the poll interval or else we don't get the timeout we're after so it's set at 9 seconds.
- We have to wait for the transaction lock to end before the close timeout clock starts ticking. Once it doesn't we have to wait for that close timeout to timeout.
- I'm hesitant dropping the close timeout in the case it actually takes some time to close property. Then we could get a false positive (the consumer was going to close properly but we didn't give it enough time.
- So then we have to wait ~15 seconds to know if the consumer is actually hanging (9 sec transaction lock + 3 second close timeout)

So anyways, let me know what you think. Is it worth digging into the session timeout error when < 6 seconds?

seebs: Hmm. I think... probably not right now? That feels like a thing that could be any of several other timeout issues compounding, and it feels like we don't need to explore it right now. Maybe make a note or a comment somewhere so if someone comes looking they get that context.

seebs (approved):
This looks reasonable to me.

PR #2231, opened 2023-01-26
Support for setting individual DatabaseOptions

travisturner:
This PR now contains the SQL support for `DATABASE` related operations:
```
CREATE DATABASE db1 WITH UNITS 2 COMMENT 'my database';
ALTER DATABASE db1 WITH units = 3;
SHOW DATABASES;
DROP DATABASE db1;
```
It currently works for:
- [ ] on-prem, classic
- [x] on-prem, serverless
- [ ] cloud, classic
- [ ] cloud, serverless

I'll focus on getting it to work for "cloud, serverless" next.


api_directive.go
```diff
@@ -324,6 +324,20 @@ func (api *API) pushJobsTableKeys(ctx context.Context, jobs chan<- directiveJobT
...
 	partComp := newPartitionsComparer(fromD.TranslatePartitionsMap(), toPartitionsMap)
 
+	// Remove any partitions which are no longer assigned to this worker.
+	// TODO(tlt): currently, this is just removing the file lock on the
+	// resource; it's not actually removing the resource from the local
+	// computer. We should do that.
+	for tkey, partitions := range partComp.removed() {
+		qtid := tkey.QualifiedTableID()
+		for _, partition := range partitions {
+			resource := api.serverlessStorage.GetTableKeyResource(qtid, partition)
```
jaffee, line 334: serverlessStorage has `RemoveTableKeyResource` which unlocks the resource as it removes it

api_directive.go
```diff
@@ -324,6 +324,20 @@ func (api *API) pushJobsTableKeys(ctx context.Context, jobs chan<- directiveJobT
 	// Get the diff between from/to directive.partitions.
 	partComp := newPartitionsComparer(fromD.TranslatePartitionsMap(), toPartitionsMap)
 
+	// Remove any partitions which are no longer assigned to this worker.
+	// TODO(tlt): currently, this is just removing the file lock on the
+	// resource; it's not actually removing the resource from the local
+	// computer. We should do that.
```
jaffee, line 330: we should... I guess it's maybe a bit unclear how to do that... we probably need to add some API. I'm going to cut a ticket

jaffee: https://molecula.atlassian.net/browse/CLOUD-1345

api_directive.go
jaffee: (same for all the other types of resources of course)

dax/controller/http/handler.go
```diff
@@ -194,6 +195,32 @@ type DatabasesRequest struct {
...
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+}
+
+// DatabaseOptionRequest represents a change to a database option. The thinking
+// is to only support changing one database option at a time to keep the
+// implementation sane. At time of writing, only WorkersMin is supported.
+type DatabaseOptionRequest struct {
+	QualifiedDatabaseID dax.QualifiedDatabaseID `json:"qdbid"`
```
jaffee, line 219: is there any way we could change this json tag to "qdbpd"? just to really stick it to dyslexic folks?

dax/storage/storage.go
```diff
@@ -66,7 +66,7 @@ func (mm *ResourceManager) GetShardResource(qtid dax.QualifiedTableID, partition
 	}
 	mm.shardResources[key] = (&Resource{
 		snapshotter: mm.Snapshotter,
-		writeLogger: mm.Writelogger,
+		writelogger: mm.Writelogger,
```
jaffee, line 69: that's weird, I thought I grepped pretty thoroughly for these

api_directive.go
travisturner: wait, I don't want to actually remove the resource though right? because another node will be taking control of it. I thought I just need to release the lock. Or am I misunderstanding this?

dax/controller/http/handler.go
travisturner: i think i'm going to set up code generation to create these with random variable names for every instance.

api_directive.go
jaffee: you are misunderstanding (which is totally unrelated to my complete lack of doc comments), removal just removes the resource from management, it does not delete it. Deletion is handled by the controller, so there's no API in the storage manager to delete resources.

travisturner: changed it to use the `Remove*Resource` methods

sql3/parser/parser.go
```diff
@@ -3283,6 +3413,36 @@ func (p *Parser) parseIntegerLiteral(desc string) (*IntegerLit, error) {
...
+		return &stmt, p.errorExpected(p.pos, p.tok, "DATABASE")
+	}
+	stmt.Database, _, _ = p.scan()
+
+	if stmt.Name, err = p.parseIdent("database name"); err != nil {
+		return &stmt, err
+	}
+
+	switch p.peek() {
+	case SET:
```
paddyjok, line 3429: Consistency of syntax question, could we re-use the syntax of `CREATE....WITH`...the specified options being the ones to be altered - so this becomes `ALTER DATABASE foo WITH UNITS = new_units;`?

paddyjok (dismissed):
Noice. I left a couple of comments on syntax, but neither of them are a blocker.

sql3/parser/parser.go
```diff
@@ -355,10 +375,97 @@ func (p *Parser) parseDropStatement() (Statement, error) {
...
+		if tok != EXISTS {
+			return &stmt, p.errorExpected(pos, tok, "EXISTS")
+		}
+		stmt.IfNotExists = pos
+	}
+
+	if stmt.Name, err = p.parseIdent("database name"); err != nil {
+		return &stmt, err
+	}
+
```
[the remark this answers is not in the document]
travisturner: This is definitely not a hill I would die on, but with that said, I'm not familiar with `CREATE DATABASE` statements using `WITH`. Where do we use `WITH` that would make it consistent? In `BULK INSERT`, right? But not in `CREATE TABLE` unless I'm missing something.

Also, I was loosely following @gthrone25's [document](https://molecula.atlassian.net/browse/CLOUD-1330).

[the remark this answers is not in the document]
paddyjok: There is no standard for this, so we can do what we want. We use WITH in BULK INSERT, I think we should use WITH in create table, but don't (yet). I used WITH in CREATE MODEL (in another branch). All I am really saying is let's not miss an opportunity to get consistent on syntax, cause once it's out, it's really hard to change.

BTW - that ticket should probably be an Argon ticket in the SQL3 epic or at the least link to an Argon ticket.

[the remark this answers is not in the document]
gthrone25: Agreed there's no one syntax to rule them all here. Consistency is my number one priority, so @paddyjok  has the bigger picture of where things are moving to align. If we are going to add WITH table, it makes sense to add it here. Out of curiosity Pat, are you generally following SQL server for everything? I generally look across multiple DBs, but it'd help to know if that's your main reference.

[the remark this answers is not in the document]
paddyjok: Bit of SQL Server, bit of SQLLite, bit of PostgreSQL, bit of MySQL. I've be preferring to be more explicit than less and to remove some of the noise syntax where it makes sense (e.g SELECT [[ALL]|DISTINCT] doesn't need ALL, but we should support noise AS for aliases because people expect it). WITH falls into the 'more explicit' category. Another one TOP (instead of LIMIT) was in the explicit category because I didn't like that we might end up with TOP(N)/TOPK and LIMIT at the same time and that'd get weird. I think, knowing what we know now, and observing our (albeit) limited user pop. we should deep-six TOP in favor of LIMIT.

Let me model some of the current and upcoming syntax in ebnf/html and I will circulate for comment (we can re-use for doc) so we can decide and move on.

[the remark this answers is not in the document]
travisturner: +1 on `LIMIT` over `TOP`, but that's just because it's been 25 years since I've used SQL Server and `LIMIT` is burned into my brain from other databases. It also seems to align well with the `LIMIT, OFFSET` pattern should we ever decide to support `OFFSET`.

I have no problem with `WITH`. I'm not used to it, but being explicit seems useful and I like that it designates the start of an options list.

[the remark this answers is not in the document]
jaffee: +1 in LIMIT and WITH. I always found  the way the options were tacked on to CREATE TABLE a bit unsettling

paddyjok (approved):
Looks good

PR #2234, opened 2023-02-03
tweak a bunch of logging and config

jaffee:
make overall logs less verbose and chatty

1 minute computer check-in interval

3 minute snapshot interval

remove CaptureLogger as it has same functionality as buffer logger

add a WithPrefix to the Logger interface so sub-services can have different prefixes


jaffee:
just that TestQueryHistory thing is failing which is ticketed

travisturner (approved):

PR #2235, opened 2023-02-06
CLI: make it more like psql

travisturner:
I'm putting this here in case anyone wants to try it out, but I'm leaving it in DRAFT until I sort out the DATABASE-specific stuff between cloud/on-prem.


jaffee (approved):

jaffee:
race test is known failure in the transaction code due to timeouts... that code is deprecated and should get removed soon

PR #2239, opened 2023-02-10
improve SQL3 test coverage

seebs:
This is not ready for merging, but it's ready to start getting looked at. You can probably ignore the `golist` file, which is just some notes I was taking.


paddyjok (approved):
lgtm

PR #2241, opened 2023-02-10
Use the correct epoch when converting field.Options.Base to timestamp

travisturner:


paddyjok (approved):
Good to go - I will add some tests at the sql level to close out my ticket.

PR #2243, opened 2023-02-12
SUP-302

jrbrinlee1:
## Overview:

This PR add the ability to use capital theta in FeatureBase field names. This functionality was added as an immediate solution to the problem that dashes cannot be used in names of FeatureBase dataframe fields (if you want to compute against them with the Ivy query language). However, the customer needs the ability to delimit FeatureBase field and dataframe fields with "_" and one other single, common character. Since Ivy can use unicode characters in identifiers (all other non alpha-numeric, ascii characters are reserved or illegal), capital theta was added arbitrarily.


tgruben (changes requested):
We need to test it through sql3 as well, i'm moderately concerned it will cause the parser to barf, but it if doesn't work ping me and we can talk to pat

rachithrr (dismissed):
looks good to me

tgruben (approved):

PR #2245, opened 2023-02-13
Store version-check file in the configured data-directory

travisturner:
This also fixes what I think is a bug.
It also un-exports everything.

I have questions.

### DIFF
--- a/verchk.go
+++ b/verchk.go
@@ -6,25 +6,31 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/google/uuid"
 )
 
-type VersionChecker struct {
-	URL string
+const (
+	versionCheckerFilename = ".client_id.txt"
+)
+
+type versionChecker struct {
+	path string
+	url  string
 }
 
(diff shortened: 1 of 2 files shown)


verchk.go
```diff
@@ -38,60 +44,54 @@ func (v *VersionChecker) CheckIn() (*VerCheckResponse, error) {
...
 	return cleanedUUID, nil
 }
 
-func (v *VersionChecker) WriteClientUUID() (string, error) {
-	filename := ".client_id.txt"
-	_, err := os.Stat(filename)
-	if err != nil {
+func (v *versionChecker) writeClientUUID() (string, error) {
+	filename := filepath.Join(v.path, versionCheckerFilename)
+	if _, err := os.Stat(filename); err != nil {
```
seebs, line 77: This logic is potentially racy. If we want to open-existing-file if it exists, the canonical idiom is to try to open the file for reading. If we succeed, great. If we fail, we should use OpenFile with `O_EXCL|O_CREATE`, so in the event of a hypothetical race, either we win and we create the file, or we lose and someone else did. Alternatively, we could just try to create it exclusively, and if it already exists, then try to read it.

PR #2248, opened 2023-02-14
(fb-1903) - Fix SQL Fanout error

paddyjok:
This PR fixes an issue with header-based routing in the `http_handler` not working on a real cluster. The fix is to route by....well, route.

### DIFF
--- a/http_handler.go
+++ b/http_handler.go
@@ -11,7 +11,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"log"
 	"math"
 	"mime"
 	"net"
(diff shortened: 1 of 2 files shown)


http_handler.go
```diff
@@ -549,7 +549,7 @@ func newRouter(handler *Handler) http.Handler {
 		router.HandleFunc("/sql", handler.chkAuthZ(handler.handlePostSQL, authz.Admin)).Methods("POST").Name("PostSQL")
 	}
 	// internal endpoint
-	router.HandleFunc("/sql", handler.chkAuthZ(handler.handlePostSQLPlanOperator, authz.Admin)).Headers("X-FeatureBase-Plan-Operator", "").Methods("POST").Name("PostSQLPlanOperator")
```
travisturner, line 552: Alternatively, you could have used `.HeadersRegexp()` and matched on any value, rather than on an empty value. But I guess if you're not using that value for anything it doesn't matter.
I don't have a strong opinion on whether adding a new endpoint is better.

travisturner (approved):

PR #2249, opened 2023-02-14
Manual buffer management in restore-tar

tgruben:
fixes #fb-1912

### DIFF
--- a/ctl/backup_tar.go
+++ b/ctl/backup_tar.go
@@ -164,16 +164,17 @@ func (cmd *BackupTarCommand) Run(ctx context.Context) (err error) {
 	tw := tar.NewWriter(w)
 	defer tw.Close()
 
+	buf := new(bytes.Buffer)
 	// Backup schema.
 	if err := cmd.backupTarSchema(ctx, tw, schema); err != nil {
 		return fmt.Errorf("cannot back up schema: %w", err)
-	} else if err := cmd.backupTarIDAllocData(ctx, tw); err != nil {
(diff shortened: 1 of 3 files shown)


.golangci.yml
```diff
@@ -10,9 +10,7 @@ run:
     - pql/pql.peg.go
 linters:
   enable:
-    - govet
-    - gofmt
-    - staticcheck
+    - gosimple
```
seebs: Is this change intentional? It seems surprising to me because I was pretty sure we did in fact want those three, especially staticcheck.

ctl/multi-reader.go
```diff
@@ -0,0 +1,55 @@
+package ctl
+
+// Reader wraps an io.Reader, allowing "clones" to be created via the View method.
+import (
+	"bytes"
+	"io"
+	"sync"
+)
+
+type MultiReader struct {
```
seebs: I think we should pick a different name for this, because `io.MultiReader` already exists, for a reader which combines several other readers. Maybe `ReReader`?

seebs (dismissed):
so this looks great, but we should probably squash the commits and clean up the commit message? but otherwise this looks pretty reasonable. we do need better testing on this but you're already on that.

tgruben:
thx @seebs i'll squash after these test clear

seebs (approved):
i think there's a typo in an error message (`tran&slate`) but this otherwise looks good to me.

i particularly like removing the `// this feels wrong but works for now` comment on a thing that was apparently not working well enough. :)

seebs (approved):
fixed typo looks good!

PR #2250, opened 2023-02-14
(fb-1978) - performance regression

paddyjok:
This PR fixes a performance regression with the calculation of aggregates on ID/STRING types.

Currently, count(*) explicitly references the _id column and a recent change to allow non-pql aggregates on ID & STRING types did not take this into account.

This change will short circuit this check to force counts on _id columns to be executed using pql.Call.

### DIFF
--- a/sql3/planner/planoptimizer.go
+++ b/sql3/planner/planoptimizer.go
@@ -644,7 +644,13 @@ func tryToReplaceGroupByWithPQLAggregate(ctx context.Context, a *ExecutionPlanne
 					}
 					switch aggregable.AggExpression().Type().(type) {
 					case *parser.DataTypeID, *parser.DataTypeString:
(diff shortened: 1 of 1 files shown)


sql3/planner/planoptimizer.go
```diff
@@ -644,7 +644,13 @@ func tryToReplaceGroupByWithPQLAggregate(ctx context.Context, a *ExecutionPlanne
 					}
 					switch aggregable.AggExpression().Type().(type) {
 					case *parser.DataTypeID, *parser.DataTypeString:
-						return thisNode, true, nil
+						// if it is any other column other than _id bail
```
JasonAllison, line 647: Considering adding commentary on why the code was necessary, ie what behavior occurs if it wasnt there.

adrwalker (approved):

PR #2252, opened 2023-02-15
Aggregation Nation (fb-1955, fb-1887) 

paddyjok:
### Overview

This PR refactors aggregation an adds the ability to aggregate on expressions and not just references.

- Removed `AGGREGATE_ILLEGAL` thru `AGGREGATE_MAX` enum constants and associated stuff in `Aggregable` since each aggregate has it's own type and we can use that instead of the enum (fb-1658)
- Refactored `Aggregable` to be clearer  - replacing `AggExpression()` with `FirstChildExpr()` and `AggAdditionalExpr()` with `Children()`. For aggregation functions, this makes the simple 1 argument case more explicit and the multiple argument case easier.
- Enabled expressions to be used as aggregate terms for `sum()`, `avg()`, `min()` and `max()`
- Ensured that we do the right thing for `_id` columns if these are used as terms
- Simplified optimizer handling for aggregates

### Additional Notes

- This PR also fixes fb-1887 by making sure the `nodeid` column is part of the right table (`fb_exec_requests` in `sql3/planner/opsystemtable.go`
- This PR also introduces some proposed support for plan checking in tests using jsonpath

### DIFF
--- a/sql3/test/defs/defs.go
+++ b/sql3/test/defs/defs.go
@@ -2,7 +2,14 @@
 package defs
 
 import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
 	"time"
+
+	"github.com/PaesslerAG/gval"
+	"github.com/PaesslerAG/jsonpath"
 )
 
 // TableTests is the list of tests which get run by TestSQL_Execute in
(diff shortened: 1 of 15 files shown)


sql3/test/defs/defs.go
```diff
@@ -202,3 +209,27 @@ func timestampFromString(s string) time.Time {
 	}
 	return tm
 }
+
+func operatorPresentAtPath(jplan []byte, path string, operator string) error {
```
seebs, line 213: I feel like this function really wants a top level description of why it takes these inputs and what it does with them. I can sort of puzzle it out *what* it does from reading it, but it's much less obvious to me *why* that is a desireable thing to do.

seebs (changes requested):
The code looks good to me, but I think the addition of the json blob query plan thing probably calls for some additional comments for future maintainers, and a couple of comment updates.

sql3/test/helpers.go
```diff
@@ -13,34 +14,35 @@ import (
 )
 
 // MustQueryRows returns the row results as a slice of []interface{}, along with the columns.
```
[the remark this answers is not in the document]
paddyjok: Good call - changed doc

sql3/test/defs/defs.go
paddyjok: Yep - done.

seebs (approved):
Looks good. Sorry about the extra MustQueryRows making shiny new compiler errors.

PR #2253, opened 2023-02-15
task pool: avoid race condition on shutdown/close

seebs:
When we close a task pool, we use a condition variable to wait for workers to exit, if any workers are still running. The workers, in turn, use the condition variable to notify that they've exited. Unfortunately, the workers aren't using the lock (the rationale was that it's all atomic ops so they don't need to), which means that it's possible to have the following sequence:

	Close(): obtain current live count
	worker: decrement live count
	worker: send broadcast to condition variable
	Close(): wait on condition variable

To resolve this, we make the worker update also request the lock.

We add a simple reproducer for this. Note that simple doesn't mean it fires completely reliably; on my laptop, the test causes a test timeout about 60% of the time without the fix. If you add a short delay between sampling the live count and waiting on the condition variable, the deadlocks move from "60% chance of hitting it in a million trials" to "nearly always".

### DIFF
--- a/task/pool.go
+++ b/task/pool.go
@@ -85,7 +85,12 @@ func (p *Pool) Stats() (live, unblocked, target int) {
 
 // Close is a Shutdown followed by waiting for all jobs to exit.
 func (p *Pool) Close() {
+	// important to note: p.cond.Wait() is actually releasing this lock,
+	// then reacquiring it when the wait succeeds. This means that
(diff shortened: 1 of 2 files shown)


vkrishnanfb (approved):
The narrowest gap where 'wait' happens after the 'broadcast' do exist and it is closed with your fix. I am assuming with this fix TestPoolShutdown() should never deadlock, right?

PR #2254, opened 2023-02-16
Fb 1975

lorycloutier:
Cleaned up version checking, removed a race condition in opening/creating the UUID file, and added error checking. 
Added server flags for check-in endpoint and UUID storage file. 
Incorporates Travis's changes to store UUID file in data directory and unexport most of verchk.go.

### DIFF
--- a/verchk.go
+++ b/verchk.go
@@ -6,25 +6,31 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/google/uuid"
 )
 
-type VersionChecker struct {
-	URL string
+const (
+	versionCheckerFilename = ".client_id.txt"
+)
+
+type versionChecker struct {
+	path string
+	url  string
 }
 
-func NewVersionChecker(endpoint string) *VersionChecker {
-	v := VersionChecker{
-		URL: endpoint,
+func newVersionChecker(path, endpoint string) *versionChecker {
+	v := versionChecker{
+		path: path,
+		url:  endpoint,
 	}
 	return &v
 }
 
-func (v *VersionChecker) CheckIn() (*VerCheckResponse, error) {
-
-	id, err := v.WriteClientUUID()
+func (v *versionChecker) checkIn() (*verCheckResponse, error) {
+	id, err := v.writeClientUUID()
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


verchk.go
```diff
@@ -73,31 +70,8 @@ func (v *versionChecker) generateClientUUID() (string, error) {
...
-			}
-
-			return id, nil
-		} else {
-			return "", err
-		}
-	}
-
-	fh, err := os.Open(filename)
+	filename := filepath.Join(v.path, v.idfile)
```
seebs, line 73: Hmm. Okay, so, imagine that I, as a user, write `-uuid-file=/tmp/uuid`. We're going to then try to use the file name `/path/to/data//tmp/uuid`.

I think we should somehow handle the case where the path name is absolute. Note that it's not strictly portable to check the first character for being `/`, but I don't remember how to do it correctly in Go.

seebs (changes requested):
I want to handle the "user provides absolute path" case in a more user-friendly manner.

verchk.go
gardavi: `filepath.Join("/path/to/data/", "/tmp/uuid")` should give you `"/path/to/data/tmp/uuid"` (I might be missing something here)

verchk.go
```diff
@@ -73,31 +70,15 @@ func (v *versionChecker) generateClientUUID() (string, error) {
...
-			}
-
-			return id, nil
-		} else {
-			return "", err
-		}
+	var filename string
+	// if v.idfile starts with a path separator then it's an absolute path
+	// otherwise it's a relative path and the file goes in the data directory
+	if v.idfile[0:1] == string(os.PathSeparator) {
```
seebs, line 76: I think this can be spelled also as
```
if v.idfile[0] == os.PathSeparator
```

verchk.go
```diff
@@ -107,10 +88,24 @@ func (v *versionChecker) writeClientUUID() (string, error) {
 	if err != nil {
 		return "", err
 	}
+	//this just checks to see if there was anything at all in the file.
```
seebs, line 91: comments should have a space between the `//` and the comment text, because the compiler uses `//foo` type stuff for some compiler directives.

seebs (approved):
Have actually requested a minor typo-type change but this is otherwise fine, so go ahead and merge once that's done.

seebs (approved):
yay

PR #2256, opened 2023-02-16
Fb 1968 timestamp data type related fixes and enhancements

vkrishnanfb:
- Remove EPOCH column constraint from TIMESTAMP data type in SQL DDL
- Support implicit conversion of int to timestamp in SQL, treat the integer value as seconds from unix epoch.
- Add new scalar SQL function ToTimestamp(num, timeunit) to help convert integers values to Timestamp at user specified granular input.

### DIFF
--- a/sql3/planner/expression.go
+++ b/sql3/planner/expression.go
@@ -1572,6 +1572,8 @@ func (n *callPlanExpression) Evaluate(currentRow []interface{}) (interface{}, er
 		return n.EvaluateFormat(currentRow)
 	case "CHARINDEX":
 		return n.EvaluateCharIndex(currentRow)
+	case "TOTIMESTAMP":
+		return n.EvaluateToTimestamp(currentRow)
 	case "STR":
 		return n.EvaluateStr(currentRow)
 	default:
--- a/sql3/planner/expression.go
+++ b/sql3/planner/expression.go
@@ -41,7 +41,6 @@ func coerceValue(sourceType parser.ExprDataType, targetType parser.ExprDataType,
 				return nil, sql3.NewErrInternalf("unexpected value type '%T'", value)
 			}
 			return pql.NewDecimal(val*int64(math.Pow(10, float64(t.Scale))), t.Scale), nil
-
 		case *parser.DataTypeTimestamp:
 			val, ok := value.(int64)
 			if !ok {
@@ -65,7 +64,6 @@ func coerceValue(sourceType parser.ExprDataType, targetType parser.ExprDataType,
 				return nil, sql3.NewErrInternalf("unexpected value type '%T'", value)
 			}
 			return pql.NewDecimal(int64(val)*int64(math.Pow(10, float64(t.Scale))), t.Scale), nil
-
 		case *parser.DataTypeTimestamp:
 			val, ok := value.(int64)
 			if !ok {
--- a/sql3/planner/opinsert.go
+++ b/sql3/planner/opinsert.go
@@ -390,7 +390,8 @@ func (i *insertRowIter) Next(ctx context.Context) (types.Row, error) {
 						return nil, errors.Wrapf(err, "converting timestamp to int64: %s", v)
 					}
 					row.Values[posVals[idx]] = i64
-				//integers passed as input for Timestamp fields will be treated as time represented in number of seconds since unix epoch
+				//integers passed as input for Timestamp fields will be treated as time represented in number of seconds since epoch defined for the field
+				//for timestamp fields created using SQL epoch will be defaulted to unix epoch
 				case int64:
 					// Convert the input seconds to target timeunit defined for the Timestamp field
 					// Add Base, which is the base epoch for Timestamp fields, to the input before saving
--- a/sql3/planner/opinsert.go
+++ b/sql3/planner/opinsert.go
@@ -402,9 +402,9 @@ func (i *insertRowIter) Next(ctx context.Context) (types.Row, error) {
 					case fbbatch.TimeUnitMilliseconds:
 						i64 = i64 + (v * 1000)
 					case fbbatch.TimeUnitMicroseconds, fbbatch.TimeUnitUSeconds:
-						i64 = i64 + (v * 1000 * 1000)
+						i64 = i64 + (v * 1000000)
 					case fbbatch.TimeUnitNanoseconds:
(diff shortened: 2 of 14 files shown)


sql3/planner/expression.go
```diff
@@ -41,6 +41,13 @@ func coerceValue(sourceType parser.ExprDataType, targetType parser.ExprDataType,
 				return nil, sql3.NewErrInternalf("unexpected value type '%T'", value)
 			}
 			return pql.NewDecimal(val*int64(math.Pow(10, float64(t.Scale))), t.Scale), nil
+		case *parser.DataTypeTimestamp:
```
paddyjok, line 44: You may have to change the `typesAreAssignmentCompatible()` function in addition to these changes.

`typesAreAssignmentCompatible()` is called when we are doing an assignment - i.e. expression in an insert list, expression passed as function call parameter, future assignment statement when we have stored code.

paddyjok: We could revisit some tests here too in the following areas - not just bulk insert. So something like:

1. ensuring we haven't broken `CAST(1234567 as timestamp)`
2. ensuring `insert into testable (_id, timestampcol) values (1, 1234567)` works as expected
3. ensuring `select datepart('YY', 1234567)` works as expected

vkrishnanfb: Tested simple insert and datepart() and cast() functions and found things broken, and made necessary updates in expressiontypes.go and opinsert.go to fix.

sql3/planner/opinsert.go
```diff
@@ -336,7 +336,25 @@ func (i *insertRowIter) Next(ctx context.Context) (types.Row, error) {
 						return nil, errors.Wrapf(err, "converting timestamp to int64: %s", v)
 					}
 					row.Values[posVals[idx]] = i64
-
+				//integers passed as input for Timestamp fields will be treated as time represented in number of seconds since unix epoch
+				case int64:
+					// Convert the input seconds to target timeunit defined for the Timestamp field
+					// Add Base, which is the base epoch for Timestamp fields, to the input before saving
+					unit := fbbatch.TimeUnit(opts.TimeUnit)
+					i64 := opts.Base
```
seebs, line 344: the comments say "seconds since unix epoch", but it looks like here we're using "opts.Base", which might not be the Unix epoch.

sql3/planner/expression.go
```diff
@@ -57,6 +64,13 @@ func coerceValue(sourceType parser.ExprDataType, targetType parser.ExprDataType,
 				return nil, sql3.NewErrInternalf("unexpected value type '%T'", value)
 			}
 			return pql.NewDecimal(int64(val)*int64(math.Pow(10, float64(t.Scale))), t.Scale), nil
+		case *parser.DataTypeTimestamp:
```
seebs, line 67: I would expect us to allow int64, but also to accept strings that contained timestamps, but this seems to only accept int64, and NOT accept strings of timestamps?

seebs (commented):
I don't fully understand this, so I've asked a couple of questions here (and also asked in the ticket to be sure people just watching the ticket see the question). So I'm not sure this is wrong but I'm uncertain, and want to know more. The code and test cases mostly look reasonable to me.

sql3/test/defs/defs_timestamp_literals.go
```diff
@@ -27,11 +29,34 @@ var timestampLiterals = TableTest{
...
+		{
+			SQLs: sqls(
+				"select datepart('yy', ts) as \"yy\" from testtimestampliterals",
+			),
+			ExpHdrs: hdrs(
+				hdr("yy", fldTypeInt),
+			),
+			ExpRows: rows(
+				row(int64(1970)),
+				row(int64(time.Now().UTC().Year())),
```
seebs, line 56: okay, i don't fully understand how current_timestamp works, and i know that this will come up extremely rarely if at all, but:

i think it's possible for this test to fail because the computation of `time.Now()` is separate from the computation/handling of `current_timestamp`. So for instance, if we build the table in featurebase, and then a millisecond later we compute our expected rows, that can result in different years. And I recognize that this is very rare, but conceptually, I think we should ensure that the timestamps we're testing against are *identical* to the timestamps that are in the database, not just probably-similar. Even if the probability is > 99.9999%.

seebs (commented):
oops missed a thing in previous comment.

sql3/planner/expression.go
vkrishnanfb: We do accept strings as input to timestamp, this existing functionality starts at line 82, case section handling string source type and timestamp target type.

sql3/planner/opinsert.go
vkrishnanfb: Comment is updated to say fields can have custom epoch, but fields created using SQL will default to unix epoch.

sql3/test/defs/defs_timestamp_literals.go
vkrishnanfb: Comparison test is updated to look for expected year only for those cases where we can pre-determine the year.

sql3/planner/expression.go
seebs: I always forget that coerceValue is switching on source type first, target type second.

sql3/test/defs/defs_date_functions.go
```diff
@@ -191,5 +218,28 @@ var datePartTests = TableTest{
...
+			//test ToTimestamp(num, timeunit) for all possible time unit values
+			SQLs: sqls(
+				"select totimestamp(1000) as \"default\", totimestamp(1000, 's') as \"s\", totimestamp(1000000, 'ms') as \"ms\", totimestamp(1000000000, 'us') as \"us\", totimestamp(1000000000, 'µs') as \"µs\", totimestamp(1000000000000, 'ns') as \"ns\"",
+			),
+			ExpHdrs: hdrs(
+				hdr("default", fldTypeTimestamp),
+				hdr("s", fldTypeTimestamp),
+				hdr("ms", fldTypeTimestamp),
+				hdr("us", fldTypeTimestamp),
+				hdr("µs", fldTypeTimestamp),
```
seebs, line 231: I appreciate supporting the actual µs as an option.

seebs (approved):
I think we should squash the commits and rebase but this otherwise looks good to me.

tgruben (approved):

PR #2257, opened 2023-02-16
ORDER BY ....what now!? (fb-1954)

paddyjok:
This PR handles the case of a term appearing in the  ORDER BY clause for a term not in the projection list for a SELECT.

Additional compiler behavior has been changed thus:

1. The compiler will now place the PlanOpOrderBy op before the PlanOpProjection op for in cases where there are only references  in the order by expression list (or positional expressions that can be resolved to references.
2. The compiler will also handle the case of a reference appearing in the order by expression list that is not projected.
3. If the order by list contains non reference expressions and references that do not appear in the projection list, the compiler will: a) generate a new projection containing the original projection's schema, b) make the order by operator a child of that projection op and c) make the original projection (with the addition of the unprojected order by expression) a child of the order by operator.

Test coverage has been added to cover these cases.

### DIFF
--- a/sql3/planner/compileselect.go
+++ b/sql3/planner/compileselect.go
@@ -11,7 +11,6 @@ import (
 	"github.com/featurebasedb/featurebase/v3/sql3"
 	"github.com/featurebasedb/featurebase/v3/sql3/parser"
 	"github.com/featurebasedb/featurebase/v3/sql3/planner/types"
-	"github.com/pkg/errors"
 )
 
 // compileSelectStatment compiles a parser.SelectStatment AST into a PlanOperator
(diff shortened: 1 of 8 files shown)


sql3/planner/compileselect.go
```diff
@@ -129,9 +128,49 @@ func (p *ExecutionPlanner) compileSelectStatement(stmt *parser.SelectStatement,
...
+			if ot.Desc.IsValid() {
+				f.Order = orderByDesc
+			}
+			orderByExprs = append(orderByExprs, f)
+		}
+
+		// if the expression is just references, we
+		// can put the sort directly after the source
+		for _, oe := range orderByExprs {
+			_, ok := oe.Expr.(*qualifiedRefPlanExpression)
```
seebs, line 155: I'm not sure I understand this. So, what I'm inferring is that if we compile a thing, and it is actually already present in the plan, we always get a `qualifiedRefPlanExpression`, but we can't get that if it's not already present in the plan? I guess the name sort of implies that.

seebs (approved):
this looks reasonable to me but i have some hesitance because honestly, i don't think i fully understand the relationships between projections and order-bys and things. but i can follow along and it looks reasonable, and the test coverage hits several of the cases. and i believe we already had test cases for order by with invalid fields, so we're not just gonna blow up if someone uses a column that isn't even valid, so... go for it!

sql3/planner/compileselect.go
paddyjok: I don't understand the question?

PR #2258, opened 2023-02-17
Make count(*) not depend on _id (fb-1974)

paddyjok:
This PR alters the behavior of `count(*)` to not depend on the `_id` column until the optimizer is sure that the operation is going to end up in a `pql.Call` graph.

This PR also replaces relevant instances of `"_id"` with he appropriate constant.

### DIFF
--- a/sql3/planner/compilealtertable.go
+++ b/sql3/planner/compilealtertable.go
@@ -92,7 +92,7 @@ func (p *ExecutionPlanner) analyzeAlterTableStatement(stmt *parser.AlterTableSta
 			return sql3.NewErrUnknownType(col.Type.Name.NamePos.Line, col.Type.Name.NamePos.Column, typeName)
 		}
 
-		if strings.ToLower(columnName) == "_id" {
+		if strings.ToLower(columnName) == string(dax.PrimaryKeyFieldName) {
 			//not allowed to add an _id column after the fact
 			return sql3.NewErrTableIDColumnAlter(col.Name.NamePos.Line, col.Name.NamePos.Column)
 		}
--- a/sql3/planner/compileselect.go
+++ b/sql3/planner/compileselect.go
@@ -83,12 +83,12 @@ func (p *ExecutionPlanner) compileSelectStatement(stmt *parser.SelectStatement, 
 		for _, agg := range aggregates {
 			InspectExpression(agg, func(expr types.PlanExpression) bool {
 				switch ex := expr.(type) {
-				case *sumPlanExpression, *countPlanExpression, *countDistinctPlanExpression,
-					*avgPlanExpression, *minPlanExpression, *maxPlanExpression,
-					*percentilePlanExpression:
+				case types.Aggregable:
 					ch := ex.Children()
-					// first arg is always the ref
-					aggregateAndGroupByExprs = append(aggregateAndGroupByExprs, ch[0])
+					// first arg is always the ref, except for count(*)
+					if len(ch) > 0 {
+						aggregateAndGroupByExprs = append(aggregateAndGroupByExprs, ch[0])
+					}
 					return false
 				}
 				return true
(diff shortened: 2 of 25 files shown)


travisturner (approved):

sql3/planner/compilealtertable.go
```diff
@@ -92,7 +92,7 @@ func (p *ExecutionPlanner) analyzeAlterTableStatement(stmt *parser.AlterTableSta
 			return sql3.NewErrUnknownType(col.Type.Name.NamePos.Line, col.Type.Name.NamePos.Column, typeName)
 		}
 
-		if strings.ToLower(columnName) == "_id" {
+		if strings.ToLower(columnName) == string(dax.PrimaryKeyFieldName) {
```
JasonAllison, line 95: Do all tables have PrimaryKeyFieldName?  How is this field denoted?

sql3/planner/compileselect.go
```diff
@@ -83,12 +83,12 @@ func (p *ExecutionPlanner) compileSelectStatement(stmt *parser.SelectStatement,
...
-					*percentilePlanExpression:
+				case types.Aggregable:
 					ch := ex.Children()
-					// first arg is always the ref
-					aggregateAndGroupByExprs = append(aggregateAndGroupByExprs, ch[0])
+					// first arg is always the ref, except for count(*)
+					if len(ch) > 0 {
+						aggregateAndGroupByExprs = append(aggregateAndGroupByExprs, ch[0])
+					}
 					return false
```
JasonAllison, line 92: I dont understand what is going on here.  Why is this returning false regardless of the conditional?  What purpose does this serve?  What is the use case of ch == 0?

Why not:
`if len(ex.Children()) > 0 {
  aggregateAndGroupByExprs = append(aggregateAndGroupByExprs, ex.Children()[0])
}`
Why create the new memory structure?  Is this a Go thing?

sql3/planner/oppqlaggregate.go
```diff
@@ -64,7 +64,7 @@ func (p *PlanOpPQLAggregate) Schema() types.Schema {
 	s := &types.PlannerColumn{
 		ColumnName:   "",
 		RelationName: "",
-		Type:         p.aggregate.FirstChildExpr().Type(),
+		Type:         p.aggregate.Type(),
```
JasonAllison, line 67: Why Type() on line 67 but String() on 45?

sql3/planner/compilealtertable.go
paddyjok: All on-disk FeatureBase tables have an "_id" column implied - this just replaces the "_id" string with a constant.

sql3/planner/compileselect.go
paddyjok: The `return false` is for the closure that's passed to `InspectExpression`, it basically means stop inspecting. 

The `ch := ex.Children()` is just being explicit. The compiler will optimize it away.

sql3/planner/oppqlaggregate.go
paddyjok: They are two different things - on line 45 we are just populating the plan, we don't need the whole type, just the text description. On 67 we are returning the Schema for the operator - this needs the complex type.

sql3/planner/oppqltablescan.go
```diff
@@ -113,6 +113,19 @@ func (p *PlanOpPQLTableScan) WithChildren(children ...types.PlanOperator) (types
...
 }
 
+func (p *PlanOpPQLTableScan) PrimaryKeyType() (parser.ExprDataType, error) {
+	tname := dax.TableName(p.tableName)
+	table, err := p.planner.schemaAPI.TableByName(context.Background(), tname)
+	if err != nil {
+		return nil, err
+	}
+
+	if table.StringKeys() {
```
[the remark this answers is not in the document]
paddyjok: This function returns the type of the primary key (_id) column used in the scan op. If `table.StringKeys() == true` then we return a string type else it's an ID type.

PR #2259, opened 2023-02-17
controller commit to DB first then send directives

jaffee:


travisturner (approved):
This all makes sense to me. Do we still need the "2" after `buildDirectives` and `sendDirectives`?

travisturner (approved):

PR #2260, opened 2023-02-17
Build target also builds fbsql, fbsql is also packaged.

j3m7:


gthrone25:
@j3m7 is this packaging the cli separately?

adrwalker:
What would happen in Cloud if this fbsql wasn't built and packaged?

nfpm.yaml
```diff
@@ -10,6 +10,8 @@ homepage: "https://molecula.com"
 contents:
   - src: ./featurebase
     dst: /usr/bin/featurebase
+  - src: ./fbsql
```
j3m7, line 13: @gthrone25 fbsql is a separate executable packaged in the same package as featurebase. The featurebase cli command no longer exists and this replaces is.

Makefile
```diff
@@ -111,6 +111,8 @@ cover-viz: cover
 # Compile Pilosa
 build:
 	$(GO) build -tags='$(BUILD_TAGS)' -ldflags $(LDFLAGS) $(FLAGS) ./cmd/featurebase
+	$(GO) build -tags='$(BUILD_TAGS)' -ldflags $(LDFLAGS) $(FLAGS) ./cmd/fbsql
```
j3m7, line 114: @adrwalker 
> What would happen in Cloud if this fbsql wasn't built and packaged?

The packaging process would fail, this adds the fbsql executable as part of the make target so it's built every time CI builds the DB

nfpm.yaml
gthrone25: ohhh nice! Gonna add a story to discuss adding a download in the UI for that

adrwalker (approved):
LGTM

PR #2261, opened 2023-02-17
forward-port tests from SQL1 tree

seebs:
This forward-ports a number of tests from the previous SQL implementation. The porting is approximate in a number of ways, and not all tests are implemented/tested yet.

In particular, several tests are currently disabled because we don't support `limit n` constructs.

The tests that were primarily tests of the parser have been brought forward as parser tests. One of them has been altered to add parentheses, because our parser interprets
	`fld1 between 1 and 3 and fld2 = 2`
as:
	`fld1 between (1 and 3) and (fld2 = 2)`
which is invalid, while the old parser apparently interpreted it as:
	`(fld1 between 1 and 3) and (fld2 = 2)`

We have not yet verified the SQL spec's requirements here, but sqlite agrees with our old parser, not our new parser, so this may be a regression.

The old tests expected an INNER JOIN to suppress duplicate values. Our new code does not, which is consistent with other SQL implementations. This is a change, but the old behavior appears to have been wrong. (You can still suppress duplicate values by specifying DISTINCT.)

In the previous implementations, a value like `count(*)` had `count(*)` as its column name. In the new implementation, it has an empty string as its column name.

Related to this, the prior implementation allowed you to write
	`select age, count(*) from grouper group by age having count > 1`
but the new implementationt requires that to be spelled as
	`having count(*) > 1`

This is consistent with other SQL implementations, so I think the new behavior is correct.

The behavior of SHOW COLUMNS and SHOW TABLES has changed, in that the specific results returned are significantly different. Perhaps more significantly, the old system spelled the former query as SHOW FIELDS, rather than SHOW COLUMNS. This may be considered a regression, in that `SHOW FIELDS` no longer works, and we should consider whether any hypothetical users might have been relying on the output of either of these. (I hope not, the new output is much better.)

Some of the old tests (the ones in handler\_test) were accommodated by adding a couple of specific test cases to existing tests, specifically:
	* handling timestamp values with `Z` rather than `+00:00`
	* a join with a WHERE clause referring to fields in both source tables

### DIFF
--- a/sql3/sql_test.go
+++ b/sql3/sql_test.go
@@ -73,9 +73,15 @@ func TestSQL_Execute(t *testing.T) {
 								exp[i] = make([]interface{}, len(headers))
 								for j := range sqltest.ExpHdrs {
 									targetIdx := m[sqltest.ExpHdrs[j].Name]
-									assert.GreaterOrEqual(t, len(sqltest.ExpRows[i]), len(headers),
-										"expected row set has fewer columns than returned headers")
-									exp[i][targetIdx] = sqltest.ExpRows[i][j]
+									if sqltest.Compare != defs.ComparePartial {
+										assert.GreaterOrEqual(t, len(sqltest.ExpRows[i]), len(headers),
+											"expected row set has fewer columns than returned headers")
+									}
+									// if ExpRows[i] is short, that might be okay if we're doing a "partial"
+									// compare.
+									if len(sqltest.ExpRows[i]) > j {
+										exp[i][targetIdx] = sqltest.ExpRows[i][j]
+									}
 								}
 							}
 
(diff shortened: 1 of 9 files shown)


sql3/sql_test.go
```diff
@@ -95,6 +101,32 @@ func TestSQL_Execute(t *testing.T) {
 								for _, row := range rows {
 									assert.Contains(t, exp, row)
 								}
+							case defs.ComparePartial:
```
paddyjok, line 104: What do we need this for?

seebs: We have tests of `SHOW TABLES` and `SHOW COLUMNS`. The results from those include timestamps of when things were created, making it impossible for statically-defined test data to match them. The best solution I could think of was to add support for "we know the following about the results"; for instance, "we should have a field named `age` with type `int`", but we can't actually write the full row of the response.

This might apply in other cases where the full results from a query aren't predictable, but we can say things about what has to be there. This is similar to the existing `CompareIncludedIn`, but that is testing that full rows are somewhere in the output results, and we can't actually provide full rows.

paddyjok: Got it.

bruce-b-molecula (approved):

PR #2262, opened 2023-02-17
add support for time quantum inserts with explicit timestamps (fb-1558)

paddyjok:
This PR adds support for time quantum inserts with explicit timestamps.

Oh, and tests.

### DIFF
--- a/sql3/errors.go
+++ b/sql3/errors.go
@@ -709,7 +709,7 @@ func NewErrInsertValueOutOfRange(line, col int, columnName string, rowNumber int
 
 func NewErrUnexpectedTimeQuantumTupleLength(line, col int, columnName string, rowNumber int, badValue []interface{}, length int) error {
 	return errors.New(
-		ErrInsertValueOutOfRange,
+		ErrUnexpectedTimeQuantumTupleLength,
 		fmt.Sprintf("[%d:%d] inserting value into column '%s', row %d, value '%v' out of range", line, col, columnName, rowNumber, badValue),
 	)
 }
--- a/sql3/errors.go
+++ b/sql3/errors.go
@@ -109,7 +109,8 @@ const (
 
 	// insert errors
 
-	ErrInsertValueOutOfRange errors.Code = "ErrInsertValueOutOfRange"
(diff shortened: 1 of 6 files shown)


sql3/errors.go
```diff
@@ -706,6 +707,13 @@ func NewErrInsertValueOutOfRange(line, col int, columnName string, rowNumber int
 	)
 }
 
+func NewErrUnexpectedTimeQuantumTupleLength(line, col int, columnName string, rowNumber int, badValue []interface{}, length int) error {
```
seebs, line 710: Unexpected Time Quantum Tuple Length

value `%v` out of range

I don't feel like these describe the same problem?

paddyjok: oops - copy pasta

sql3/planner/opinsert.go
```diff
@@ -264,6 +265,59 @@ func (i *insertRowIter) Next(ctx context.Context) (types.Row, error) {
...
+						return nil, sql3.NewErrUnexpectedTimeQuantumTupleLength(0, 0, columnName, rowNumber+1, v, len(v))
+					}
+
+					tupleType, ok := iv.Type().(*parser.DataTypeTuple)
+					if !ok {
+						return nil, sql3.NewErrInternalf("unexpected tuple type '%T'", v[0])
+					}
+
+					// first member must be a timestamp or coercable as one
+					cval, err := coerceValue(tupleType.Members[0], parser.NewDataTypeTimestamp(), v[0], parser.Pos{Line: 0, Column: 0})
```
seebs, line 281: do we not have a valid position we can use here?

also... should the timestamp be the first value? my intuition would have been that the timestamp was the last value in the tuple. did we specify this anywhere? is there a thing we're doing this to be compatible with?

seebs (approved):
I think this looks okay, but I'm a little uncertain about the order of the tuple, because I thought we had historically done (value, timestamp) for inserting things with timestamps. Changing that isn't the end of the world but I want to at least bring it up.

sql3/planner/opinsert.go
paddyjok: Completely arbitrary. If someone can point to prior art where it's the other way, happy to change it.

PR #2264, opened 2023-02-17
create model, create function... all the goodies

paddyjok:
This PR wraps all the work done in Q1 around models, functions and other various pieces.

1. Implements experimental `COPY` statement that allows copying data from one table to another table in the database, or between a database and the cloud
2. Implements experimental `CREATE MODEL` and associated `DROP MODEL` statements that enable a vehicle to embed simple 'native' statistical models like linear regression into the database
3. Implements experimental `PREDICT` statement that allows a users to inference on a previously created model
4. Implements the bones of `CREATE FUNCTION` that allows the creation of stored code - this has been stubbed out post demos done on this branch
5. Re-introduces the `LIMIT` clause on select statements. Semantically identical to `TOP`. Queries can have either `TOP` or `LIMIT`, it is en error to have both. `TOP` will be deprecated in time.
6. Implements `CORR()` aggregate function - returns the correlation of two sets of values
7. Implements `VAR()` aggregate function - variance
8. Refactors `DateLiteralPlanExpression` --> `TimestampLiteralPlanExpression`
9. Adds relevant tests

### DIFF
--- a/sql3/parser/ast_test.go
+++ b/sql3/parser/ast_test.go
@@ -404,11 +404,8 @@ func TestCreateFunctionStatement_String(t *testing.T) {
 				Type: &parser.Type{Name: &parser.Ident{Name: "int"}},
 			},
 		},
-		ReturnDef: &parser.ParameterDefinition{
-			Name: &parser.Variable{Name: "@scalar"},
-			Type: &parser.Type{Name: &parser.Ident{Name: "int"}},
-		},
-	}, `CREATE FUNCTION func (@param1 int) RETURNS @scalar int AS BEGIN END`)
+		ReturnType: &parser.Type{Name: &parser.Ident{Name: "int"}},
+	}, `CREATE FUNCTION func (@param1 int) RETURNS int AS BEGIN END`)
 
 	AssertStatementStringer(t, &parser.CreateFunctionStatement{
 		IfNotExists: pos(0),
@@ -419,11 +416,8 @@ func TestCreateFunctionStatement_String(t *testing.T) {
 				Type: &parser.Type{Name: &parser.Ident{Name: "int"}},
 			},
 		},
-		ReturnDef: &parser.ParameterDefinition{
-			Name: &parser.Variable{Name: "@scalar"},
-			Type: &parser.Type{Name: &parser.Ident{Name: "int"}},
-		},
-	}, `CREATE FUNCTION IF NOT EXISTS func (@param1 int) RETURNS @scalar int AS BEGIN END`)
+		ReturnType: &parser.Type{Name: &parser.Ident{Name: "int"}},
+	}, `CREATE FUNCTION IF NOT EXISTS func (@param1 int) RETURNS int AS BEGIN END`)
 }
 
 func TestCreateViewStatement_String(t *testing.T) {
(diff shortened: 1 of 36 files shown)


sql3/parser/ast_test.go
```diff
@@ -914,38 +908,38 @@ func TestSelectStatement_String(t *testing.T) {
...
 	// }, `WITH "cte" ("x", "y") AS (SELECT *) VALUES (1, 2), (3, 4)`)
 
-	AssertStatementStringer(t, &parser.SelectStatement{
-		Columns: []*parser.ResultColumn{{Star: pos(0)}},
-		Union:   pos(0),
-		Compound: &parser.SelectStatement{
-			Columns: []*parser.ResultColumn{{Star: pos(0)}},
-		},
-	}, `SELECT * UNION SELECT *`)
+	// AssertStatementStringer(t, &parser.SelectStatement{
```
seebs, line 911: I'm a bit concerned about a lot of these getting commented out at once, because they used to work, so if we're having to comment them out, that feels like a regression?

seebs:
There's some commented-out tests in `parser_test.go` around line 3354 that were from the forward-port of tests from sql1. Well, not commented out, inside an `if false { ... }`. Sorry.

They won't pass because right now they don't have the actual statement trees that the parser should produce, but we should add those and enable them now.

sql3/planner/executionplanner.go
```diff
@@ -154,12 +169,19 @@ func (p *ExecutionPlanner) analyzePlan(ctx context.Context, stmt parser.Statemen
 		return nil
 	case *parser.DropViewStatement:
 		return nil
+	case *parser.DropModelStatement:
```
seebs, line 172: Looking at this, I'm struck by the asymmetry that we support both create and drop for models, but not for functions. Should we support drop function also?

sql3/planner/expression.go
```diff
@@ -2901,7 +2914,15 @@ func (p *ExecutionPlanner) compileCallExpr(expr *parser.Call) (_ types.PlanExpre
 		return agg, nil
 
 	default:
-		return newCallPlanExpression(parser.IdentName(expr.Name), args, expr.ResultDataType), nil
+		// could be a udf - try to look it up in functions
+		fn, err := p.getFunctionByName(strings.ToLower(callName))
+		if err != nil {
+			return nil, err
+		}
+		if fn != nil {
```
seebs, line 2922: So far as I can tell, this branch is irrelevant, because this cashes out to

```
if x != nil {
   return foo(x)
}
return foo(nil)
```

But that can actually just be done by `return foo(x)`. similarly, we can just call `newCallPlanExpression(..., fn)`, and if fn is nil, we'll do exactly what we're currently doing if fn is nil anyway, which is passing a nil in that slot.

sql3/planner/expressionagg.go
```diff
@@ -943,6 +944,311 @@ func (n *percentilePlanExpression) WithChildren(children ...types.PlanExpression
 	return newPercentilePlanExpression(children[0], children[1], n.returnDataType), nil
 }
 
+// aggregator for CORR()
+type aggregateCorr struct {
```
seebs, line 948: I feel like somewhere in here, we could really use a comment explaining what is actually happening. I last took a statistics class something like 33 years ago and I do not actually know what is happening here. Some quick summary of the name of the algorithm or something, and an explanation of what we're tracking and why tracking it is useful seems rewarding.

sql3/planner/expressionagg.go
```diff
@@ -943,6 +944,311 @@ func (n *percentilePlanExpression) WithChildren(children ...types.PlanExpression
...
+		}
+
+		val = thisVal.Float64()
+
+	case *parser.DataTypeID:
+		thisVal, ok := v.(int64)
+		if !ok {
+			return sql3.NewErrInternalf("unexpected type conversion '%T'", v)
+		}
+
```
seebs, line 1155: I think we're a ways out from naturally-generated IDs out of range, but we have historically at least in principle supported pretty close to a full 63-bit range of ID values. I feel like we should probably support int64 values which aren't converted to float64 if at all possible.

sql3/planner/expressionagg.go
```diff
@@ -943,6 +944,311 @@ func (n *percentilePlanExpression) WithChildren(children ...types.PlanExpression
...
+}
+
+func (n *corrPlanExpression) WithChildren(children ...types.PlanExpression) (types.PlanExpression, error) {
+	if len(children) != 2 {
+		return nil, sql3.NewErrInternalf("unexpected number of children '%d'", len(children))
+	}
+	return newCorrPlanExpression(children[0], children[1], n.returnDataType), nil
+}
+
+// aggregator for VAR()
```
seebs, line 1108: Again, would love a top-level comment on this explaining what we're tracking and why tracking it is interesting or useful.

sql3/planner/opcopy.go
```diff
@@ -0,0 +1,515 @@
...
+func (p *PlanOpCopy) Schema() types.Schema {
+	return types.Schema{}
+}
+
+func (p *PlanOpCopy) Iterator(ctx context.Context, row types.Row) (types.RowIterator, error) {
+	child, err := p.ChildOp.Iterator(ctx, row)
+	if err != nil {
+		return nil, err
+	}
+	if len(p.url) > 0 {
```
seebs, line 55: i think this is unidiomatic compared to `if p.url != ""`

sql3/planner/opcopy.go
```diff
@@ -0,0 +1,515 @@
...
+					if !ok {
+						return nil, sql3.NewErrInternalf("unexpected type '%T'", row[i])
+					}
+					fmt.Fprintf(&rowBuf, `"%s"`, val.Format(time.RFC3339Nano))
+
+				case *parser.DataTypeStringSet:
+					val, ok := row[i].([]string)
+					if !ok {
+						return nil, sql3.NewErrInternalf("unexpected type '%T'", row[i])
+					}
```
seebs, line 447: i believe in both this case and DataTypeIDSet, we should check for/support NULL values.

I *think* that null values should always show up as a slice that is nil, as contrasted with a zero-length slice, but I slightly distrust that, and think we should in full generality probably accept a nil in `row[i]` as a NULL for the inserted value. I believe in inserts we can write a literal NULL in those slots.

sql3/planner/oppredict.go
```diff
@@ -0,0 +1,253 @@
...
+		childSchema:   childSchema,
+		regres:        new(regression.Regression),
+		inferenceRefs: make([]*qualifiedRefPlanExpression, 0),
+	}
+}
+
+func (i *linearRegressionPredictIter) Next(ctx context.Context) (types.Row, error) {
+	if i.hasStarted == nil {
+		// label column
+		i.regres.SetObserved("Murders per annum per 1,000,000 inhabitants")
```
seebs, line 144: i think i was expecting the linear regression to be more user-defined and less hard-coded?

seebs (commented):
Looking at this, I'm reminded of a thing I saw once which had a couple of rednecks doing electronics repair, and they opened something up, and stared at it for a bit, and then one of them said "Gosh, Vern. Lookit all them wires in there."

That's basically where I'm at here. I don't feel like I fully understand the model/predict/etc functionality, and would benefit from some higher level doc comments outlining what these things are and why they're like that.

I made a few more specific comments in-line. I don't yet feel comfortable approving, but I'm not actively disapproving, although I think some of the comments merit fixing. I'll have another look if it isn't approved by someone else in the next day or so.

sql3/planner/expressionagg.go
paddyjok: https://lmgtfy.app/?q=correlation+coefficient

tgruben (approved):

paddyjok:
> There's some commented-out tests in `parser_test.go` around line 3354 that were from the forward-port of tests from sql1. Well, not commented out, inside an `if false { ... }`. Sorry.
> 

> They won't pass because right now they don't have the actual statement trees that the parser should produce, but we should add those and enable them now.
@seebs do you have the code for these - they may have been mangled in a rebase...

sql3/parser/ast_test.go
paddyjok: Yeah the parser test may have worked, but no syntax support for union yet....

sql3/planner/expression.go
paddyjok: Yep - good catch

sql3/planner/executionplanner.go
paddyjok: Yes, but functions are (very) experimental, so drop function can come later.

sql3/planner/expressionagg.go
paddyjok: can't do variance without floats, so everything with these new types is float.

sql3/planner/opcopy.go
paddyjok: Do we know for a fact we can have a null value in a set?

sql3/planner/oppredict.go
paddyjok: Yeah...it should pull this out of the schema...

sql3/planner/opcopy.go
paddyjok: fixed

tgruben (approved):

PR #2267, opened 2023-02-21
improve ramdisk config in Makefile, use it for everything in tests

seebs:
So, we were special-casing creating a ramdisk, and setting a special environment variable for it, for boltdb translate files, to improve performance.

But actually, etcd and test cluster data and so on all go in $TMPDIR, and if you move all of those also into a ram disk, you get way better performance. But 2GB may not be enough for that.

So!

We unify on $TMPDIR, we bump the default size to 4GB, we make the size configurable, and we stop using the name RAMDISK. This should improve performance on MacOS significantly for `make test` and things like it, and also simplifies our lives by not having a special case for the boltdb translate files.

### DIFF
--- a/.gitlab/.gitlab-ci.yml
+++ b/.gitlab/.gitlab-ci.yml
@@ -216,7 +216,7 @@ run go tests race:
   script:
     - echo "Running featurebase race tests..."
     - PKG_LIST=$(go list ./... | grep -Ev 'internal/clustertests|simulacraData|batch|idk|v3/dax/test/dax' | paste -s -d, -)
-    - RAMDISK=/mnt/ramdisk go test -race -v -timeout=10m ${PKG_LIST//,/ }
+    - TMPDIR=/mnt/ramdisk go test -race -v -timeout=10m ${PKG_LIST//,/ }
   tags:
     - docker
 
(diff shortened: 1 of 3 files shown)


seebs:
update: with 4GB ramdisk, `make test` tends to fail for me, with 8GB, it completes for me.

on my mac laptop, `make test` runs in 3 minutes with this change. i haven't had it complete yet running without the ramdisk, and i'll be checking the timing with just the `RAMDISK` setting rather than this change if and when that completes.

seebs:
So, using master branch and RAMDISK: about 9 minutes to complete `make test`

So this is definitely a big improvement in test runtime.

seebs:
rebased to pass CI

time results for `make test` on my laptop:
- no RAMDISK: test dies after 10 minutes because it runs too long
- RAMDISK: about 9 minutes
- TMPDIR: about 3 minutes

jaffee (changes requested):
I believe there are still references to RAMDISK in the CI config files which may be related to the test failure in the pipeline

jaffee (approved):

seebs:
So this is failing in CI, because CI's ramdisk is being mounted noexec, and if we set `$TMPDIR`, `go test` tries to put the test binaries there. Gonna ask to change the CI config and then retry that.

PR #2268, opened 2023-02-21
add name field to recordTime field

jrbrinlee1:
We weren't passing the name of the recordTime field when updated the IDK schema from the avro schema in the kafka consumer. This was preventing time quantums from being used. This PR corrects that mistake.


seebs:
Do we have a test for this that actually *uses* it, like, takes a thing from the avro schema in kafka, stores data, and verifies that the correct time quantums are getting populated?

Can we easily add one? I feel like that would be a really good thing to confirm/deny.

seebs (approved):
I'm willing to approve this, but if you can easily add a test that actually confirms that data submitted this way is showing up in the right time quantums, I would prefer that.

jrbrinlee1:
> I'm willing to approve this, but if you can easily add a test that actually confirms that data submitted this way is showing up in the right time quantums, I would prefer that.

@seebs I've added some additional testing. Let me know what you think.

seebs (approved):

PR #2269, opened 2023-02-21
Fix bad FLAGS env var passed when building to package.

j3m7:
This PR fixes a bad FLAGS value (-o featurebase) passed when building for the packaging target.


seebs (approved):
this seems like it probably fixes a problem i just hit, so yes

PR #2270, opened 2023-02-22
Limit memory usage on backuptar/restoretar

tgruben:


ctl/filebuffer.go
```diff
@@ -0,0 +1,116 @@
...
+	buf     bytes.Buffer
+	file    *os.File
+	tempDir string
+	reading bool
+	files   []*os.File
+	mu      sync.Mutex
+}
+
+func (fb *FileBuffer) Write(p []byte) (n int, err error) {
+	if fb.reading {
```
travisturner, line 26: I noticed a mutex covering some members of the FileBuffer. Is there no risk of concurrent access in the Write method?

ctl/filebuffer.go
```diff
@@ -0,0 +1,116 @@
+package ctl
```
travisturner, line 1: Because this one file (`FileBuffer`) seems so stand-alone and general purpose, it seems like it could be in its own package so that another packages could import and use it by calling something like `buffer.NewFileBuffer()` (or whatever its package name might be... `buffer` is probably not right. `files`?). It would also be nice to have its own unit tests.

ctl/filebuffer.go
```diff
@@ -0,0 +1,116 @@
...
+
+import (
+	"bytes"
+	"io"
+	"io/ioutil"
+	"os"
+	"sync"
+)
+
+func NewFileBuffer(max int, temp string) *FileBuffer {
```
travisturner, line 11: Include some godocs describing what the inputs mean.
```
// NewFileBuffer returns a file buffer which will use an in-memory buffer, until `max` bytes have been written, at which point it will write the contents of memory to a file, and continue writing future data to the file. The file will be written to `temp` directory.
```

travisturner (changes requested):
I tried to review this without much context. I think I understand the gist of it, but I can jump on a zoom if it would be easier to review in person. Your call.

ctl/filebuffer.go
```diff
@@ -0,0 +1,116 @@
...
+			return
+		}
+		_, err = io.Copy(fb.file, &fb.buf)
+		fb.buf.Reset()
+	}
+	return
+}
+
+func (fb *FileBuffer) Len() (int64, error) {
+	if fb.file != nil {
```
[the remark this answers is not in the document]
tgruben: no, its when the memory buffer is being used, return its lengh, but if the file is being used as the buffer use the files length

ctl/filebuffer.go
tgruben: not in how im using it, basically you do single process writing to fill it up and then launch potentially multiple readers

travisturner (approved):

PR #2272, opened 2023-02-22
Separate the featurebase and fbsql make build targets

travisturner:
Using the same build target was problematic because they shared the same flags. Since the `-o` output flag was used, the fbsql binary was overwriting the featurebase binary.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -127,6 +127,7 @@ build-fbsql:
 
 package:
 	GOOS=$(GOOS) GOARCH=$(GOARCH) $(MAKE) build
+	GOOS=$(GOOS) GOARCH=$(GOARCH) $(MAKE) build-fbsql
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -116,9 +116,12 @@ cover:
 cover-viz: cover
 	$(GO) tool cover -html=build/coverage.out
 
-# Compile Pilosa
+# Build featurebase
 build:
 	$(GO) build -tags='$(BUILD_TAGS)' -ldflags $(LDFLAGS) $(FLAGS) ./cmd/featurebase
+
```
j3m7, line 122: If separating to a separate build target please ensure the package target builds fbsql so it gets included in cloud environments

j3m7 (changes requested):

j3m7 (approved):

HHans09 (approved):

PR #2274, opened 2023-02-24
handle count(*) in `having` correctly

seebs:
After tracking through a whole bunch of unrelated code and coming up with a 40-line patch that mostly fixed this, I realized that the fact that it worked for count(field) meant something was making it work, added `countStarPlanExpression` to that case, and magically everything got better.

Writing one line of code: 30 seconds
Figuring out which one line of code to write: two days


seebs:
@tgruben, tagged you on review because Pat looked at the original second commit, figured out it was wrong, and fixed it, so now the second commit in particular needs someone *else* to look at it and verify that this makes sense.

jrbrinlee1 (approved):
Looks good to me

PR #2277, opened 2023-02-27
move closing of profiling stuff to Close method so it runs for durati…

jaffee:
…on of process

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -73,6 +73,8 @@ type Command struct {
 	// done will be closed when Command.Close() is called
 	done chan struct{}
 
+	traceCloser io.Closer
+
 	logOutput      io.Writer
 	queryLogOutput io.Writer
 	logger         loggerLogger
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -883,6 +882,14 @@ func (m *Command) Close() error {
 
 		err := eg.Wait()
 		_ = testhook.Closed(pilosa.NewAuditor(), m, nil)
+		if m.Config.DataDog.Enable {
+			defer profiler.Stop()
+		}
+		if m.traceCloser != nil {
+			defer m.traceCloser.Close()
```
tgruben, line 889: why defer?

jaffee: probably doesn't matter, but I figured it was best to stop the profiler when the Close actually completed... in case some part of the Close ended up hanging or something.

tgruben (approved):

PR #2280, opened 2023-02-27
fix backup tests so they're actually checking their checksums

seebs:
In fixing the issue where backups could get log messages inserted into their tar files, we segregated out "what you're logging errors to" from "what you're producing output to" in reasonably comprehensible ways, which worked for the backups, but the specific way we did it broke the checksum, so we were printing `hash:foo` messages to stdout, and comparing that empty strings were identical to each other in the actual tests.


JasonAllison (commented):
Looks good to me as long as NewChkSumCommand is satisfied

lorycloutier (approved):
Looks good to me.

PR #2282, opened 2023-02-28
implement select from time quantum columns (fb-1654)

paddyjok:
This PR enables range queries on timequantum fields.

- Introduces two new sql data types `idsetq` and `stringsetq`  
- Removes`timequantum` as a valid constraint on set types 
- Removes `subtable()` table valued function (and associated)
- Introduces rangeq() scalar function for time quantum filtering and optimizer logic to push these down into PQL call graphs


sql3/parser/parser_test.go
```diff
@@ -250,40 +250,6 @@ func TestParser_ParseTimeUnitConstraints(t *testing.T) {
 	})
 }
 
-func TestParser_ParseTimeQuantumConstraints(t *testing.T) {
```
seebs, line 253: So we took this test out, because there's no longer time quantum constraints. We are instead parsing these some other way. Where do we test the new thing's parsing?

paddyjok: It's implicit in the tests that exist - `timequantum` constraints still exist, but only on time quantum types, not set types.

sql3/planner/expressionpql.go
```diff
@@ -96,6 +96,85 @@ func (p *ExecutionPlanner) generatePQLCallFromExpr(ctx context.Context, expr typ
...
+
+			case *nullLiteralPlanExpression:
+				toValue = nil
+
+			default:
+				return nil, sql3.NewErrInternalf("unexpected argument type '%T'", expr.args[1])
+			}
+
+			if fromValue == nil && toValue == nil {
+				return nil, sql3.NewErrInternalf("replace me")
```
seebs, line 145: I think this wants to be replaced but I'm not sure what with. I assume the reason this is an Internalf is that we shouldn't be able to get here without having hit an earlier error, possibly in parsing?

paddyjok: Good catch - fixed in subsequent commit

sql3/parser/parser_test.go
paddyjok: Oh - actually, this is not correct. test reverted to work properly.

fhaynes (approved):
Bear in mind I'm not super familiar with the original code, but I didn't catch any errors. Also I assume docs/tests are coming.

PR #2283, opened 2023-03-01
fb-2013 remove sql endpoint feature flag.

vkrishnanfb:
Default config will have the sq.endpoint-enabled set to true. Server instances started with default config will accept SQL requests.


paddyjok (changes requested):
Let's remove the flag altogether...

paddyjok (approved):
lgtm

PR #2285, opened 2023-03-01
Use json.Number decoder to handle large ints in sql wire protocol

tgruben:
fixes fb-1916


wire_response.go
```diff
@@ -56,6 +73,7 @@ func (s *WireQueryResponse) UnmarshalJSONTyped(in []byte, typed bool) error {
 	if err := json.Unmarshal(in, &aux); err != nil {
 		return err
 	}
+	bypass, _ := s.decodeDataWithNumber(in)
```
seebs, line 76: I am distrustful of disregarding this error. It looks like we can get here if we get an error calling the decoder, or if we didn't get a []interface{}, and i don't know whether those can happen, but... I feel like we should handle it Just In Case.

sql3/sql_complex_test.go
```diff
@@ -2978,3 +2978,62 @@ func TestPlanner_BulkInsertParquet(t *testing.T) {
 		assert.Equal(t, row[1], row[2])
 	})
 }
+
+func TestPlanner_BulkInsert_FP1916(t *testing.T) {
+	c := test.MustRunCluster(t, 3)
+	defer c.Close()
+	node := c.GetNode(0).Server
+	_, _, _, err := sql_test.MustQueryRows(t, node, `create table greg-test (
```
seebs, line 2986: I have two suggestions for this:

1. Drop a bunch of these fields because we're not using them.
2. Make the number being converted more obviously defined in terms of the limit we're testing against, or possibly the range of int. Say, use `1<<62`.
3. Maybe also test at least one number that's out of even that expanded range and confirm that we actually handle it... however we expect to handle it?

... I recognize that's not actually two but I'm just gonna run with it.

wire_response.go
```diff
@@ -85,22 +103,38 @@ func (s *WireQueryResponse) UnmarshalJSONTyped(in []byte, typed bool) error {
...
 			switch hdr.BaseType {
 			case dax.BaseTypeID, dax.BaseTypeInt:
-				if _, ok := s.Data[i][j].(float64); ok {
-					s.Data[i][j] = int64(s.Data[i][j].(float64))
+				jn := bypass[i].([]interface{})[j] // forced to do this to handle ints bigger than 2^53
+				if v, ok := jn.(json.Number); ok {
+					if x, err := v.Int64(); err == nil {
+						s.Data[i][j] = x
+					} else {
+						return errors.Wrap(err, " can't be decoded as int64")
```
seebs, line 111: This seems to have an extra leading space (as does the other one below)?

seebs (approved):
I think we should probably remove the spaces from the messages, and maybe make sure we're testing the failure case, but this seems to be reasonably clearly correct.

wire_response.go
tgruben: it can happen in tests, but i'll clear up a bit

seebs (approved):

PR #2287, opened 2023-03-02
This changes the server sub-command to ignore unknown flag. Fixes FB-2019

fhaynes:
This uses the cobra library options to mark the sql3 flag as deprecated so the DB doesn't, ya know, explodey if someone tries to pass it in after we remove the flag.

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -47,6 +47,9 @@ func newServeCmd(stderr io.Writer) *cobra.Command {
 	serveCmd := &cobra.Command{
 		Use:   "server",
 		Short: "Run FeatureBase.",
+		FParseErrWhitelist: cobra.FParseErrWhitelist{
+			UnknownFlags: true,
+		},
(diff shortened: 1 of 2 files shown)


cmd/server.go
```diff
@@ -47,6 +47,9 @@ func newServeCmd(stderr io.Writer) *cobra.Command {
 	serveCmd := &cobra.Command{
 		Use:   "server",
 		Short: "Run FeatureBase.",
+		FParseErrWhitelist: cobra.FParseErrWhitelist{
```
paddyjok, line 50: We can probably remove this now that we handle deprecation....

paddyjok (changes requested):

cmd/server.go
fhaynes: Indeedly do, removing

paddyjok (approved):
Lgtm

fhaynes:
Merging

PR #2288, opened 2023-03-02
address various test irregularities and confusions

seebs:
This turns out to be two interacting weird edge case bugs, where one bug was actually causing a problem, and another bug was making it much harder to see what happened.

### DIFF
--- a/dax/controller/client/client.go
+++ b/dax/controller/client/client.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"time"
 
 	"github.com/featurebasedb/featurebase/v3/dax"
 	"github.com/featurebasedb/featurebase/v3/dax/computer"
@@ -27,8 +28,9 @@ var _ dax.Schemar = (*Client)(nil)
 // Client is an HTTP client that operates on the Controller endpoints exposed by
 // the main Controller service.
 type Client struct {
-	address dax.Address
-	logger  logger.Logger
+	address    dax.Address
+	httpClient *http.Client
+	logger     logger.Logger
 }
 
 // New returns a new instance of Client.
@@ -36,6 +38,9 @@ func New(address dax.Address, logger logger.Logger) *Client {
 	return &Client{
 		address: address,
 		logger:  logger,
+		httpClient: &http.Client{
+			Timeout: time.Second * 30,
+		},
 	}
 }
 
@@ -44,9 +49,10 @@ func New(address dax.Address, logger logger.Logger) *Client {
 func (c *Client) Health() bool {
 	url := fmt.Sprintf("%s/health", c.address.WithScheme(defaultScheme))
 
-	if resp, err := http.Get(url); err != nil {
+	if resp, err := c.httpClient.Get(url); err != nil {
 		return false
 	} else if resp.StatusCode != http.StatusOK {
+		defer resp.Body.Close()
 		return false
 	}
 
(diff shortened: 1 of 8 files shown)


dax/queryer/translator.go
```diff
@@ -26,15 +27,28 @@ func NewServerlessTranslator(controller dax.Controller) *serverlessTranslator {
 	}
 }
 
+var fbClientCache = map[dax.Address]*featurebase_client.Client{}
```
travisturner, line 30: the only thing that worries me about this is whether the lack of cache invalidation will bite us. If a new compute node registers with the controller later, with the same Address as a compute node that has died and been removed on the controller, would this client now be referencing the old, dead compute node?

travisturner (commented):
I'll need to look at this again when I'm less tired. And I'm on the road Friday, so not sure if I'll be able to get to it right away. I'm going to add @jaffee to the list of reviewers anyway because I'd like him to be aware of this.

idk/serverless/importer.go
```diff
@@ -33,19 +33,28 @@ func NewImporter(controller dax.Controller, qdbid dax.QualifiedDatabaseID, tbl *
...
+	if client != nil {
+		return client, nil
+	}
+	client, err := featurebaseclient.NewClient(address.HostPort(),
 		featurebaseclient.OptClientRetries(2),
 		featurebaseclient.OptClientTotalPoolSize(1000),
 		featurebaseclient.OptClientPoolSizePerRoute(400),
 		featurebaseclient.OptClientPathPrefix(address.Path()),
-		//featurebaseclient.OptClientStatsClient(m.stats),
+		//featurebase_client.OptClientStatsClient(m.stats),
```
jaffee, line 51: uh, why underscore?

dax/controller/client/client.go
```diff
@@ -27,15 +27,17 @@ var _ dax.Schemar = (*Client)(nil)
 // Client is an HTTP client that operates on the Controller endpoints exposed by
 // the main Controller service.
 type Client struct {
-	address dax.Address
-	logger  logger.Logger
+	address    dax.Address
+	httpClient http.Client
```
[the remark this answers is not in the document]
seebs: so, if i make it a `*http.Client`, i have to initialize it on first use, and have a mutex to govern the initialization and keep from racing. and since i'm just replacing DefaultClient, I assumed that the default settings, which are used by DefaultClient, were fine. I didn't want to change the settings without a concrete reason.

dax/queryer/translator.go
seebs: Isn't the Address just the URL?

A client that continues making requests to the URL should work, because if the new node came in with the same address, that URL now goes to the new node. I think.

idk/serverless/importer.go
seebs: So, if you look at the other copy of `fbClient`, it has underscores, and i cut and pasted it here and then edited it, so i didn't catch this.

dax/controller/client/client.go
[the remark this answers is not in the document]
seebs: hmm. on further thought, i don't actually think i find the argument persuasive, so i think i will just copy the querier/client behavior. creating a new `http.Client` object doesn't actually consume meaningful resources.

jaffee (approved):

dax/queryer/translator.go
travisturner: Yeah, it's just the URL. I just wasn't sure if `Client` held on to any information which may no longer apply if the underlying host changed. But you're probably right. This isn't a problem if the client's only state is the URL to which it connects.

dax/controller/client/client.go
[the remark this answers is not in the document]
travisturner: Either way is ok with me. I was just curious.

PR #2290, opened 2023-03-02
 fb-1915 Support large id's in NDJSON

tgruben:
uses json.Decoder to allow for large integer values in ndjson format in bulk import

### DIFF
--- a/sql3/planner/opbulkinsert.go
+++ b/sql3/planner/opbulkinsert.go
@@ -616,7 +616,7 @@ func (i *bulkInsertSourceNDJsonRowIter) Next(ctx context.Context) (types.Row, er
 				case *parser.DataTypeStringSet:
 					switch v := evalValue.(type) {
 					case json.Number:
-						result[idx] = v.String()
+						result[idx] = []string{v.String()}
 					case []interface{}:
 						setValue := make([]string, 0)
 						for _, i := range v {
--- a/sql3/planner/opbulkinsert.go
+++ b/sql3/planner/opbulkinsert.go
@@ -4,6 +4,7 @@ package planner
 
 import (
 	"bufio"
+	"bytes"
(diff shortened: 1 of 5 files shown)


sql3/sql_complex_test.go
```diff
@@ -3038,3 +3039,77 @@ func TestPlanner_BulkInsert_FP1916(t *testing.T) {
...
+	assert.NoError(t, err)
+	vprint.VV("results %#v", results)
+	if diff := cmp.Diff([][]interface{}{
+		{"8924809397503602651"},
+		{"8924809397503602652"},
+		{"8924809397503602653"},
+	}, results); diff != "" {
+		t.Fatal(diff)
+	}
+	/*
```
seebs, line 3110: i think we can drop this commented-out version, but apart from that this looks reasonable?

although i'm a bit confused because `id_col` is the input value, and then we're doing `select _id`, but i guess that's the ID? this may actually be the largest shard number we've ever tried to use.

sql3/planner/opbulkinsert.go
```diff
@@ -616,13 +615,8 @@ func (i *bulkInsertSourceNDJsonRowIter) Next(ctx context.Context) (types.Row, er
...
 					switch v := evalValue.(type) {
-					case float64:
-						if v == float64(int64(v)) {
-							result[idx] = []string{fmt.Sprintf("%d", int64(v))}
-						} else {
-							result[idx] = []string{fmt.Sprintf("%f", v)}
-						}
-
+					case json.Number:
+						result[idx] = v.String()
```
rachithrr, line 619: should this be string instead of []string{} ?

rachithrr, line 619: should this be string instead of []string{} ?

PR #2292, opened 2023-03-02
Remove pre_clone_script

gardavi:


.gitlab/.gitlab-ci.yml
```diff
@@ -28,11 +28,6 @@ variables:
   CI_IMAGE: "${CI_REGISTRY_IMAGE}/ci-builder:0.0.1"
   CI_PRE_CLONE_SCRIPT: |
     set -x
-    if [ -d "$CI_PROJECT_DIR" ]; then
-      cd "$CI_PROJECT_DIR"
-      find . -type f -exec chmod 0644 {} + 
-      find . -type d -exec chmod 0755 {} + 
-    fi
```
gardavi, line 35: Before, this script was basically running on the runner itself and just silently failing because our jobs that run in docker (the shell executor ones) create files owned by `0:0` (`root`) which the `find -exec` fails on.

Turns out, we don't need to do any of the find stuff here since they get wiped by a `git clean -ffdx` that happens later. I'm not sure why that command can delete these coverage files that we leave behind that are owned by root, but I'm not questioning it.

jrbrinlee1 (approved):

PR #2293, opened 2023-03-03
Add datetimename function to SQL3

lorycloutier:
FB-1896
Added the datetimename function, which returns parts of a timestamp as strings. Month and day of the week are named ("January", "Monday") while others are returned as a string of digits ("2023"). Added tests to the test definitions.


sql3/test/defs/defs_date_functions.go
```diff
@@ -241,5 +241,44 @@ var datePartTests = TableTest{
 			),
 			Compare: CompareExactUnordered,
 		},
+		{
```
vkrishnanfb, line 244: This test definition need to be organized by function. Earlier I mistakenly added toTimeStamp() tests to dataPart() tests. I am separating them as part of the dateTimeAdd() work. The broken down definitions would look like:

`var datePartTests = TableTest{......}

// toTimestamp tests
var toTimestampTests = TableTest{......}

// datetimeAdd tests
var datetimeAddTests = TableTest{......}

// datetimeName tests
var datetimeNameTests = TableTest{......}`

lorycloutier: Oh, good catch, I forgot to change the function being called on those. Not sure how that passed when I ran the tests.

sql3/planner/inbuiltfunctionsdate.go
```diff
@@ -196,3 +220,82 @@ func (n *callPlanExpression) EvaluateToTimestamp(currentRow []interface{}) (inte
...
+		return nil, err
+	}
+
+	dateEval, err := n.args[1].Evaluate(currentRow)
+	if err != nil {
+		return nil, err
+	}
+
+	// nil if anything is nil
+	if intervalEval == nil || dateEval == nil {
```
seebs, line 236: Can/should this happen? I don't fully understand this, but I guess I'm wondering whether either of those should be an error, rather than a non-error nil return.

sql3/test/defs/defs_date_functions.go
```diff
@@ -241,5 +241,49 @@ var datePartTests = TableTest{
 			),
 			Compare: CompareExactUnordered,
 		},
+		// dateTimeName tests
+		// make sure returning a string as the year still works
+		// if this works then the other parts converted to strings of digits should also work
+		{
+			SQLs: sqls(
+				"select _id, datetimename('yy', ts) from dateparttests",
```
seebs, line 249: I am distrustful of `yy` as a name here. Do we have a spec for what the range names/types should be? Is there a reason `y` is duplicated? In particular, if I were going to see `yy`, I'd expect the result to be `12`, with `yyyy` producing `2012`.

sql3/planner/inbuiltfunctionsdate.go
seebs: Reading it more: I think if intervalEval is nil, that should be an error, probably. I think it may make sense to allow nulls and assert that DateTimeName('m', NULL) is NULL, because that's sort of how SQL rolls with nulls, but I think DateTimeName(NULL, date) should be an error, probably?

sql3/test/defs/defs_date_functions.go
```diff
@@ -241,5 +241,49 @@ var datePartTests = TableTest{
...
+		{
+			SQLs: sqls(
+				"select _id, datetimename('w', ts) from dateparttests",
+			),
+			ExpHdrs: hdrs(
+				hdr("_id", fldTypeID),
+				hdr("", fldTypeInt),
+			),
+			ExpRows: rows(
+				row(int64(1), "Thursday"),
```
seebs, line 284: I don't think this can possibly be correct, because you have a header saying fldTypeInt, and a string "Thursday".

I'm inferring that we maybe don't enforce the header column types, and maybe we should, but we should definitely not be claming that "Thursday" is an integer. Thursday is a man [Chesterton, 1908], and a man is not a number [McGoohan, 1967].

seebs (dismissed):
i don't like the fldTypeInt/"Thursday" mismatch.

sql3/planner/inbuiltfunctionsdate.go
```diff
@@ -196,3 +220,82 @@ func (n *callPlanExpression) EvaluateToTimestamp(currentRow []interface{}) (inte
...
+	}
+
+	switch strings.ToUpper(interval) {
+	case intervalYear:
+		return fmt.Sprint(date.Year()), nil
+
+	case intervalYearDay:
+		return fmt.Sprint(date.YearDay()), nil
+
+	case intervalMonth:
```
seebs, line 269: So, the test case below expects "November" for month, apparently. What if I wanted "11"?

sql3/test/defs/defs_date_functions.go
lorycloutier: This is working off the constants provided in inbuiltfunctionsdate.go - they're inconsistent and cleaning them up would probably be a good idea, but I think that's a separate ticket since it'll touch a whole bunch of different stuff that's using them.

sql3/planner/inbuiltfunctionsdate.go
lorycloutier: Numerical values for months and weekdays are returned by datepart, soon to be datetimepart, rather than by datetimename.

sql3/planner/inbuiltfunctionsdate.go
lorycloutier: `fbsql> select datepart('m', '2022-10-11T12:31:45.234Z');

----
 10

fbsql> select datepart(NULL, '2022-10-11T12:31:45.234Z');

------
 NULL

fbsql> select datepart('m', NULL);

------
 NULL`
 it looks like it can happen. I think it's debatable whether or not it should be an error? it at least seems consistent to me that if you try to get a NULL part of a timestamp you get NULL back, but possibly an error would be more useful?

sql3/planner/inbuiltfunctionsdate.go
```diff
@@ -196,3 +220,82 @@ func (n *callPlanExpression) EvaluateToTimestamp(currentRow []interface{}) (inte
...
+		return fmt.Sprint(date.Hour()), nil
+
+	case intervalMinute:
+		return fmt.Sprint(date.Minute()), nil
+
+	case intervalSecond:
+		return fmt.Sprint(date.Second()), nil
+
+	case intervalMillisecond:
+		return fmt.Sprint(date.Nanosecond() * 1000 * 1000), nil
```
vkrishnanfb, line 327: time.Nanosecond() represents the second offset in nano seconds, to extract milli, micro and nano seconds from this value the calculation needs to as follows, it was coded defective in the existing datepart() and I fixed it as part of the work I did for datetimeAdd() :
	case intervalMillisecond:
		return int64(date.Nanosecond() / 1000000), nil
	case intervalMicrosecond:
		return int64((date.Nanosecond() % 1000000) / 1000), nil
	case intervalNanosecond:
		return int64(date.Nanosecond() % 1000), nil

lorycloutier: I think that code answers the wrong question though? like if the time is 11:22:33.444555666, and the query asks for nanoseconds, I think datepart and datetimename are supposed to return the antire fractional seconds part, expressed as nanoseconds, so it'd want to be 444555666, not just 666?

lorycloutier: I think I've got it sorted out to what it probably should be now. MS gives 444, US gives 444555, and NS gives 444555666. Good catch on the part where MS needed to be / instead of * though, that went right by me.

vkrishnanfb (approved):
Looks good to me.

PR #2299, opened 2023-03-07
Minor cleanup to some fbsql flags and meta-commands

travisturner:
This also stubs out an integration test for fbsql.

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -366,6 +366,12 @@ func (cmd *Command) setupConfig() error {
 func (cmd *Command) executeAndWriteQuery(qry query) error {
 	queryResponse, err := cmd.executeQuery(qry)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


cli/cli_integration_test.go
```diff
@@ -0,0 +1,246 @@
...
+				c.outline <- line
+				line = []byte{}
+				continue
+			}
+			line = append(line, b)
+		}
+	}()
+}
+
+type comparitor string
```
jaffee, line 173:
```suggestion
type comparator string
```

jaffee (approved):

cli/cli_integration_test.go
travisturner: I'm gonna back out that last commit anyway. It was just stubbing out a test framework, but I need to fix some things on it before it's ready.

PR #2302, opened 2023-03-07
Add platform specification when building binaries

andreacappelletti97:


travisturner (commented):
Since we're pinning these to a specific platform, I think we should add comments to these make targets indicating that they are intended to be used for linux\amd64 specific environments, such as linux instances in AWS ECS.

Because this code is open source, these make targets are also available to anyone using featurebase, so we should make it clear that these particular make targets are not meant to be used for general builds.

Also, while you're at it, can you fix the tab/space indentation differences?

travisturner (approved):

PR #2303, opened 2023-03-07
Fb 1897

lorycloutier:
Change datepart function to datetimepart 
FB-1897 
The specs for the datetimepart function are, as far as i can tell, identical to the existing datepart function. Per Pat, replaced the datepart function with datetimepart rather than just adding datetimepart as an alias. Made sure existing tests that were using datepart got switched over.

### DIFF
--- a/sql3/test/defs/defs_date_functions.go
+++ b/sql3/test/defs/defs_date_functions.go
@@ -5,11 +5,11 @@ import (
 	"time"
 )
 
-// datepart tests
-var datePartTests = TableTest{
+// datetimepart tests
+var dateTimePartTests = TableTest{
 
 	Table: tbl(
-		"dateparttests",
+		"datetimeparttests",
 		srcHdrs(
 			srcHdr("_id", fldTypeID),
 			srcHdr("a", fldTypeInt, "min 0", "max 1000"),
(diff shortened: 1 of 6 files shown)


sql3/test/defs/defs_date_functions.go
```diff
@@ -24,21 +24,21 @@ var datePartTests = TableTest{
 		{
 			name: "DatePartIncorrectParamsCount",
```
rachithrr, line 25: Can you change the names as well.

rachithrr (commented):
I see fbsql binary pushed here, but i have a feeling that it shouldn't be. 

Looks good otherwise.

travisturner (dismissed):
Please remove the `fbsql` binary from this PR.

Also, there is a make target `make install-fbsql` which will install fbsql into your go/bin (which you should have as part of your path, allowing you to just run `fbsql` at a comand line). You probably shouldn't be building fbsql into the featurebase root directory.

sql3/test/defs/defs_date_functions.go
lorycloutier: Fixed. I think the names got added in a different ticket; they weren't there when I was doing the original work.

lorycloutier:
Fixed, that was an accident and I had to look up how to remove it.

rachithrr (approved):

PR #2307, opened 2023-03-08
FB-1894: Implement DateTimeDiff()

rachithrr:


sql3/planner/inbuiltfunctionsdate.go
```diff
@@ -247,13 +271,58 @@ func (n *callPlanExpression) EvaluateDateTimeFromParts(currentRow []interface{})
...
+
 	dt := time.Date(timestamps[0], time.Month(timestamps[1]), timestamps[2], timestamps[3], timestamps[4], timestamps[5], timestamps[6]*1000*1000, time.UTC)
 	if dt.Year() < 0 || dt.Year() > 9999 {
-		return nil, sql3.NewErrYearOutOfRange(0, 0, dt.Year())
+		return nil, sql3.NewErrInvalidDatetimePart(0, 0, dt.Year())
 	}
 	return dt, nil
 }
 
+func isValidDateTimeParts(timestamps []int) (value int, ok bool) {
```
travisturner, line 285: I'm not really going too deep on these code reviews.... I just periodically skip through things and make a note on anything that catches my eye. So take this or ignore it.

This function, on its own, is kind of a mystery. I can't really tell you what it's supposed to be doing. Because of that, I'd say it's a good candidate for go docs. The idea is to help out the developer that comes behind you (even if that developer is you 6 weeks from now).

What is the data in `timestamps`? What is the `int` value that's returned. When is it `ok`? It's not obvious, even looking at the code why I would use this function.

travisturner (commented):
Just a comment about helping out future us.

lorycloutier (dismissed):
Looks good

sql3/planner/inbuiltfunctionsdate.go
rachithrr: Thanks for checking this out. 
Yes, you are right. Normally, i'd write some comments for exported functions, but this makes sense. I'll update it.

lorycloutier (approved):

PR #2308, opened 2023-03-09
fbsql integration test framework

travisturner:
Introduce fbsql integration test framework.

This also adds support for `pset location` to set the geo location (i.e.
time zone) in which timestamps should be displayed. And it adds support
for comments (lines starting with `--`) in the line reader/splitter.

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -56,9 +56,9 @@ type Command struct {
 
 	Queryer Queryer `json:"-"`
 
-	Stdin  io.ReadCloser `json:"-"`
+	stdin  io.ReadCloser `json:"-"`
 	stdout io.Writer     `json:"-"`
-	Stderr io.Writer     `json:"-"`
+	stderr io.Writer     `json:"-"`
 
 	// output is where actual results are written. This might point to stdout,
 	// or to a file, based on the current configuration.
@@ -116,9 +116,9 @@ func NewCommand(logdest logger.Logger) *Command {
 		splitter:   newSplitter(newReplacer(variables)),
 		workingDir: newWorkingDir(),
 
-		Stdin:  Stdin,
+		stdin:  Stdin,
 		stdout: Stdout,
-		Stderr: Stderr,
+		stderr: Stderr,
 
 		output:       Stdout,
 		writeOptions: defaultWriteOptions(),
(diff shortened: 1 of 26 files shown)


cli/cli.go
```diff
@@ -117,7 +117,7 @@ func NewCommand(logdest logger.Logger) *Command {
 		workingDir: newWorkingDir(),
 
 		Stdin:  Stdin,
-		Stdout: Stdout,
+		stdout: Stdout,
```
adrwalker, line 120: Why does `stdout` have to be lowercase?

travisturner: I needed to add a setter method for `stdout`, because it has to do more than just set the value. We don't currently need that for `Stdin` or `Stderr` so I didn't bother changing those.

I suppose if we want to be consistent/proactive, we could make setter methods for those as well. I can do that.

travisturner: @adrwalker I pushed a commit which adds setters for those.

cli/cli_integration_test.go
```diff
@@ -0,0 +1,255 @@
...
+func TestCLIIntegration(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping integration test")
+	}
+
+	ctx := context.Background()
+
+	t.Run("Stubbed Framework", func(t *testing.T) {
+		mc := test.MustRunManagedCommand(t)
+		defer mc.Close()
```
JasonAllison, line 27: Out of curiosity, what is the reasoning behind the defer .Close() here?

travisturner: `test.MustRunManagedCommand()` starts up an serverless instance of FeatureBase. The deferred `Close()` just ensures that that running process is closed cleanly and releases any resources it's holding onto. It would probably get cleaned up either way, but this is good practice. One thing we add is a wrapper around the defer which checks the error on `Close()` and fails the test if it doesn't close cleanly.

cli/cli_integration_test.go
```diff
@@ -0,0 +1,255 @@
...
+	case compHasPrefix:
+		require.True(c.t, strings.HasPrefix(string(outline), line), msg...)
+	case compWithFormat:
+		require.True(c.t, compareByteSlices(outline, []byte(line)), msg...)
+	default:
+		c.t.Fatalf("invalid comparator: %s", comp)
+	}
+}
+
+func (c *comparer) Write(b []byte) (n int, err error) {
```
seebs, line 223: i'm sorta concerned about the performance implications of channel overhead per-byte. we might want to revisit this later if these tests end up being expensive. but it's probably fine for now as an initial implementation.

seebs (approved):
this looks reasonable except maybe for the channel-op-per-byte overhead, but that's fixable if it becomes a problem.

cli/cli_integration_test.go
travisturner: yeah, I agree with you. I think the way we compare the SEND/EXPECT lines in the test will evolve.

PR #2313, opened 2023-03-10
Fix failing selects on views defined with date literals

vkrishnanfb:
This PR fixes defect FB-2065 failing select queries on views defined referencing built-in literals (CURRENT_DATE, CURRENT_TIMESTAMP) or string date literals('9999-12-12T00:00:00Z').


sql3/parser/ast.go
```diff
@@ -1919,6 +1919,7 @@ func (lit *BoolLit) String() string {
 type DateLit struct {
 	ValuePos Pos       // literal position
 	Value    time.Time // literal value
+	Token    Token     // token found in the sql
```
travisturner, line 1922: I would add a comment here about how Token is meant to be handled.

```suggestion
	ValuePos Pos       // literal position
	Value    time.Time // literal value

	// Token will be a non-zero value if the date found in the SQL is a built-in
	// literal. This is maintained here in the struct for cases, such as SQL
	// stored for a view, where the SQL string is to be parsed again.
	Token Token
```
or something like that. Maybe I'm not understanding it correctly, but you can describe it however it makes sense.

travisturner (commented):
This all makes sense to me. I just left a few comments about comments.

vkrishnanfb:
@paddyjok This PR is ready for review, the changes you have requested are implemented.

sql3/planner/expression.go
```diff
@@ -1998,6 +1998,48 @@ func (n *boolLiteralPlanExpression) WithChildren(children ...types.PlanExpressio
 	return n, nil
 }
 
+// represents date system variable such as CURRENT_DATE and CURRENT_DATETIME
+type dateSysVariablePlanExpression struct {
```
paddyjok, line 2002: Do we need a variable expression type this specific? We could have a single variable expression type and then switch on name...

sql3/planner/expression.go
```diff
@@ -1998,6 +1998,48 @@ func (n *boolLiteralPlanExpression) WithChildren(children ...types.PlanExpressio
...
+}
+
+func newDateSysVariablePlanExpression(name string, value time.Time) *dateSysVariablePlanExpression {
+	return &dateSysVariablePlanExpression{
+		name:  name,
+		value: value,
+	}
+}
+
+func (n *dateSysVariablePlanExpression) Evaluate(currentRow []interface{}) (interface{}, error) {
```
paddyjok, line 2014: Why pre-load the value somewhere else? Why not just evaluate here? 

Also consider a future where compilation and execution are separated in time - e.g. create view compiles and stores a binary representation of an execution plan which gets executed some time later.

sql3/planner/expression.go
```diff
@@ -2611,6 +2653,14 @@ func (p *ExecutionPlanner) compileExpr(expr parser.Expr) (_ types.PlanExpression
 	case *parser.DateLit:
 		return newDateLiteralPlanExpression(expr.Value), nil
 
+	case *parser.SysVariable:
+		switch expr.Token {
```
paddyjok, line 2657: See comment above on deferring evaluation until execute time.

sql3/planner/expressionanalyzer.go
```diff
@@ -151,6 +152,15 @@ func (p *ExecutionPlanner) analyzeExpression(ctx context.Context, expr parser.Ex
...
 
+	case *parser.SysVariable:
+		// retrieve the system variables store from the context
+		sysvar := ctx.Value(types.SysVarCtxKey).(*types.SysVars)
+		// from the store get the value for the referenced system variable
+		// using a shared store allows us to assign same value for all references of a system variable, especially CURRENT_TIMESTAMP enjoys a consistent value.
+		dt := sysvar.GetSystemVariable(e.Token)
+		e.Value = dt
+		return e, nil
+
```
paddyjok, line 163: I looked around - it turns out no-one does a single value for 'current_time' variables that's consistent for the request/transaction - you get what you get when you execute it each time. So I don't think this is a requirement.

vkrishnanfb: Will work on deferring evaluation to execution and generalizing the variable expression type. 

Evaluating all date time scalar references simultaneously (to same value) within the context of a single statements execution is what ANSI sql specification prescribes (see spec text extracted). I know oracle, postgres adhere to this, not sure about others. Here is an example of the effect:

`insert into demo(id, time_created, last_update) select id, current_timestamp, current_timestamp from ten_million_rows_table;`

All the 10 million rows will have the same timestamp value for time_created and last_update regardless of the time first row is inserted and the last. 

On the other hand, the following block will assign each row a different time:

`declare cursor x for select id from ten_million_rows_table

open cursor x
loop
   fetch x into _id
   insert into demo(id, time_created, last_update) values(_id, current_timestamp, current_timestamp)
end loop
`

Following is the ANSI specification, rule #3 talks about simultaneous evaluation:

        6.8  <datetime value function>

         Function

         Specify a function yielding a value of type datetime.

         <current date value function> ::= CURRENT_DATE

         <current time value function> ::=
                CURRENT_TIME [ <left paren> <time precision> <right paren> ]

         <current timestamp value function> ::=
                CURRENT_TIMESTAMP [ <left paren> <timestamp precision> <right paren> ]

         General Rules

         1) The <datetime value function>s CURRENT_DATE, CURRENT_TIME, and
            CURRENT_TIMESTAMP respectively return the current date, current
            time, and current timestamp; the time and timestamp values are
            returned with time zone displacement equal to the current time
            zone displacement of the SQL-session.

         2) If specified, <time precision> and <timestamp precision> respec-
            tively determine the precision of the time or timestamp value
            returned.

         3) If an SQL-statement generally contains more than one reference
            to one or more <datetime value function>s, then all such ref-
            erences are effectively evaluated simultaneously. The time of
            evaluation of the <datetime value function> during the execution
            of the SQL-statement is implementation-dependent.

paddyjok: Not saying we shouldn't do simultaneous eval - saying it's going to be complex to get right in a distributed env, and asking if we really need to take the time to do it now. I'd argue that we can defer and move on - we have a bunch of other stuff that is probably more impactful in our ability to be successful.

vkrishnanfb:
All feedback comments are addressed. Evaluation is deferred until execution, generalized the expression type and removed sysVariable execution context store. This PR is ready for review.

paddyjok (approved):
lgtm

PR #2314, opened 2023-03-10
Linters!

travisturner:


batch/batch_test.go
```diff
@@ -2045,7 +2045,7 @@ func mutexClearRegression(t *testing.T, importer featurebase.Importer, sapi feat
 	}
 
 	col := uint64(0)
-	row := uint64(1)
+	row := uint64(0)
```
seebs, line 2048: I'm slightly surprised by this change, but on further study, I assume `ineffassign` is allowing ineffective initializations to zero as expressing the type of a newly-created variable, but if you set it to one and then overwrite it before using it, that seems suspicious.

batch/egpool/egpool.go
```diff
@@ -77,7 +77,7 @@ func (eg *Group) processJobs() {
 	defer func() {
 		if !finished {
 			if p := recover(); p != nil {
-				eg.err(ErrPanic{p})
+				eg.err(PanicError{p})
 			} else {
 				eg.err(ErrGoexit)
```
seebs, line 82: Should this name also be changed? I don't really know what's being changed here. Or is `ErrGoexit` a value, rather than a type?

sql3/planner/opquery.go
```diff
@@ -130,10 +131,13 @@ func (i *queryIterator) Next(ctx context.Context) (types.Row, error) {
 		// either error or no more rows, either way update the request
 		requestId, ok := fbcontext.RequestID(ctx)
 		if !ok {
-			return nil, sql3.NewErrInternalf("unable to get request id from context")
+			return nil, errors.Wrapf(sql3.NewErrInternalf("unable to get request id from context"), "next on child: %s", err)
```
seebs, line 134: I'm not understanding this but I'm a bit confused by it, because it looks like we're creating a NewErrInternal, but we're also wrapping an existing err, and I'm just a bit unclear on the intended output message.

http_handler.go
```diff
@@ -3752,16 +3752,16 @@ func (h *Handler) handleReserveIDs(w http.ResponseWriter, r *http.Request) {
 
 	ids, err := h.api.ReserveIDs(req.Key, req.Session, req.Offset, req.Count)
 	if err != nil {
-		var esync ErrIDOffsetDesync
+		var esync IDOffsetDesyncError
```
tgruben, line 3755: doesn't the pattern ErrIDOffsetDesync match the others in this PR?

tgruben (changes requested):

sql3/planner/opquery.go
seebs: ... Also I'm not sure whether "err" is valid here, because it doesn't look like `err` was returned by the thing that we got the !ok from, so this seems like it could be an unrelated error?

seebs (approved):
Approved-ish, I actually am not sure about that errors.Wrapf but everything else looks good.

batch/egpool/egpool.go
travisturner: yes, `ErrGoexit` is a variable, which are supposed to start with `Err`

sql3/planner/opquery.go
travisturner: I was a little confused by this logic as well, but what I noticed while fixing the linter complaint was that if we were unable to get the request id from the context, then we were losing the original error... so this was my attempt to include both errors in the return.

http_handler.go
travisturner: so an error variable should be `Err---`
an error type shoud be `---Error`

tgruben (approved):

PR #2316, opened 2023-03-10
handle empty avg agg

jrbrinlee1:
AVG() returns null (as opposed to a very large magnitude value) when no records are returned by the WHERE filter.


paddyjok (approved):
lgtm

PR #2317, opened 2023-03-12
Enable linter: stylecheck

travisturner:
This enables the stylecheck linter, but excludes some staticchecks for now. The following are ignored because they will take a bit of time to address, but the intention is to address them and remove them from the exclusion list.

ST1000: at least one file in a package should have a package comment
ST1003: golang naming standards
ST1016: methods on the same type should have the same receiver name
ST1020: comment on exported function

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -28,6 +28,7 @@ linters:
     - prealloc
     # - predeclared (20 to fix)
     # - stylecheck (quite a lot to fix, but we should definitely work on this)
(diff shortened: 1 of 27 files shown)


jaffee (dismissed):

travisturner:
@jaffee this can use another approval.
@jrbrinlee1 and I talked about the test and we decided to remove it and revisit how that test works.

jaffee (approved):

PR #2320, opened 2023-03-15
SQL3 tests newMessageError and all the constructors

tgruben:


sql3/planner/wireprotocol_test.go
```diff
@@ -0,0 +1,174 @@
...
+				err:   errors.New("ABC"),
+			},
+			wantErr: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := newMessageError(tt.args.p, tt.args.reader)
+			if (err != nil) != tt.wantErr {
```
travisturner, line 54: can we convert these to use the `testify` package?

wireprotocol/wireprimitives.go
```diff
@@ -117,6 +117,8 @@ func WriteSchema(schema types.Schema) ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
+// Note this expects the token to be consumed prior to execution
```
travisturner, line 120: comments should star with `// ReadSchema ...`
also, I don't quite understand the comment.

travisturner (commented):
couple questions. mainly just suggesting we should use testify in our tests.

wireprotocol/wireprimitives.go
tgruben: i simply produced a buffer with WriteSchema and then consumed that same buffer with ReadSchema and it failed and I was just attempting to communicate that is not expected behavior

rachithrr (approved):
lgtm

PR #2335, opened 2023-03-21
Add HasDirective to dax.Node struct to force Directive on restart

travisturner:
For on-prem serverless, if we restart the process containing the controller and computer(s), when they come back up, the controller doesn't know that the computers have been restarted, so it doesn't send them a directive. This change forces the controller to send a directive upon startup by a computer.


travisturner:
@jaffee we can go over this via zoom if you want. 
I think i want to change `HasDirective` to `ForceDirective` and invert the boolean anyway.

jaffee (approved):

PR #2337, opened 2023-03-21
SQL test backfilling

seebs:
This is actually two related sets of test cases in the parser/scanner stuff. (tickets 2073, 2074)

There's a couple of things in here which may want more careful attention. One is the previously-commented-out test case for a double inner join. It was commented out because stringizing the resulting data structure produced wildly different output, which broke the basic sanity checks. After studying this a bit, I've concluded that the code to "rewrite last source to nest next join on right side" logic was, in fact, precisely backwards -- we simply didn't want that code at all. Taking it out and expecting a data structure that makes sense also produces a data structure which stringizes back to the original query.

We add support for StringLit to know whether they came from blobs or not, so they can stringize to blobs again, so the round-trip testing works with them.

We add a test case that tests tuple assignment. I note that tuple assignment looks weird -- you have to use {} for the values, but () for the column names.

There's a number of still-unreached cases, but most of them are things where we check a token type with peek to decide to call a function, then that function asserts that the next token is that type. This shouldn't be capable of failing, but it's not a bad idea to have the extra checking there, I think?

### DIFF
--- a/sql3/parser/ast.go
+++ b/sql3/parser/ast.go
@@ -299,6 +299,8 @@ func CloneExpr(expr Expr) Expr {
 		return expr.Clone()
 	case *StringLit:
 		return expr.Clone()
+	case *TupleLiteralExpr:
+		return expr.Clone()
 	case *UnaryExpr:
 		return expr.Clone()
 	case *Variable:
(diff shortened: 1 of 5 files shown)


paddyjok (approved):
Yep - I think this is good.

PR #2338, opened 2023-03-21
fbsql disconnect from database with `\c -`

travisturner:
This adds the ability to disconnect from the current database by passing a hyphen to the `\c` meta-command.

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -487,7 +487,7 @@ func (cmd *Command) connectToDatabase(dbName string) error {
 		p = newNopPrinter()
 	}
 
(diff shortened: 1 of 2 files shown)


cli/cli.go
```diff
@@ -487,7 +487,12 @@ func (cmd *Command) connectToDatabase(dbName string) error {
 		p = newNopPrinter()
 	}
 
-	if dbName == "" {
+	// Providing a blank ("") or hyphen ("-") dnName is the equivalent of
```
jaffee, line 490:
```suggestion
	// Providing a blank ("") or hyphen ("-") dbName is the equivalent of
```

jaffee (approved):

PR #2341, opened 2023-03-22
Sqldb rip boltdb

jaffee:
this also includes the environment var changes (see the first commit if you want to review those separately)


travisturner (approved):
So much red! I love it.

travisturner (approved):

PR #2345, opened 2023-03-22
Add tests cases for coverage in expressiontypes.go

lorycloutier:
FB-2041
Added tests for typeIsTimeQuantum and typeIsSet and made them pass. 
Added tests for DataTypeTuple cases in typesAreAssignmentCompatible. 
Timestamp conversion checking is handled before it gets to that point but I left those branches in as a backstop.
Checking to see if DataType[String,ID]SetQuantum can be assigned to themselves doesn't appear to be reachable currently but left those branches in, because something may use them in future.

### DIFF
--- a/sql3/test/defs/defs_set_functions.go
+++ b/sql3/test/defs/defs_set_functions.go
@@ -304,3 +304,93 @@ var setParameterTests = TableTest{
 		},
 	},
 }
+
+// test IDSetQ and StringSetQ functions
+var setTimeQuantumTests = TableTest{
+	name: "selectwithsetqliterals",
+	Table: tbl(
+		"selectwithsetqliterals",
+		srcHdrs(
+			srcHdr("_id", fldTypeID),
+			srcHdr("a", fldTypeInt, "min 0", "max 1000"),
+			srcHdr("b", fldTypeInt, "min 0", "max 1000"),
+			srcHdr("ssq1", fldTypeStringSetQ, "timequantum 'YMD'"),
+			srcHdr("isq1", fldTypeIDSetQ, "timequantum 'YMD'"),
+		),
+	),
+	SQLTests: []SQLTest{
+		// can't find example syntax or docs for how to put time quantum fields in with srcRow()
+		// so i'm inserting them with SQL statements.
+		{
+			SQLs: sqls(
+				"insert into selectwithsetqliterals(_id, a, b, ssq1, isq1) values (1, 10, 100, {'2022-01-03T00:00:00Z', ['foo']}, {'2022-01-01T00:00:00Z', [99, 101]})",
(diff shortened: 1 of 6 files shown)


sql3/test/defs/defs_set_functions.go
```diff
@@ -304,3 +304,93 @@ var setParameterTests = TableTest{
...
+			srcHdr("_id", fldTypeID),
+			srcHdr("a", fldTypeInt, "min 0", "max 1000"),
+			srcHdr("b", fldTypeInt, "min 0", "max 1000"),
+			srcHdr("ssq1", fldTypeStringSetQ, "timequantum 'YMD'"),
+			srcHdr("isq1", fldTypeIDSetQ, "timequantum 'YMD'"),
+		),
+	),
+	SQLTests: []SQLTest{
+		// can't find example syntax or docs for how to put time quantum fields in with srcRow()
+		// so i'm inserting them with SQL statements.
```
vkrishnanfb, line 323: Appreciate this! In some situations I think using SQL to setup test data for a SQL test is a better approach. Once I spent considerable time scratching my head due to a faulty insert generated by the test framework.

sql3/test/defs/defs_timequantum.go
```diff
@@ -99,6 +141,7 @@ var timeQuantumTest = TableTest{
 			ExpErr: "calling ranqeq() usage invalid",
 		},
 		{
+			name: "stringset-rangeq",
```
vkrishnanfb, line 144: At line 129 the expected error message has a typo. It's missing 'c' in 'calling'.

sql3/test/defs/defs_set_functions.go
travisturner: I'd prefer we update the test framework to handle time quantum fields, rather than start doing inserts in the result tests.

sql3/test/defs/defs_timequantum.go
```diff
@@ -111,6 +154,20 @@ var timeQuantumTest = TableTest{
 			),
 			Compare: CompareExactUnordered,
 		},
+		{
```
vkrishnanfb, line 157: This comment is for the StringSetQ query test. Based on my understanding of where `rangeq(a.ss1, '2022-01-02T00:00:00Z', null)` the result should include `'test4'` and `'test5'`, because the timestamp on those records are on or after (>=) the `from` parameter value `'2022-01-02'`. I may be misunderstanding how rangeq() works.

sql3/test/defs/defs_set_functions.go
lorycloutier: Probably a good idea at some point, but at a quick look, I think it'd be enough work to require its own ticket.

vkrishnanfb (approved):

PR #2349, opened 2023-03-24
Force rebuilding idk pilosa images

gardavi:


seebs (commented):
This is a quasi-approval: This looks right to me, but I want to have the test for the problem this fixes ready to go before merging it, and may want to just merge this patch into that branch and commit them together.

seebs:
I've stolen this commit for pull 2350, which puts it on top of the addition of a test which fails without it, in the hopes of being more confident this actually fixes the problem.

gardavi:
Doesn't need to stick around :)

PR #2350, opened 2023-03-24
make IDK check featurebase version

seebs:
After a weird CI failure that suddenly went away, I have theorized that we're getting cached versions of the featurebase docker image in CI when we don't intend to, and thus, running IDK tests against *the wrong version of featurebase*. This checks for that problem.

### DIFF
--- a/.gitlab/.gitlab-ci.yml
+++ b/.gitlab/.gitlab-ci.yml
@@ -353,7 +353,7 @@ run go tests idk race:
     - cd ./idk/
     - echo $PROJECT
     - echo $CI_COMMIT_REF_SLUG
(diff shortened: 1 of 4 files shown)


seebs:
So `go tests race` failed in `TestCLIIntegration/Stubbed_Framework/table`, saying "race detected during execution of test", but it didn't actually dump the usual RACE DETECTED message showing what was being written or read. I don't know why, but I'm willing to try rerunning it and hope it goes away since it isn't in any of the affected code, but I'm creating a ticket for it.

seebs:
nevermind the actual output is indeed there. still ignoring it for this PR but making a ticket for it.

PR #2356, opened 2023-03-30
cli kafka runner updates

jrbrinlee1:


cli/batch/sql.go
```diff
@@ -175,10 +175,26 @@ func buildBulkInsert(tbl *dax.Table, fields []*dax.Field, ids []interface{}, row
...
+	var keyType string
+	if len(ids) > 0 {
+		switch ids[0].(type) {
+		case uint64:
+			fmtStr = "%d"
+			keyType = dax.BaseTypeID
+		case []byte:
+			fmtStr = "%s"
+			keyType = dax.BaseTypeString
+		}
```
travisturner, line 195: A few things I wonder here.

- My first thought was "why is IDK giving us an invalid table and can we fix that instead of adding all of this switch logic?"
- If we do keep this switch, there should probably be a default case which returns an error for an unexpected type.
- The comment confused me at first because it mentions `[]uint64` and `[]byte`, but then handles `uint64` and `[]byte`. Is the comment meant to read `[][]byte`?

cli/cli_kafka_runner_test.go
travisturner, line 1: Thanks for adding runner tests!
I would like to see this file moved to the `cli/kafka` package and renamed `runner_test.go`, if doing that makes sense. Also, ideally its package is `kafka_test`.

cli/cli_kafka_runner_test.go
```diff
@@ -0,0 +1,343 @@
...
+// These variables are used throught tests to refer to resouces needed for the
+// tests (i.e. where to find the kafka service)
+var (
+	pilosaHost     string
+	pilosaTLSHost  string
+	pilosaGrpcHost string
+	kafkaHost      string
+	registryHost   string
+	certPath       string
+)
```
travisturner, line 29: Can we include the default values here instead of in a bunch of `if/else` blocks below?
For example, 
```
pilosaHost = "pilosa:10101"
```
based on the current logic below.

With that said, I think the default should be the local value, and running in docker or CI should set up with configuration or environment variables.

cli/cli_kafka_runner_test.go
```diff
@@ -0,0 +1,343 @@
...
+
+// The variables above are set here. If local is set to true, the tests assume
+// all services are running locally. Otherwise, we use the name of the
+// containers that these tests will have access to in CI.
+func init() {
+	local := true
```
travisturner, line 35: I think if we can avoid using `init()` functions that's best. Even if it's just a `setup()` function which is then called in the test run.

The hard-coded `local` here obviously won't work, but I'm sure you know that. Just a reminder that test should be configured outside of the code.

Maybe look at the way config is setup here: https://github.com/FeatureBaseDB/featurebase/blob/master/dax/controller/sqldb/test.go

cli/cli_kafka_runner_test.go
```diff
@@ -0,0 +1,343 @@
...
+		} else {
+			registryHost = "schema-registry:8081"
+		}
+	}
+	if certPath, ok = os.LookupEnv("IDK_TEST_CERT_PATH"); !ok {
+		certPath = "/certs"
+	}
+}
+
+// Struct used for TestRunner which contains the information needed to ingest
```
travisturner, line 73: In general, go doc comments on types should begin with the type name, so in this case it would look like:
```
// KafkaRunnerTest is used for ...
```

Also, I'm guessing this type doesn't need to be exported (i.e. it should be called `kafkaRunnerTest`). As it is now, and with it being in the `cli` package (as opposed to `cli_test`), it's part of the exported, public API.

In general, default to making everything unexported (start with a lowercase letter) first. That includes all the members of the struct.

cli/kafka/config.go
```diff
@@ -44,18 +54,73 @@ type ConfigForIDK struct {
...
+	IDField     string
+	PrimaryKeys []string
+	Fields      []idk.RawField
+
+	Encode string
+}
+
+// Generate a Config struct based on a configuration file
+// Default values for Config struct are defined here
+func ConfigStructFromFile(cfgFile string) (cfg Config, err error) {
```
travisturner, line 67: I would recommend called this `ConfigFromFile` (the word "Struct" isn't really helpful).

cli/kafka/config.go
```diff
@@ -8,6 +8,14 @@ import (
...
 	"github.com/featurebasedb/featurebase/v3/idk"
 	"github.com/pkg/errors"
+	"github.com/spf13/viper"
+)
+
+// Kafka message encoding types supported - changes here should be
+// propogated to the Config struct and ValidateConfig error message
+const (
+	JSON string = "json"
+	Avro string = "avro"
```
travisturner, line 18: I would give these names specific to their intended use, so something like:
```
encodingTypeJSON = "json"
encodingTypeAvro = "avro"
```
Note a couple things:
- unexported because you don't use these outside of this package
- the "string" part is implied so you don't need it; a linter will complain.

cli/kafka/config.go
```diff
@@ -44,18 +54,73 @@ type ConfigForIDK struct {
...
-	} else if len(c.Topics) == 0 {
+	}
+
+	if len(c.Topics) == 0 {
 		return errors.Errorf("at least one topic is required")
-	} else if len(c.Fields) > 0 {
+	}
+
+	// validate on a by encoding basis
+	switch c.Encode {
```
travisturner, line 110: is `c.Encode` always expected to have a value? Can it be blank? In other words, do we need a default case here which returns an error? Or is it ok to leave all other cases unvalidated?

cli/kafka/config.go
```diff
@@ -70,46 +135,64 @@ func ValidateConfig(c Config) error {
...
+		// handle primary key
 		if fld.PrimaryKey {
-			out.IDField = fld.Name
-			foundPK = true
+			switch keyType := fld.SourceType; keyType {
+			case dax.BaseTypeID:
+				// no-op
+			case dax.BaseTypeIDSet, dax.BaseTypeStringSet, dax.BaseTypeIDSetQ, dax.BaseTypeStringSetQ:
+				// key should be field that cannot contain multiple values
+				return out, errors.Errorf("Invalid")
```
travisturner, line 189: This error message could be more descriptive; at the very least include the value which is invalid.

cli/kafka/config.go
```diff
@@ -70,46 +135,64 @@ func ValidateConfig(c Config) error {
...
+			default:
```
travisturner, line 190: I think I would be explicit about which types _can_ be used as string primary keys, and have the default case return the error (rather than being explicit about which _can't_).

cli/kafka/config_test_data/config00.toml
travisturner, line 1: Just a note on test data directories. Go will ignore a directory that matches `testdata` (see [this](https://dave.cheney.net/2016/05/10/test-fixtures-in-go)), so you might put this in the directory path `cli/kafka/testdata/config/config00.toml` or something like that.

travisturner (changes requested):
This looks really good.
I left various comments which you can take or leave.
When you're ready to move this out of draft let me know and I'll spend a bit more time running things locally.

cli/cli_kafka_runner_test.go
jrbrinlee1: Made this change. I wasn't able to use the newKafkaRunner method after changing the package type of the runner test so I just changed it to an exported method.

jrbrinlee1: Eh also, noticing why I had it in the cli package to begin with, because I depend on some unexported method and fields. Hmm.. can look at redesigning things.

cli/internal/kafka_util.go
travisturner, line 1: These seem to be used just for tests. I don't think we need an internal package for that; they should just be part of the `*_test` package, and they can include `t.Helper()`

cli/cli_kafka_runner_test.go
```diff
@@ -0,0 +1,343 @@
...
+		if err != nil {
+			t.Fatalf("running runner: %s", err)
+		}
+
+		// and check the results...
+		for _, q := range test.Tests {
+
+			client := &http.Client{}
+			var data = strings.NewReader(q.Query)
+			req, err := http.NewRequest("POST", fmt.Sprintf("http://localhost:10101/index/%s/query", kafkaRunner.Index), data)
```
travisturner, line 255: This url to featurebase classic query endpoint seems problematic. I'm assuming these tests don't work in serverless. It seems like the query should just fbsql to run sql instead of pql commands.

travisturner (commented):
couple more comments.

cli/batch/sql.go
```diff
@@ -191,9 +207,10 @@ func buildBulkInsert(tbl *dax.Table, fields []*dax.Field, ids []interface{}, row
 	// bulk insert as one line in the NDJSON payload. We re-use the map for each
 	// row.
 	m := make(map[string]interface{})
+
 	for i := range rows {
 		// Write the ID value.
-		m[string(dax.PrimaryKeyFieldName)] = ids[i]
+		m[string(dax.PrimaryKeyFieldName)] = fmt.Sprintf(fmtStr, ids[i])
```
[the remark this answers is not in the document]
jrbrinlee1: I didn't spend a lot of time here. But in some case IDK gives us a `[]byte` if the table is keyed. And then it writes the wrong thing to FeatureBase. Here is a test showing that. 
```
Error:          Not equal: 
                                expected: []interface {}{[]interface {}{"1", "a", 20, []interface {}{"hob1", "hob2"}}, []interface {}{"2", "b", 21, []interface {}{"hob2", "hob3"}}, []interface {}{"3", "c", 22, []interface {}{"hob3", "hob4"}}, []interface {}{"4", "d", 23, []interface {}{"hob4", "hob5"}}, []interface {}{"5", "e", 24, []interface {}{"hob5", "hob6"}}, []interface {}{"6", "f", 26, []interface {}{"hob6", "hob7"}}}
                                actual  : []interface {}{[]interface {}{"MQ==", "a", 20, []interface {}{"hob1", "hob2"}}, []interface {}{"Mg==", "b", 21, []interface {}{"hob2", "hob3"}}, []interface {}{"Mw==", "c", 22, []interface {}{"hob3", "hob4"}}, []interface {}{"NA==", "d", 23, []interface {}{"hob4", "hob5"}}, []interface {}{"NQ==", "e", 24, []interface {}{"hob5", "hob6"}}, []interface {}{"Ng==", "f", 26, []interface {}{"hob6", "hob7"}}}
```
I can think about how to avoid this.

cli/batch/sql.go
jrbrinlee1: - I'll take a look at the IDK stuff now. I spent a hour or so trying to figure it out but didn't get to the bottom of it. 
- Good call on the default case. I'll keep that in mind if I change this.
- Thanks for call this out. Yes, it should be `[][]byte`.

cli/kafka/config.go
jrbrinlee1: Default value is define by viper and is JSON. Could add something here so it's clearer when reading though.

cli/batch/sql.go
jrbrinlee1: I found a solution that is odd. Basically, IDK is incorrectly setting config options for the index it tracks. Then dax creates a table that can have the incorrect configuration based on this.

More involved answer, in idk/ingest.go, in the main.clone() function, we create an index with default configuration. Meaning keys is set to false. I don't know if this was a problem prior to the dax changes there but IDK doesn't seem to mind. Anyways, then, when we create a SQL batch, we generate the dax table. We do this by calling IndexInfoToTable(). This sets the type of `_id` based on whether or not the index IDK was tracking had keys or not (which is keys is always false). Thus, tbl.StringKeys() always return false in batching code of the kafka runner.

cli/cli_kafka_runner_test.go
jrbrinlee1: I made the PQL decision quickly after having issues comparing SQL results but have this working now with  the /sql endpoint.

cli/batch/sql.go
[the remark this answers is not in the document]
jrbrinlee1: I decided to change this so it's only formatting with keyed indexes to handle the []byte vs string case.

cli/kafka/config.go
jrbrinlee1: Yep, think this makes sense to me.

cli/kafka/config.go
jrbrinlee1: Yep, missed this. Thank for catching!

cli/batch/sql.go
travisturner: I'd like to understand this a bit more. Can you tell where the table/index options are incorrectly set to begin with? One thing to understand is whether this test is being run against serverless or classic FeatureBase. That will determine which implementation of the `SchemaManager` is being used in the `idk.Main.clone()` method. It sounds to me like the call to `Schema()` there does not contain the table, and I'd like to understand why.

jrbrinlee1:
@travisturner updates based on your comments above are complete. Just working on getting the tests and builds into CI.

cli/batch/sql.go
jrbrinlee1: Yes, that's correct. It's running against classic FeatureBase. I was assuming that meant I was supposed to have a NoOpSchemaManager because that's what I have. But maybe that's wrong? And yes, `index = schema.Index(m.Index)` returns an index with all defaults set. So, essentially, it's not going anywhere to get a schema of an index anywhere.

jrbrinlee1: Ah, normal ingeters setup their client which becomes the schema manager... https://github.com/FeatureBaseDB/featurebase/blob/master/idk/ingest.go#L859. Making that change and will see what happens.

jrbrinlee1: Yep, this fixed things & avro tests are passing now. thanks for the questions @travisturner!

travisturner (changes requested):

.gitlab/.gitlab-ci.yml
```diff
@@ -192,25 +192,42 @@ build fbsql amd64:
...
     - date
-    - GOOS="linux" GOARCH="amd64" make docker-build-fbsql BUILD_CGO=1
+    - GOOS="linux" GOARCH="amd64" make docker-build-fbsql
     - GOOS="darwin" GOARCH="amd64" make docker-build-fbsql
   artifacts:
     paths:
       - ./build/fbsql_*
 
-build fbsql arm64:
+build fbsql arm64 darwin:
```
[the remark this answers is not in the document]
jrbrinlee1: the `build fbsql amd64` step builds darwin amd64. 

I broke arm64 our because `Dockerfile-fbsql-linux` can and did run on `shell-arm64` but the `Dockerfile-fbsql-darwin` doesn't. It gives an exec error. I believe, based on a bit of research, this is because, if I set the `shell-arm64`, `Dockerfile-fbsql-linux` is able to use the `arm64` version of the golang container. But `Dockerfile-fbsql-darwin` depends on the customer image I built which is `amd64`. Take a look at `Dockerfile-darwin-cgo-builder`. There was no way to pull that `Xcode.xip` file from a public server so I built the image locally and pushed it to docker hub.

Happy to talk this one through.

cli/kafka/config.go
```diff
@@ -23,6 +31,11 @@ type Config struct {
 
 	Table  string  `mapstructure:"table" help:"Destination table name."`
 	Fields []Field `mapstructure:"fields"`
+
+	Encode             string `mapstructure:"encode" help:"Encoding format (currently supported formats: avro, json)"`
+	AllowMissingFields bool   `mapstructure:"allow-missing-fields" help:"allow missing fields in messages from kafka"`
+	MaxMessages        int    `mapstructure:"max-messages" help:"max messages read from kakfka"`
+	ConfluentConfig    string `mapstructure:"confluent-config" help:"max messages read from kakfka"`
```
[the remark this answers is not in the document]
jrbrinlee1: Good catch, just updated.

cli/kafka.go
```diff
@@ -1,35 +1,22 @@
...
-	if err != nil {
-		return nil, fmt.Errorf("error reading configuration file '%s': %v", cfgFile, err)
-	}
 
-	cfg := kafka.Config{}
-	if err := v.Unmarshal(&cfg); err != nil {
-		return nil, errors.Wrap(err, "unmarshalling config")
+	cfg, err := kafka.ConfigFromFile(cfgFile)
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
jrbrinlee1: Ah yep, I must have thought ConfigFromFile was going to include that info but it doesn't. Thanks!

cli/kafka.go
```diff
@@ -57,11 +44,16 @@ func (cmd *Command) newKafkaRunner(cfgFile string) (*kafka.Runner, error) {
...
+	flds, err := kafka.ConfigToFields(cfg, idkCfg.PrimaryKeys)
 	if err != nil {
 		return nil, errors.Wrap(err, "getting fields from config")
 	}
 
+	// for avro, let the SchemaManager and IDK handle fields
+	if cfg.Encode == "avro" {
+		flds = nil
+	}
+
```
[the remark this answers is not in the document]
jrbrinlee1: Yep, make sense. Was trying a few different things so just placed it here to test and then never revisited. I I think I need to `return nil, nil` here because `*sqlBatcher.NewBatch()` is looking for `nil` vs and empty slice.
```
func (b *sqlBatcher) NewBatch(cfg fbbatch.Config, tbl *dax.Table, flds []*dax.Field) (fbbatch.RecordBatch, error) {
	fields := flds
	if b.fields != nil {
		fields = b.fields
	}
        ...
```

cli/kafka/config.go
```diff
@@ -44,18 +57,75 @@ type ConfigForIDK struct {
...
+		return validateConfigAvro(c)
+	}
+
+	return nil
+
+}
+
+func validateConfigJSON(c Config) error {
+
+	if len(c.Fields) > 0 {
```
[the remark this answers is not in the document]
jrbrinlee1: Yep, great idea! Made this change.

cli/kafka/runner.go
```diff
@@ -7,61 +7,103 @@ import (
...
+		Topics:  cfg.Topics,
+		Header:  cfg.Fields,
+		Timeout: cfg.Timeout,
+	}
+
+	// NewSource should be set based on the encoding of the source (e.g. JSON, Avro)
+	if cfg.Encode == encodingTypeAvro {
+		kr.GetAvroNewSource(cfg)
+	} else if cfg.Encode == encodingTypeJSON {
+		kr.GetJSONNewSource(cfg)
```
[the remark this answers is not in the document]
jrbrinlee1: If it's not one of these two options, we'll catch it in the call to `ValidateConfig()`. Also, the default config options is "json" so if the user didn't specify a `encode` config, it would be "json".

But, I'll make this change so it's more readable and we have multiple checks for the same thing. Just need to update a few places if we ever wanted to make a change to the defaults. But I don't see that as an issue.

idk/ingest.go
```diff
@@ -309,13 +309,51 @@ func (m *Main) clone() (*Main, error) {
 
 	index = schema.Index(m.Index)
 
-	// use a copy (schema race condition issues)
+	// use a copy (schema race condition issu)
```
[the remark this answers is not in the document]
jrbrinlee1: lol, sorry. I caught this as well just earlier today. Must have been one of those cases I thought I was typing in another window and I missed it in my check before committing.

idk/ingest.go
```diff
@@ -309,13 +309,51 @@ func (m *Main) clone() (*Main, error) {
...
+// 		index = schema.Index(m.Index)
+// 	}
+
+// 	// use a copy (schema race condition issues)
+// 	mClone := *m
+// 	mClone.index = index
+
+// 	return &mClone, nil
+// }
+
```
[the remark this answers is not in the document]
jrbrinlee1: Nope, also saw this this morning. It'll be removed in the next commit.

idk/ingest.go
```diff
@@ -2164,7 +2203,7 @@ func (m *Main) newBatch(clientFields []*pilosaclient.Field) (pilosabatch.RecordB
 	ii := pilosaclient.FromClientIndex(m.index)
 	tbl := pilosacore.IndexInfoToTable(ii)
 
-	// Fields.
+	// Fields. // this is giving bad fieldInfos
```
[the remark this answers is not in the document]
jrbrinlee1: Also saw this this morning. It was a note to self when I was trying to debug. It'll be removed in the next commit.

wire_response.go
```diff
@@ -210,6 +211,15 @@ func (ii IDSet) String() string {
...
 
+// SortedInt64Slice returns the values in a IDSet field in a int64 slice that is
+// sorted
+func (ii IDSet) SortedInt64Slice() []int64 {
+	var idSetSlice = make([]int64, len(ii))
+	copy(idSetSlice, ii)
+	sort.Slice(idSetSlice, func(i, j int) bool { return idSetSlice[i] < idSetSlice[j] })
+	return idSetSlice
+}
+
```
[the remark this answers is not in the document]
jrbrinlee1: It was just cleaner / less convoluted to implement here. Can't remember the issue exactly but we have IDSet and StringSet types which are aliases for []string and []int64 or something. So I was checking type and converting things, etc. What is the down side to having this in the pilosa package? I mean, no ones asked for it yet, but sorting the output of a set fields could be a feature at some point.

But wrt to the copy, I think my thought was to prevent modifying the object itself just to be safe. But now that I think about it again, I'm not entirely sure what it's being safe from.

If you don't want this in the pilosa package, I can remove it and implement in the test package. Otherwise, I'll remove the copy and keep as is. Please let me know.

idk/kafka_sasl/source.go
```diff
@@ -154,32 +171,41 @@ func (r *Record) Schema() interface{} { return nil }
...
 	if idx < base {
 		return errors.New("cannot commit a record that has already been committed")
 	}
 	section, remaining := r.src.spool[:idx-base], r.src.spool[idx-base:]
-	// sort by increasing partition, decreasing offset
 	sort.Slice(section, func(i, j int) bool {
-		if section[i].Partition != section[j].Partition {
+		if *section[i].Topic != *section[j].Topic {
+			return *section[i].Topic < *section[j].Topic
+		} else if section[i].Partition != section[j].Partition {
```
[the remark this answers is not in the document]
jrbrinlee1: I didn't look at this but I can. I copied this from the `kafka` consumer. At one point Q2 was incorrectly committing offsets with the `kafka` consumer and this was the logic that fixed it.

In general, I tried to move kafka source logic to the kafka_sasl source where it made sense to do so. Since I joined, I haven't known any customers to use the `kafka_sasl` consumer but Q2 has use the `kafka` consumer heavily. So I'm working of the assumptions that the `kafka` consumer is better tested so changes made there sound be made in the `sasl` consumer as well.

idk/kafka_sasl/source.go
```diff
@@ -193,14 +219,29 @@ func (r *Record) Data() []interface{} {
 
 // Open initializes the kafka source.
 func (s *Source) Open() error {
-	if len(s.Header) == 0 {
-		return errors.New("needs header specification file")
+	cfg, err := common.SetupConfluent(&s.ConfluentCommand)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
jrbrinlee1: Updated in both consumers.

idk/kafka_sasl/source.go
```diff
@@ -242,6 +292,10 @@ func (s *Source) Open() error {
 		return errors.Wrap(err, "subscribe topics")
 	}
 
+	if s.Verbose {
```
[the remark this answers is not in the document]
travisturner: I suppose it makes sense to wrap things in the `if s.Verbose` if it's doing additional work, but if it's just wrapping a `Debugf()`, we should remove the wrap.

wire_response.go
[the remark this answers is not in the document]
jrbrinlee1: Yeah, okay.. this was just me being dumb. Moved this logic to the test.

cli/kafka/config.go
```diff
@@ -70,46 +140,85 @@ func ValidateConfig(c Config) error {
...
 		return out, errors.New("fields cannot be empty")
 	}
 
-	// rawFields wil be the same as c.Fields, but possibly enhanced.
+	// rawFields will be the same as c.Fields, but possibly enhanced.
 	rawFields := make([]idk.RawField, 0, len(c.Fields))
 
-	var foundPK bool
+	var stringKeys bool
+	primaryKeys := []string{}
```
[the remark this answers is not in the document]
jrbrinlee1: Yes, this is my understanding as well. But I can double check this with a test and add the information to docs and code accordingly.

PR #2360, opened 2023-03-30
handle tag-only commits with no hash in version

seebs:
When doing the release process, we generate version numbers that have a version tag but don't have a hash. The IDK test against the expected hash doesn't work in this context. Let's check for an expected tag first.

### DIFF
--- a/idk/ingest_test.go
+++ b/idk/ingest_test.go
@@ -89,6 +89,19 @@ func TestFeaturebaseVersion(t *testing.T) {
 		t.Fatalf("unmarshalling version: %v", err)
 	}
 	v := vh.Version
+	expectedTag := os.Getenv("IDK_FEATUREBASE_TAG")
+	if expectedTag != "" {
+		// if a tag is set, we're in a tagged pipeline, and the version
+		// should just be vX.Y or something similar, without a commit
+		// hash.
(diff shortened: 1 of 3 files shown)


idk/ingest_test.go
```diff
@@ -89,6 +89,19 @@ func TestFeaturebaseVersion(t *testing.T) {
...
+		// should just be vX.Y or something similar, without a commit
+		// hash.
+		if v != expectedTag {
+			t.Fatalf("version %s does not match expected tag %s", v, expectedTag)
+		}
+		t.Logf("featurebase version %q matches expectations", expectedTag)
+		return
+	}
+	// no tag. we expect to always have a hash, and if there's a specific
+	// expected hash, we verify it.
```
gardavi, line 104: I independently came up with this lol
```go
expectedTagVersion := os.Getenv("FEATUREBASE_TAG")
if expectedTagVersion != "" {
	if expectedTagVersion != v {
		t.Fatalf("featurebase version: %s, expected version: %s", v, expectedTagVersion)
	}
	t.Logf("featurebase version %q matches tag version", expectedTagVersion)
	return
}
```

gardavi (approved):
LGTM; 👍

PR #2362, opened 2023-03-31
make percentile work with where clauses and on decimal fields

seebs:
no worries this one should be a simple one-liner, right?

actually it's four distinct bugs.

```
fbsql=# select percentile(x, 50) from i where x < 13;
------
 5.00

fbsql=# select percentile(x, 50) from i where x < 13.0;
-------
 10.00
```

In this, we see the following bugs:

1. With `where x < 13`, we actually use an SQL3-internal implementation of Percentile, rather than the PQL implementation, because `13` is an int64 and not a float64, and we don't allow int64 in Decimal conditions.
2. With `where x < 13.0`, the `WHERE` clause is not actually exported to the PQL Percentile call as a filter.
3. With or without the `WHERE` clause, PQL Percentile produces nonsense results (always the minimum, basically) for decimal fields.
4. The SQL3-internal implementation of Percentile is actually just a copy of `Count`, rather than actually trying to implement percentile.

This took some sorting.

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -1294,129 +1294,310 @@ func (e *executor) executeMax(ctx context.Context, qcx *Qcx, index string, c *pq
 }
 
 // executePercentile executes a Percentile() call.
-func (e *executor) executePercentile(ctx context.Context, qcx *Qcx, index string, c *pql.Call, shards []uint64, opt *ExecOptions) (_ ValCount, err error) {
+//
+// To compute the percentile, we find the maximum and minimum values that match
+// our filter (or an implicit filter of "value isn't null"), and also count values
+// matching our filter. We convert our percentile to an approximate number of
+// values that should be higher or lower than the desired value. If either of those
+// is zero, we return the minimum/maximum; otherwise, we do a binary search of
+// the range between minimum and maximum, looking for a value which has the
+// desired number of values higher or lower than it.
+//
+// Unfortunately, each step in this process is its own, separate, cluster-wide
+// query. this should be replaced with a modern probabilistic algorithm, which
+// could accumulate statistical information per shard and combine that information
+// in a single pass.
+func (e *executor) executePercentile(ctx context.Context, qcx *Qcx, index string, c *pql.Call, shards []uint64, opt *ExecOptions) (result interface{}, err error) {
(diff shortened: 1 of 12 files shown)


executor.go
```diff
@@ -1358,65 +1444,156 @@ func (e *executor) executePercentile(ctx context.Context, qcx *Qcx, index string
...
+	var averageMinMax func() interface{}
+	var minLessthanMax func() bool
+	var maxValueUnder func(interface{})
+	var minValueOver func(interface{})
+
+	if field.options.Type == FieldTypeDecimal {
+		minPtr := minVal.DecimalVal
+		maxPtr := maxVal.DecimalVal
+		if minPtr == nil || maxPtr == nil {
+			return nil, fmt.Errorf("unexpectedly nil min/max values")
```
lorycloutier, line 1468: for diagnosing things I think I'd prefer if nil minPtr and nil maxPtr gave separate error messages.

seebs:
Talked to Pat, got more feedback, abandoned my plan to complain about every failure to convert to PQL, because the actual reason the filters were getting lost was *coincidental* to that -- or rather, required both that *and* that the filter in question was using an aliased name. Anyway, this should actually be done now.

executor.go
seebs: that's a good call and i wish i'd read this BEFORE i propagated changes to dax.

lorycloutier (approved):

sql3/test/defs/defs_aggregate.go
```diff
@@ -514,9 +514,7 @@ var percentileTests = TableTest{
 				hdr("p_rows", fldTypeDecimal2),
 			),
 			ExpRows: rows(
-				// This should probably be (1200, 2), not (1000, 2).
-				// TODO: look into this when investigating the percentile/WHERE bug.
-				row(pql.NewDecimal(1000, 2)),
+				row(pql.NewDecimal(1150, 2)),
```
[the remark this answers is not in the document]
seebs: So far as I know, a valid "median" can be a value that's not in the set. It's not maybe entirely consistently or rigorously defined, anyway.

PR #2363, opened 2023-03-31
Fb 2048

lorycloutier:
Added test coverage for ExecutionPlanner.mapper in the case where the context gets cancelled mid-query.
This involved making changes to MustRunQuery such that it can be provided with a context to use. If it's given nil, it works exactly as it did before and makes its own context, but if it's given a context, it uses that and also pauses briefly to make sure that the context can be cancelled in the right place. This will allow us, in the future, to write more tests for cancelled contexts as well.


sql3/test/helpers.go
```diff
@@ -14,14 +14,23 @@ import (
 )
 
 // MustQueryRows returns the row results as a slice of []interface{}, along with the columns, the query plan as a []byte or an error.
-func MustQueryRows(tb testing.TB, svr *featurebase.Server, q string) ([][]interface{}, []*featurebase.WireQueryField, []byte, error) {
+func MustQueryRows(tb testing.TB, svr *featurebase.Server, c context.Context, q string) ([][]interface{}, []*featurebase.WireQueryField, []byte, error) {
```
seebs, line 18: I... have bad news.

There's a fairly strong idiom that ctx parameters Come First. There's also a strong idiom that testing.TB parameters Come First. But I think that, for maintainence-is-easier reasons, we should move the context in front of the server here.

I'm aware this requires a large search and replace elsewhere, sorry.

sql3/sql_complex_test.go
```diff
@@ -127,6 +127,26 @@ func TestPlanner_SystemTableFanout(t *testing.T) {
...
 		}
 	})
+
+	// additional testing for *ExecutionPlanner.mapper is going here because
+	// this is where the existing coverage comes from.
+	t.Run("MapperContextCancellation", func(t *testing.T) {
+		ctx, cancel := context.WithCancel(context.Background())
+		errch := make(chan error)
+		go func() {
+			_, _, _, err := sql_test.MustQueryRows(t, server, ctx, `select * from fb_performance_counters`)
```
seebs, line 137: There is a subtle bug here, which we should probably fix, which is that we Can't Do This In A Goroutine.

The testing stuff strictly requires that, if you call `t.Fatalf` or the equivalent, you do so from the goroutine the test starts in.

In short: You should put the `cancel()` in a goroutine (after a sleep), and call MustQueryRows directly after launching that goroutine.

seebs (changes requested):
The basic approach seems good to me, and I think I buy the delay and context stuff, but the parameter order is unidiomatic in a way I think could confuse us later, and the cancel has to be the part in a goroutine due to a subtlety of the testing API.

seebs (approved):
Looks good to me. I think I'd slightly prefer squashing these into a single commit, because that would make the output of `git blame` more useful, but honestly up to you.

PR #2365, opened 2023-04-03
don't panic on a MIN that isn't a call

seebs:
parseOperand was assuming that any reference to MIN in a place where an operand was expected was a call, which it should be, but it might not be. parseCallExpression panics if it doesn't find a parenthesis, because it's never supposed to be called when we don't know we have one.

The test for this is in with MinMaxColumnConstraints, even though it's actually a test of MinMaxFunctionCalls, because that's where the other tests involving the special MIN/MAX tokens live.

### DIFF
--- a/sql3/parser/parser.go
+++ b/sql3/parser/parser.go
@@ -2742,8 +2742,12 @@ func (p *Parser) parseOperand() (expr Expr, err error) {
 	case VARIABLE:
 		return &Variable{Name: lit, NamePos: pos}, nil
 	case MIN, MAX:
-		ident := &Ident{Name: lit, NamePos: pos, Quoted: tok == QIDENT}
-		return p.parseCall(ident)
+		pk := p.peek()
+		if pk == LP {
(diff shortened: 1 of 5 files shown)


rachithrr (dismissed):
This looks good. 
I have a question in general, 
why do we handle MIN and MAX specifically here ? why not AVG and COUNT, which are similar ? 


we handle all these functions in executionplanner.

rachithrr (approved):
lgtm. 

On a different note. 
While reviewing this, i found that the query 
```
create table sads(_id id, val int min -10 max -20)
```
crashes with nil pointer. 

maybe, we should handle this gracefully.

seebs:
oh, good catch. i'll fix that too while i'm looking at it.

oh. d'oh. that one's my fault from the trackexistence stuff.

seebs:
hah. i was just thinking "huh we don't have a test case for max being less than min. well, surely it works." i was wrong.

good catch.

rachithrr (approved):
All good 🥇

PR #2367, opened 2023-04-04
Cloud 1469 meter serverless consumption in amberflo

andreacappelletti97:


dax/controller/controller.go
```diff
@@ -784,6 +784,18 @@ func (c *Controller) Databases(ctx context.Context, orgID dax.OrganizationID, id
 	return c.Schemar.Databases(tx, orgID, ids...)
 }
 
+// get the list of workers being used by a database
+func (c *Controller) GetDatabaseWorkerCount(ctx context.Context, qdbid dax.QualifiedDatabaseID) (int, error) {
+
+	tx, err := c.Transactor.BeginTx(ctx, false)
+	if err != nil {
+		return -1, errors.Wrap(err, "beginning tx")
```
jaffee, line 792: this should return 0, not -1. Go convention is to return the zero value when there's a non-nil error

jaffee: (applies everywhere else as well)

dax/controller/balancer/balancer.go
```diff
@@ -639,6 +639,10 @@ func (b *Balancer) WorkersForTable(tx dax.Transaction, roleType dax.RoleType, qt
 	return workers, nil
 }
 
+func (b *Balancer) WorkerCount(tx dax.Transaction, qdbid dax.QualifiedDatabaseID) (int, error) {
```
jaffee, line 642: wrong name? Maybe change the name on the balancer to be WorkerCount

dax/controller/controller.go
```diff
@@ -784,6 +784,18 @@ func (c *Controller) Databases(ctx context.Context, orgID dax.OrganizationID, id
 	return c.Schemar.Databases(tx, orgID, ids...)
 }
 
+// get the list of workers being used by a database
+func (c *Controller) GetDatabaseWorkerCount(ctx context.Context, qdbid dax.QualifiedDatabaseID) (int, error) {
```
travisturner, line 788: Go convention typically avoids using the term "get" in methods. There is an exception in the handler because that "get" signifies that it's related to the GET http method. This method should be something like `DatabaseWorkerCount()`.

travisturner (commented):
left a few comments.

dax/controller/sqldb/workerjob.go
```diff
@@ -72,12 +72,22 @@ func jobsForWorker(dt *DaxTransaction, worker *models.Worker, roleType dax.RoleT
...
-	return cnt, errors.Wrap(err, "getting count")
+	query := dt.C.Q()
+
+	if roleType != "" {
+		query = query.Where(fmt.Sprintf("role_%s = true", roleType))
+	}
+
+	if qdbid.DatabaseID != "" {
+		query = query.Where("database_id = ?", qdbid.DatabaseID)
+	}
```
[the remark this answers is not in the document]
jaffee: yeah let's just error if databaseID is empty, this was my fault

dax/controller/controller.go
andreacappelletti97: Conclusion: just call it `WorkerCount()`

PR #2373, opened 2023-04-06
implemented query hints (flatten) (fb-2124)

paddyjok:
This PR implements  query options hints that govern behavior for distinct and group by on set columns.

1. introduces syntax to support query hints on table source references `...from table_name with (hint(param))`
2. introduces the `flatten(column)` hint 
3. ensures `distinct` with and without flatten with a single set column ref works correctly
4. ensures `group by` with and without flatten with a single set column ref works correctly

### DIFF
--- a/sql3/parser/ast.go
+++ b/sql3/parser/ast.go
@@ -4230,13 +4230,14 @@ func (n *QualifiedTableName) String() string {
 	}
 
 	if n.With.IsValid() {
-		buf.WriteString(" WITH")
+		buf.WriteString(" WITH (")
 		for i, o := range n.QueryOptions {
 			if i > 0 {
 				buf.WriteString(", ")
 			}
 			fmt.Fprintf(&buf, " %s", o.String())
 		}
+		buf.WriteString(")")
 	}
 	return buf.String()
 }
--- a/sql3/parser/ast.go
+++ b/sql3/parser/ast.go
@@ -90,6 +90,7 @@ func (*RollbackStatement) node()        {}
 func (*SavepointStatement) node()       {}
 func (*SelectStatement) node()          {}
 func (*StringLit) node()                {}
+func (*TableQueryOption) node()         {}
 func (*TableValuedFunction) node()      {}
 func (*TimeUnitConstraint) node()       {}
 func (*TimeQuantumConstraint) node()    {}
@@ -4139,15 +4140,45 @@ func (c *ResultColumn) String() string {
 	return c.Expr.String()
 }
 
+type TableQueryOption struct {
+	OptionName   *Ident
+	LParen       Pos
+	OptionParams []*Ident
+	RParen       Pos
+}
+
+func (n *TableQueryOption) Clone() *TableQueryOption {
+	if n == nil {
+		return nil
+	}
+	other := *n
+	other.OptionName = n.OptionName.Clone()
+	other.OptionParams = cloneIdents(n.OptionParams)
+	return &other
+}
+
+func (n *TableQueryOption) String() string {
+	var buf bytes.Buffer
+	buf.WriteString(n.OptionName.String())
+	buf.WriteString("(")
+	for i, o := range n.OptionParams {
+		if i > 0 {
+			buf.WriteString(", ")
+		}
+		fmt.Fprintf(&buf, " %s", o.String())
+	}
+	buf.WriteString(")")
+	return buf.String()
+}
+
 type QualifiedTableName struct {
-	Name          *Ident                // table name
-	As            Pos                   // position of AS keyword
-	Alias         *Ident                // optional table alias
-	Indexed       Pos                   // position of INDEXED keyword
-	IndexedBy     Pos                   // position of BY keyword after INDEXED
-	Not           Pos                   // position of NOT keyword before INDEXED
-	NotIndexed    Pos                   // position of NOT keyword before INDEXED
-	Index         *Ident                // name of index
+	Name          *Ident // table name
+	As            Pos    // position of AS keyword
+	Alias         *Ident // optional table alias
+	With          Pos    // position of WITH keyword
+	LParen        Pos
+	QueryOptions  []*TableQueryOption
+	RParen        Pos
 	OutputColumns []*SourceOutputColumn // output columns - populated during analysis
 }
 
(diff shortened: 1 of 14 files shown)


sql3/parser/ast.go
```diff
@@ -4187,10 +4229,14 @@ func (n *QualifiedTableName) String() string {
...
 
-	if n.Index != nil {
-		fmt.Fprintf(&buf, " INDEXED BY %s", n.Index.String())
-	} else if n.NotIndexed.IsValid() {
-		buf.WriteString(" NOT INDEXED")
+	if n.With.IsValid() {
+		buf.WriteString(" WITH")
+		for i, o := range n.QueryOptions {
+			if i > 0 {
+				buf.WriteString(", ")
```
seebs, line 4236: I don't think you need the space after the comma, because the option will add a space before itself.

sql3/planner/oppqldistinctscan.go
```diff
@@ -265,26 +265,18 @@ func (i *distinctScanRowIter) Next(ctx context.Context) (types.Row, error) {
 			row[0] = pql.NewDecimal(val, t.Scale)
 
 		case *parser.DataTypeIDSet:
-			val, ok := result.([]uint64)
+			val, ok := result.(int64)
```
seebs, line 268: I don't actually know of anything wrong here but this feels like a surprising change in semantics. Or is this because we're changing the semantics of distinct or something?

sql3/test/defs/defs_groupby.go
```diff
@@ -257,3 +257,171 @@ var groupByTests = TableTest{
...
+			srcRow(int64(3), []int64{1, 4}, []string{"a", "d"}),
+			srcRow(int64(4), []int64{3, 2}, []string{"c", "b"}),
+		),
+	),
+	SQLTests: []SQLTest{
+		{
+			SQLs: sqls(
+				"select distinct ids1 from groupby_set_test with (flatter(foo))",
+			),
+			ExpErr: "unknown query hint 'flatter'",
```
seebs, line 282: I was already moving the mouse to say "i think i found a typo" when my reading caught up with the expected error. :)

sql3/test/defs/defs_groupby.go
```diff
@@ -257,3 +257,171 @@ var groupByTests = TableTest{
...
+var groupBySetDistinctTests = TableTest{
+	Table: tbl(
+		"groupby_set_test",
+		srcHdrs(
+			srcHdr("_id", fldTypeID),
+			srcHdr("ids1", fldTypeIDSet),
+			srcHdr("ss1", fldTypeStringSet),
+		),
+		srcRows(
+			srcRow(int64(1), []int64{1, 2}, []string{"a", "b"}),
```
seebs, line 271: I'd like to have at least one duplicate set value so the counts we're testing against aren't all 1, which makes it easier to verify that the flattening, etc., is working correctly.

sql3/planner/oppqldistinctscan.go
paddyjok: Yeah - the original code was never hit, and when it was result was not as we expected.

sql3/test/defs/defs_groupby.go
```diff
@@ -257,3 +257,171 @@ var groupByTests = TableTest{
...
+				row([]string{"d", "e"}),
+				row([]string{"a", "d"}),
+				row([]string{"b", "c"}),
+			),
+			Compare:        CompareExactUnordered,
+			SortStringKeys: true,
+		},
+		{
+			SQLs: sqls(
+				"select distinct ss1 from groupby_set_test with (flatten(ss1))",
```
seebs, line 376: can we also do `select distinct ss1, ids1`? how does that interact with flattening?

i'm not sure what i expect but whatever it is, i'd like to be sure it works for all three of:
* neither flattened
* both flattened
* only one flattened

because that feels like it's a case where things might blow up.

seebs (changes requested):
This looks good but I would like to push for slightly more comprehensive testing just because this is a pretty major new feature surface.

sql3/parser/ast.go
paddyjok: Comma is OK, but I missed some parens...

sql3/test/defs/defs_groupby.go
paddyjok: In this case - flatten is ignored, since we have more than on column ref.

sql3/test/defs/defs_groupby.go
```diff
@@ -323,6 +324,40 @@ var groupBySetDistinctTests = TableTest{
...
+		},
+		{
+			SQLs: sqls(
+				"select distinct ids1, ss1 from groupby_set_test with (flatten(ids1))",
+			),
+			ExpHdrs: hdrs(
+				hdr("ids1", fldTypeIDSet),
+				hdr("ss1", fldTypeStringSet),
+			),
+			ExpRows: rows(
```
seebs, line 352: This doesn't look like `ids1` got flattened?

I am ... not actually sure what I expect here. I think I sort of expect the id set column values to all be 1, 2, 3, or 4, rather than sets, because they're being flattened. And I think that implies something like

1, {"a", "b"}
2, {"a", "b"}
3, {"d", "e"}
4, {"d", "e"}

etcetera? but now i'm not sure.

seebs (dismissed):
so, the behavior of flatten when there's more than one column being selected is "just ignore it", which i think maybe ought to get a ticket, but that's the understood/expected behavior, so i think we can be okay with this for now? but maybe it should be an error to specify a flatten that won't have any effect, because otherwise it might confuse users.

So i'm okay with this as is, although i think i'd prefer "error on a flatten we won't do" because otherwise users won't understand why it's not happening, if we can do that now. otherwise, i guess, ticket it for doing that next.

seebs:
looking at the CI failures, they look intentional in that we changed the semantics of group by. e.g., we would now need:

```
--- a/sql3/test/defs/defs_groupby.go
+++ b/sql3/test/defs/defs_groupby.go
@@ -225,7 +225,7 @@ var groupByTests = TableTest{
                },
                {
                        SQLs: sqls(
-                               "SELECT COUNT(*), is1 FROM groupby_test group by is1",
+                               "SELECT COUNT(*), is1 FROM groupby_test with(flatten(is1)) group by is1",
                        ),
                        ExpHdrs: hdrs(
                                hdr("", fldTypeInt),
```

seebs:
We do need to fix the existing tests, because we've fixed group by semantics to match SQL semantics by default, which is intentional, but it does mean the existing tests will need updates. I'm fine with either direction of changing that (changing expected test results or adding the with(flatten(...)) to them).

seebs (approved):

tgruben (approved):

PR #2374, opened 2023-04-07
tidy up show tables behavior

paddyjok:
This PR changes `show tables` functionality in the following ways:

1. all tables are sorted by ID in the output
2. system tables (currently predicated by a "fb_" name prefix) are filtered out by default
3. `show tables` now has a `with system` option to show system tables


cli/testdata/table
```diff
@@ -33,16 +37,11 @@ EXPECT:
...
-EXPECTCOMP:WithFormat:| fb_____________________ | fb_____________________ |       |            | {timestamp} | {timestamp} | false |          0 |             |
-EXPECTCOMP:WithFormat:| fb_____________________ | fb_____________________ |       |            | {timestamp} | {timestamp} | false |          0 |             |
-EXPECTCOMP:WithFormat:| fb_____________________ | fb_____________________ |       |            | {timestamp} | {timestamp} | false |          0 |             |
-EXPECT:+-------------------------+-------------------------+-------+------------+----------------------+----------------------+-------+------------+-------------+
+EXPECT:+-------+-------+-------+------------+----------------------+----------------------+-------+------------+-------------+
+EXPECT:| _id   | name  | owner | updated_by | created_at           | updated_at           | keys  | space_used | description |
+EXPECT:+-------+-------+-------+------------+----------------------+----------------------+-------+------------+-------------+
+EXPECTCOMP:WithFormat:| users | users |       |            | 1970-01-01T00:00:00Z | 1970-01-01T00:00:00Z | false |          0 |             |
```
travisturner, line 43: this should use `{timestamp}` instead of actual timestamps. (which makes me wonder why those timestamps are unix time 0... that may need to get fixed somewhere else).

travisturner (changes requested):

cli/testdata/table
travisturner: included this in #2376

travisturner (approved):

PR #2378, opened 2023-04-07
some AST and parser test coverage expansion

seebs:
Now that we support function bodies, we add a test case or two for them, which reveal that we're not round-tripping, so we tweak things slightly to make that work. There's an open issue there with semicolon semantics that we should revisit once we allow more than one kind of trivial expression in function bodies.

Trying to figure out why the second case in time parsing never got evaluated, I discovered that it has no reason to exist. Took this as an excuse to standardize on a single time-parser that lives in one place and is used by everything else, which makes the code a bit shorter in most cases, and lets us be more confident that the semantics are always the same. This looks intrusive but it's actually pretty trivial.

Improves test coverage in ast.go and walk.go slightly, and might actually help slightly with some other files by removing lines of code that were actually unreachable but weren't obviously-unreachable until you spend ten minutes staring closely at the docs for the `time` package.

### DIFF
--- a/sql3/parser/ast.go
+++ b/sql3/parser/ast.go
@@ -198,6 +198,8 @@ func CloneStatement(stmt Statement) Statement {
 		return stmt.Clone()
 	case *ReleaseStatement:
 		return stmt.Clone()
+	case *ReturnStatement:
+		return stmt.Clone()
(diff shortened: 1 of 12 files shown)


seebs:
this got merged in a different branch.

PR #2379, opened 2023-04-10
fb-2026 : added test cases with cache type & size for copy

HHans09:
This PR caters to the test cases added for testing `copy` with cache type & cache size different than the default Pilosa Cache type & size for fld type - `fldTypeID, fldTypeIDSet, fldTypeString, fldTypeStringSet`


HHans09:
Merged in private repo

